PR #153, opened 2023-02-22
update peer deps to allow use of core 4.0

aruniverse:
should hopefully prevent install warnings when trying to use core@3 with appui@4


raplemie (approved):
LGTM

PR #216, opened 2023-03-18
Update to TypeScript 5.0

wgoehrig:
**Note:** Depends on https://github.com/iTwin/itwinjs-core/pull/5242

With the 4.0 release coming up, now's as good a time as ever to catch up on the past year and a half of typescript updates!

Since we're jumping up by six(!) TypeScript versions (note that TypeScript does not follow semver), it's worth checking out the following to see what's changed since 4.4:
- [TS 4.5 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-5.html)
- [TS 4.6 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-6.html)
- [TS 4.7 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-7.html)
- [TS 4.8 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-8.html)
- [TS 4.9 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html)
- [TS 5.0 release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/)

In general, with these updates, you can expect better type inferencing and faster compilation.  A few other highlights are:

- [The `satisfies` Operator](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#the-satisfies-operator) - Allows an expression to keep its (more specific) inferred type, while still letting you ensure it will always _match_ a type alias or interface.  Very cool!
- [Auto-Accessors in Classes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-4-9.html#a-nameauto-accessors-in-classes-auto-accessors-in-classes) - This skips a lot of boilerplate when defining trivial getters and setters.  Seems silly, but opens the door to some some interesting possibilities when used with...
- [Decorators!](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#decorators) - Over 5 years in the making, ECMAScript decorators are almost here, and TypeScript support for them is no longer behind an opt-in compiler flag.  There are some important [differences](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#differences-with-experimental-legacy-decorators) between the new ES decorators and the legacy "experimental" decorators, so for now I've left the `--experimentalDecorators` compiler flag enabled, since iTwin.js had been using experimental decorators in a few places.  We should plan to convert those to support the new "official" decorator behavior and remove the the `--experimentalDecorators` flag in future PR.

As with all TypeScript updates, there were some breaking changes.  In addition to the lib.d.ts changes, most of the new _compiler_ errors this time around were caused by a significant [change to how enums work](https://devblogs.microsoft.com/typescript/announcing-typescript-5-0/#all-enums-are-union-enums) in TS 5.0.  But due to the new decorator support, I've also had to update several of our eslint plugins, so most of the changes in this PR are fixing new lint errors.  I've tried to split each category of lint error fix into separate commits, in the hope of making this mammoth change a bit easier to review.

Lastly, TypeScript 4.7 and 5.0 both introduced features that should make it easier for us to eventually drop support for CommonJS, while still supporting bundlers like webpack!  Stay tuned for more on that front.. 😄


raplemie:
@wgoehrig Do you know if this will force all our package users to adopt 5.0 ? (Are there breaking changes in the generated d.ts that you already know before I dig in the release notes ?)

wgoehrig:
> @wgoehrig Do you know if this will force all our package users to adopt 5.0 ? (Are there breaking changes in the generated d.ts that you already know before I dig in the release notes ?)

I'm not aware of any that we should expect right out of the gate, but of course if you start using some of the new language features like `satisfies` that can be breaking.

IIRC, users can sometimes avoid being impacted by breaking changes with the [skipLibCheck](https://www.typescriptlang.org/tsconfig#skipLibCheck) compiler option, which I believe a lot of apps already set.  

Another (last resort) option if this were to become a significant issue is the `typesVersions` [package.json field](https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions), which would allow you to publish older .d.ts files alongside the new ones.  I've always hoped to avoid that though, because I have no idea how we would be generating those (multiple builds?  🤮)

GerardasB (approved):
Would be great to add `typescript.tsdk` to `settings.json`

wgoehrig:
@GerardasB @raplemie Any idea what to do about the [test failures](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2035454&view=logs&j=9f3a3068-7880-5994-433d-b92c1f9d379b&t=429853d2-592c-5f99-537b-6a3154c944ed&l=4545)?

As best I can tell, it looks like a (semver-compatible) update to @testing-library/dom or one of it's dependencies has broken the behavior of `getByRole` in that the `title` attribute is no longer being considered when determining the accessible name.  I'm not sure how relevant this is, but [this issue](https://github.com/A11yance/aria-query/issues/379#issuecomment-1227863154) on GitHub seems to suggest this may be intentional...

Tests are failing at:

https://github.com/iTwin/appui/blob/627e041687eac6787810a2c8be3d2cfb14f0106a/ui/components-react/src/test/properties/renderers/value/ArrayPropertyValueRenderer.test.tsx#L60
https://github.com/iTwin/appui/blob/627e041687eac6787810a2c8be3d2cfb14f0106a/ui/components-react/src/test/properties/renderers/value/StructPropertyValueRenderer.test.tsx#L49
https://github.com/iTwin/appui/blob/627e041687eac6787810a2c8be3d2cfb14f0106a/ui/core-react/src/test/button/UnderlinedButton.test.tsx#L30

NancyMcCallB:
> @GerardasB @raplemie Any idea what to do about the [test failures](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2035454&view=logs&j=9f3a3068-7880-5994-433d-b92c1f9d379b&t=429853d2-592c-5f99-537b-6a3154c944ed&l=4545)?
> 
> As best I can tell, it looks like a (semver-compatible) update to @testing-library/dom or one of its dependencies has broken the behavior of `getByRole` in that the `title` attribute is no longer being considered when determining the accessible name. I'm not sure how relevant this is, but [this issue](https://github.com/A11yance/aria-query/issues/379#issuecomment-1227863154) on GitHub seems to suggest this may be intentional...
> 
> Tests are failing at:
> 
> https://github.com/iTwin/appui/blob/627e041687eac6787810a2c8be3d2cfb14f0106a/ui/components-react/src/test/properties/renderers/value/ArrayPropertyValueRenderer.test.tsx#L60
> 
> 
> https://github.com/iTwin/appui/blob/627e041687eac6787810a2c8be3d2cfb14f0106a/ui/components-react/src/test/properties/renderers/value/StructPropertyValueRenderer.test.tsx#L49
> 
> 
> https://github.com/iTwin/appui/blob/627e041687eac6787810a2c8be3d2cfb14f0106a/ui/core-react/src/test/button/UnderlinedButton.test.tsx#L30

Are spans not found in a query by role now? These are all implemented with spans.

wgoehrig:
> Are spans not found in a query by role now? These are all implemented with spans.
They are found w/ the role, but their accessible name no longer matches the `title` - just the text contents

GerardasB:
> @GerardasB @raplemie Any idea what to do about the [test failures](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2035454&view=logs&j=9f3a3068-7880-5994-433d-b92c1f9d379b&t=429853d2-592c-5f99-537b-6a3154c944ed&l=4545)?
> 
> As best I can tell, it looks like a (semver-compatible) update to @testing-library/dom or one of its dependencies has broken the behavior of `getByRole` in that the `title` attribute is no longer being considered when determining the accessible name. I'm not sure how relevant this is, but [this issue](https://github.com/A11yance/aria-query/issues/379#issuecomment-1227863154) on GitHub seems to suggest this may be intentional...
> 
> Tests are failing at:
> 
> https://github.com/iTwin/appui/blob/627e041687eac6787810a2c8be3d2cfb14f0106a/ui/components-react/src/test/properties/renderers/value/ArrayPropertyValueRenderer.test.tsx#L60
> 
> 
> https://github.com/iTwin/appui/blob/627e041687eac6787810a2c8be3d2cfb14f0106a/ui/components-react/src/test/properties/renderers/value/StructPropertyValueRenderer.test.tsx#L49
> 
> 
> https://github.com/iTwin/appui/blob/627e041687eac6787810a2c8be3d2cfb14f0106a/ui/core-react/src/test/button/UnderlinedButton.test.tsx#L30

Do not update package versions other than typescript? :)
I'm fine with using `byTitle` query directly..

PR #234, opened 2023-03-27
Refactored frontstage codemod

AntanasBuk:
- Rewrote `frontstage-to-config` transformer to better differentiate and reflect how each individual JSX element (and its attributes) is handled (including `Frontstage`, `StagePanel`, `Zone` and `Widget` elements).
- Extended cases where element attributes' possible value expressions have their respective handles.
- Transformations are now less destructive saving all attributes that are needed for constructing a new config property in case of failed handling process.
- Transformations are now also performed on `StagePanel`, `Zone` and `Widget` elements that exist outside of `Frontstage`.


common/changes/@itwin/appui-codemod/codemod-frontstage-copy-paste_2023-03-27-15-28.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-codemod",
+      "comment": "extended jsx element to config transformation handling for more complex cases",
```
GerardasB, line 5: This is used to generate CHANGELOG files, capitalize first letter.

tools/codemod/src/v4.0.0/__testfixtures__/frontstage-to-config/Frontstage1.output.tsx
```diff
@@ -4,8 +4,8 @@
 *--------------------------------------------------------------------------------------------*/
```
GerardasB, line 4: How is this different from `Frontstage.output.tsx`? File names should be descriptive as well.

tools/codemod/src/utils/TransformWidget.ts
```diff
@@ -0,0 +1,135 @@
...
+    .removeProperty("isToolSettings")
+    .removeProperty("isStatusBar")
+    .removeProperty("fillZone")
+    .removeProperty("syncEventIds")
+    .removeProperty("stateFunc")
+    .renameProperty("iconSpec", "icon")
+    .renameProperty("isFloatingStateSupported", "canFloat")
+    .renameProperty("badgeType", "badge")
+    .renameProperty("element", "content")
+    .renameProperty("allowedPanelTargets", "allowedPanels", (_path, property) => {
```
GerardasB, line 70: Very similar to `widget` transform, can we reuse this?

tools/codemod/src/utils/TransformWidget.ts
```diff
@@ -0,0 +1,135 @@
+import { JSCodeshift, ASTPath, JSXElement, MemberExpression, ObjectExpression, ObjectProperty, Collection } from "jscodeshift";
+import { ObjectExpressionCollection, useObjectExpression } from "./ObjectExpression";
+import { ConfigExpression, getObjectProperty, jsxElementToUniFormat, uniFormatToObjectExpression } from "./TransformUtils";
+
+export function transformWidget(j: JSCodeshift, widget: ASTPath<JSXElement>): ConfigExpression { // TODO: registerMethod
+  const uni = jsxElementToUniFormat(j, widget);
+  const widgetObj = uniFormatToObjectExpression(j, uni);
+
+  // Add ---------------------------
```
GerardasB, line 9: Remove these comments, most of the code should be self describing, w/ some inline comments for complicated cases.

tools/codemod/src/utils/TransformUtils.ts
```diff
@@ -0,0 +1,128 @@
...
+
+export type ExpressionKind = Identifier | FunctionExpression | ThisExpression | ArrayExpression | ObjectExpression | Literal | SequenceExpression | UnaryExpression | BinaryExpression | AssignmentExpression | MemberExpression | UpdateExpression | LogicalExpression | ConditionalExpression | NewExpression | CallExpression | ArrowFunctionExpression | YieldExpression | GeneratorExpression | ComprehensionExpression | ClassExpression | Super | TaggedTemplateExpression | TemplateLiteral | MetaProperty | AwaitExpression | ImportExpression | ChainExpression | OptionalCallExpression | OptionalMemberExpression | JSXIdentifier | JSXExpressionContainer | JSXElement | JSXFragment | JSXMemberExpression | JSXText | PrivateName | TypeCastExpression | DoExpression | BindExpression | ParenthesizedExpression | DirectiveLiteral | StringLiteral | NumericLiteral | BigIntLiteral | NullLiteral | BooleanLiteral | RegExpLiteral | Import | TSAsExpression | TSNonNullExpression | TSTypeParameter | TSTypeAssertion;
```
GerardasB, line 3: Code smell. Can we figure out a better way w/o redefining this type?

tools/codemod/src/utils/TransformStagePanel.ts
```diff
@@ -0,0 +1,207 @@
+import { JSCodeshift, ASTPath, JSXElement, ObjectProperty, SpreadProperty, Property, ObjectExpression } from "jscodeshift";
+import { concatExpressions } from "./ExpressionHandles";
+import { buildProperty, ConfigExpression, ExpressionKind, getObjectProperty, isExpressionKind, jsxElementToUniFormat } from "./TransformUtils"
+
+export function transformStagePanel(j: JSCodeshift, stagePanel: ASTPath<JSXElement>): ConfigExpression { // TODO: registerMethod
```
GerardasB, line 5: Functions that are v4.0.0 specific, shouldn't live under `utils`.

GerardasB (changes requested):

GerardasB:
> _ No description provided. _

PR usually contains a description.

tools/codemod/src/v4.0.0/__testfixtures__/frontstage-to-config/Frontstage1.output.tsx
AntanasBuk: It is based on filenames that already exist in the codebase (i.e. StandardFrontstageProvider, Frontstage1, Frontstage2 ...)

tools/codemod/src/utils/TransformStagePanel.ts
AntanasBuk: Also, I will rename `frontstage-to-config` to `element-to-config` if that's alright

tools/codemod/src/utils/TransformUtils.ts
AntanasBuk: As we discussed earlier, since `jscodeshift` uses `ast-types` dependency and does not expose `Kinds`, the only ways to redefine these that I can think of are: 1) use `ast-types` dependency ourselves 2) use lookup types on some loosely related `ast-type` interface that `jscodeshift` exposes (but imo that would smell even worse) 3) use `Expression` interface instead of `ExpressionKind`, but that would require casting it every time we use some `jscodeshift` builder method that requires parameter of type `ExpressionKind`

tools/codemod/src/utils/TransformWidget.ts
AntanasBuk: The only reusable part here is the construction of `canFloat` property anything else would break the use of fluent api here and just bring more confusion as to what attributes `Widget` transformer handles

GerardasB: All `Widget` properties are a subset of `WidgetConfig` properties, not only `canFloat`? Renames/removals are duplicated?

tools/codemod/src/utils/TransformStagePanel.ts
GerardasB: Reusable part should be in the `utils` directory, transformer specific part should be contained in the transfomer file itself.
Sure for the rename.

tools/codemod/src/utils/TransformUtils.ts
GerardasB: We've also discussed, that the need for these types indicates misuse of the indended `jscodeshift` API. One solution would be to use https://www.typescriptlang.org/docs/handbook/2/indexed-access-types.html, but ideally I'd like to get rid of these types altogether.

PR #245, opened 2023-04-03
Promote APIs in use by apps

NancyMcCallB:
Non-public APIs in use by apps promoted to public.


common/api/summary/core-react.exports.csv
```diff
@@ -157,7 +157,7 @@ public;RadialMenuProps
 public;RadialSizeType = ProgressRadialProps["size"]
 public;RatioChangeResult
 public;ReactMessage
-internal;Rectangle 
+public;Rectangle
```
GerardasB, line 160: Keep internal, use https://www.itwinjs.org/reference/core-geometry/cartesiangeometry/range2d/

common/api/summary/core-react.exports.csv
```diff
@@ -174,7 +174,7 @@ public;SettingsProvidersChangedEventArgs
 public;SettingsTabEntry
 public;SettingsTabsProvider
 internal;shallowDiffers: (a:
-internal;Size 
+public;Size
```
GerardasB, line 177: Keep internal, use inline object or https://www.itwinjs.org/reference/core-geometry/cartesiangeometry/xy/ for comparison.

common/api/summary/core-react.exports.csv
```diff
@@ -40,7 +40,7 @@ public;ContextMenuItemProps
 public;ContextMenuProps 
 public;ContextSubMenu 
 public;ContextSubMenuProps 
-internal;Corner
+public;Corner
```
GerardasB, line 43: Keep internal, this is only used with `Rectangle`.

GerardasB (changes requested):

common/api/summary/core-react.exports.csv
```diff
@@ -140,7 +140,7 @@ public;OmitChildrenProp
 public;Orientation
 public;OutsideClickEvent = PointerEvent | MouseEvent | TouchEvent
 internal;percentInRange(percent: number): number
-internal;Point 
+public;Point
```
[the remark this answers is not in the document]
raplemie: If we can't remove it right away, maybe we can document and point users to a public version of these ?

ui/appui-react/src/appui-react/UiFramework.ts
```diff
@@ -614,15 +614,15 @@ export class UiFramework {
     UiFramework.dispatchActionToStore(ConfigurableUiActionId.UseToolAsToolSettingsLabel, value, true);
   }
 
-  /** @alpha */
+  /** @public */
```
raplemie, line 617: Just so user dont have to search the setter method to understand the difference.
```suggestion
  /** When `true`, panels will close as soon as the mouse leave the panel.
   * When `false` (default), panels will close on next click outside the panel.
   * @public */
```

ui/appui-react/src/appui-react/configurableui/InternalConfigurableUiManager.ts
```diff
@@ -26,26 +26,26 @@ import { InternalModalDialogManager } from "../dialog/InternalModalDialogManager
 import { SyncUiEventDispatcher } from "../syncui/SyncUiEventDispatcher";
 
 /** Ui Activity Event Args interface.
- * @internal
+ * @public
```
raplemie, line 29: This file is not exported in the barrel, so even if they are marked public, they will not show up.
These events are part of the internal timers and the results of the timers are exposed in events like "onFronstageDeactivatedEvent". Do we really want to expose those ? (If so, we should move these events somewhere else as well).

NancyMcCallB: DR uses this in their testing. I'll check in with them.

common/api/summary/core-react.exports.csv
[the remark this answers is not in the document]
GerardasB: There is no public version, it's just utility classes for working with PointProps structure. Apps can use their own util libraries or use ones provided by core-geometry.

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -405,14 +405,14 @@ export class WidgetDef {
...
-   * @alpha
+   * @public
    */
   public show() {
     InternalFrontstageManager.onWidgetShowEvent.emit({ widgetDef: this });
   }
 
   /** Opens the widget and expands it to fill full size of the stage panel.
-   * @alpha
+   * @public
```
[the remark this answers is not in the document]
NancyMcCallB: Apps use them. You want to leave them alpha? There's not another way to do what they want to do.

common/api/summary/core-react.exports.csv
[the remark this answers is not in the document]
NancyMcCallB: Jason said he will refactor to use PointProps, so I've moved it back to internal.

raplemie (approved):

GerardasB (approved):

Issue #67, opened 2022-08-04
UI: Some FrontstageDef and WidgetDef operations only acts on Open widget


raplemie:
# WidgetDef
`show`: This operation only "bring to front" widget that are NOT hidden (Open -> make active, Closed -> Open and activate)
`expand`: Same behavior as show above, do not expand the full height of the panel as described.

# FrontstageDef
`floatWidget`: Will not do anything if widget is hidden, would expect either the widget to open and float, or if open after this has been called, to be shown floating.
`popupWidget`: Same as above

raplemie:
Needs verification, some part were done directly to adress this ("Show") but many refactor were done in the frontstage area, so it might not be an issue anymore.

raplemie:
All these issues are now resolved.

PR #249, opened 2023-04-04
Fix `widgetDef` `show` and `expand` methods

raplemie:
When `show` and `expand` are called, the widgets will now open if currently closed, as the JSDoc mentions.

With the new UI, the "other" widgets are never really closed, only to their minimum sizes, the `expand` method will now mimic the panel to be fully open, up to the limit of minimum sizes of the other widget.

part of #67


GerardasB (approved):

NancyMcCallB:
There should be something in NextVersion.ms about this change. While we see it as a bug, the number of times I've had consumers rely on a bug behavior as a feature is significant.

raplemie:
Should I complete it in 4.0 or we'd prefer waiting ? I feel if we keep adding to 4.0, we'll never release...

NancyMcCallB (approved):

PR #252, opened 2023-04-05
Remove 2.x and 3.0 deprecations from imodel-components-react

NancyMcCallB:
A few interfaces and components that were deprecated in 2.x and 3.0 were still in imodel-components-react. Removed them for the 4.0 release.


aruniverse (commented):
will everything marked deprecated in 3.x also be removed or?
if im not mistaken this was never answered in the chat we had

raplemie (approved):
LGTM

PR #253, opened 2023-04-05
Fix `StagePanelDef.pinned` not reflecting current component pin state correctly and related issues

AntanasBuk:
- Getting `StagePanelDef.pinned` now brings this value from its related React component.
- In standalone build from test-apps "Layout Info" widget now displays info about the changed state of the selected stage panel size and pin states immediately after the state updates.
- Wrote e2e tests to confirm this behavior.


ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
```diff
@@ -116,7 +125,32 @@ export class StagePanelDef extends WidgetHost {
...
+      const state = UiFramework.frontstages.activeFrontstageDef.getPanelCurrentState(this);
+      return state[2];
+    }
+    // istanbul ignore next
+    return false;
+  }
+
+  public set pinned(pinned: boolean) {
+    if (this._pinned === pinned)
+      return;
```
raplemie, line 139: Why is the getter relying on external state if the setter do not ?

If the internal state is not reliable for the getter, maybe the setter should not have an internal state at all ? (and rely on the getter)

AntanasBuk: A fair point. As I see it, the sole purpose of having this internal state is to confirm if the state's value has actually become out of sync with its external source (since setters are always called on all `StagePanelsDef`s whenever one of them changes) and needs to emit some onChange event. Additionally, I just reused code of `size` and `panelState`. So, if I were to rely only on external getter, I wouldn't know whether or not the onChange event should be emited.
A possible solution could be to shift responsibility of keeping tract of the previous `StagePanelsDef`s states onto code that triggers these setters (namely `FrontstageDef` as it already does so), but the code would get way more cumbersome and I wouldn't be sure if it would be the better pattern here since the internal state of all these props were already here even before my changes :).
Sooo... Any suggestions?

raplemie: You could have just said: Other setters are working like this in this class :D

Alright, that seems off but for this PR let stick to the standard of this file.

ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
```diff
@@ -100,9 +108,10 @@ export class StagePanelDef extends WidgetHost {
       const side = toPanelSide(this.location);
       frontstageDef.nineZoneState = setPanelSize(frontstageDef.nineZoneState, side, size);
       const panel = frontstageDef.nineZoneState.panels[side];
-      if (panel.size === this._size)
+      if (panel.size !== size) {
```
raplemie, line 111: If this is `true`, it is only because the `size` have been constrained by the max/min, I think we probably should send the change event anyway. (and not return right away)

common/api/appui-react.api.md
```diff
@@ -3539,6 +3541,14 @@ export class OpenMessageCenterEvent extends UiEvent<{}> {
 // @internal
 export function packNineZoneState(state: NineZoneState): SavedNineZoneState;
 
+// @public
+export interface PanelPinnedChangedEvent {
```
GerardasB, line 3545: Rename to `PanelPinnedChangedEventArgs`.

common/changes/@itwin/appui-react/stagepanel-pinned-fix_2023-04-05-15-17.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "Fix bug where pinned stage panel def prop is out of sync with its react element. Fix layout info widget not updating info on change stage panel def change.",
```
GerardasB, line 5: From these comments a changelog file is generated. Reword first sentence and instead of mentioning test-app specific layout info you could mention that an event is added to listen to pinned changes. You could also add multiple changes objects instead of adding everything into a single comment property.

full-stack-tests/ui/tests/stage-panel.test.ts
```diff
@@ -0,0 +1,54 @@
+/*---------------------------------------------------------------------------------------------
+* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+* See LICENSE.md in the project root for license terms and full copyright notice.
+*--------------------------------------------------------------------------------------------*/
+import { expect, Page, test } from '@playwright/test';
+import assert from 'assert';
+import { frontstageLocator, panelLocator, tabLocator, widgetLocator } from './Utils';
+
+
```
GerardasB, line 9: Spacing.

test-apps/appui-test-app/appui-test-providers/src/ui/widgets/LayoutWidget.tsx
```diff
@@ -50,11 +50,27 @@ function usePanelState(location: StagePanelLocation) {
...
+  const [pinned, setPinned] = React.useState(panelDef?.pinned);
+  React.useEffect(() => {
+    setPinned(panelDef?.pinned);
+  }, [panelDef]);
+  React.useEffect(() => {
+    const remove = UiFramework.frontstages.onPanelPinnedChangedEvent.addListener((e) => {
+      if (e.panelDef.location === location)
+        setPinned(e.pinned);
+    });
+    return remove;
```
GerardasB, line 64:
```suggestion
    return UiFramework.frontstages.onPanelPinnedChangedEvent.addListener((e) => {
      if (e.panelDef.location === location)
        setPinned(e.pinned);
    });
```

ui/appui-react/src/appui-react/framework/FrameworkFrontstages.ts
```diff
@@ -210,11 +210,16 @@ export interface FrameworkFrontstages {
   /** Get Widget State Changed event. */
   readonly onWidgetStateChangedEvent: WidgetStateChangedEvent;
 
-  /** Get Widget State Changed event.
+  /** Get Panel State Changed event.
```
GerardasB, line 213:
```suggestion
  /** Get panel state changed event.
```

ui/appui-react/src/appui-react/framework/FrameworkFrontstages.ts
```diff
@@ -210,11 +210,16 @@ export interface FrameworkFrontstages {
...
    * @alpha
    */
   readonly onPanelStateChangedEvent: PanelStateChangedEvent;
 
+  /** Get Panel Pinned Changed event.
```
GerardasB, line 218:
```suggestion
  /** Get panel pinned changed event.
```

ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
```diff
@@ -182,17 +216,17 @@ export class StagePanelDef extends WidgetHost {
 
     if (!config)
       return;
-    this._size = config.size;
+    this.size = config.size;
```
GerardasB, line 219: Assign to field to avoid firing change events when initializing?

```suggestion
    this._size = config.size;
```

ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
```diff
@@ -182,17 +216,17 @@ export class StagePanelDef extends WidgetHost {
...
     this._defaultSize = config.size;
     this._maxSizeSpec = config.maxSize;
     this._minSize = config.minSize;
     if (config.defaultState !== undefined) {
-      this._panelState = config.defaultState;
       this._defaultState = config.defaultState;
+      this.panelState = config.defaultState;
```
GerardasB, line 225: Assign to field to avoid firing change events when initializing?
```suggestion
      this._panelState = config.defaultState;
```

ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
```diff
@@ -182,17 +216,17 @@ export class StagePanelDef extends WidgetHost {
...
     }
     this._resizable = config.resizable ?? true;
     if (config.pinned !== undefined)
-      this._pinned = config.pinned;
+      this.pinned = config.pinned;
```
GerardasB, line 229: Assign to field to avoid firing change events when initializing?
```suggestion
      this._pinned = config.pinned;
```

GerardasB (changes requested):

ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
AntanasBuk: One of unit tests was not passing due to `ninezonestate` not being updated after setting pinned state (through passing a config) and calling a getter. But this was only after my changes, originally the test expected to receive the `pinned` value from the internal state. So I am not sure if the test should be removed since it tests something else now (and undo my changes here), or leave it as it is now?

ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
AntanasBuk: My code change wasn't correct here since I misunderstood how the previous code worked. I'll revert the changes. Also, I've consulted Gerardas on whether or not we should emit if the size after the resize did not change and he was against it, so for now I'll just return. Unless you know some use case or a common pattern that needs different behavior.

GerardasB (approved):

ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
```diff
@@ -102,7 +110,6 @@ export class StagePanelDef extends WidgetHost {
       const panel = frontstageDef.nineZoneState.panels[side];
       if (panel.size === this._size)
         return;
-      size = panel.size;
```
raplemie, line 105: The new size of the panel should be reported to the def, so this line should remain. (otherwise, if I say: `StagePanelDef.size = 10;` and the min value is 20, the panel will be set to 20, but the inner memory will be kept as if it was 10, which is not true.

ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
```diff
@@ -258,3 +290,13 @@ export const setPanelSize = produce((
   const panel = nineZone.panels[side];
   panel.size = size === undefined ? size : Math.min(Math.max(size, panel.minSize), panel.maxSize);
 });
+
+/** @internal */
+export const setPanelPinned = produce((
```
raplemie, line 295: Can you type this function and the `setPanelSize` one above ? 
```suggestion
export const setPanelPinned: (nineZone: NineZoneState, side: PanelSide, pinned: boolean) => NineZoneState = produce((
```

That should reduce the number of lines in the api .md file

ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
AntanasBuk: Good catch!

ui/appui-react/src/appui-react/stagepanels/StagePanelDef.tsx
AntanasBuk: Yup, that did the trick

PR #256, opened 2023-04-06
ViewSelector Fuzzy Search

kconner1:


ui/appui-react/src/appui-react/pickers/ViewSelector.tsx
```diff
@@ -247,10 +248,18 @@ export class ViewSelector extends React.Component<ViewSelectorProps, ViewSelecto
       });
 
       if(this._searchInput.length > 0){
-        views3dFiltered = views3d.filter((view) => view.name?.toLowerCase().includes(this._searchInput.toLowerCase()));
-        views2dFiltered = views2d.filter((view) => view.name?.toLowerCase().includes(this._searchInput.toLowerCase()));
-        sheetsFiltered = sheets.filter((view) => view.name?.toLowerCase().includes(this._searchInput.toLowerCase()));
-        unknownFiltered = unknown.filter((view) => view.name?. toLowerCase().includes(this._searchInput.toLowerCase()));
+        const views3dFuzzySearchResults = fuzzy.search(views3d, ["name"], this._searchInput).results as ListItem[];
```
NancyMcCallB, line 251: I notice that this searches the whole list with each successive letter added, instead of narrowing the search to the results from each letter. Is this intended?

kconner1: I am not sure I completely understand your question. From my perspective, we would want to search the whole list with each additional letter instead of just the filtered list from the previous letter. This is because if, for example, the user entered an incorrect letter for the first letter, then correct letters for the succeeding letters, and we only search the filtered list based on the first letter, no/incorrect results may appear.

Could you explain what you mean by "narrowing the search to the results from each letter"?

NancyMcCallB: I mean that as I type in the search box, if I type in "b" I get any entry with a "b". If I type "ba" and get any entry with a "b" OR an "a" when I would expect to get any entry with a "b" AND and "a." 
Maybe it's worth checking with UX?

NancyMcCallB (approved):

PR #261, opened 2023-04-11
Fix selection scope field

GerardasB:
This PR converts `SelectionScopeField` to a function component and fixes an issue with stale options of a `Select`.
Fallback labels defined in `selectionScopeLabels` are removed, since labels are provided by `presentation` layer.

### DIFF
--- a/docs/changehistory/NextVersion.md
+++ b/docs/changehistory/NextVersion.md
@@ -12,6 +12,7 @@ Table of contents:
 - [@itwin/appui-react](#itwinappui-react)
   - [Component changes](#component-changes)
   - [Static manager classes](#static-manager-classes)
+  - [SelectionScopeField](#selectionscopefield)
 - [@itwin/core-react](#itwincore-react)
   - [SCSS variables](#scss-variables)
 - [@itwin/components-react](#itwincomponents-react)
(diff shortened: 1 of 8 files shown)


saskliutas (approved):

docs/changehistory/NextVersion.md
```diff
@@ -12,6 +12,7 @@ Table of contents:
 - [@itwin/appui-react](#itwinappui-react)
   - [Component changes](#component-changes)
   - [Static manager classes](#static-manager-classes)
+  - [SelectionScopeField](#selectionscopefield)
```
raplemie, line 15: Should that be in `Component changes`, or should we have this title level for each components ?

ui/appui-react/src/appui-react/statusfields/SelectionScope.tsx
```diff
@@ -33,47 +32,39 @@ interface SelectionScopeFieldProps extends CommonProps {
  * in the Redux state.
  * @public
  */
-class SelectionScopeFieldComponent extends React.Component<SelectionScopeFieldProps> {
-  private _label = UiFramework.translate("selectionScopeField.label");
-  private _toolTip = UiFramework.translate("selectionScopeField.toolTip");
-  private _scopeOptions: SelectOption<string>[] = [];
+function SelectionScopeFieldComponent(props: SelectionScopeFieldProps) {
+  const [label] = React.useState(UiFramework.translate("selectionScopeField.label"));
+  const [toolTip] = React.useState(UiFramework.translate("selectionScopeField.toolTip"));
```
raplemie, line 37: There is no reason to have `useState` here since these are strings, and the `translate` is always called but dropped (which means that if an application ever implement live language switch, the language switch will be missed)

GerardasB: Changed for consistency.

PR #262, opened 2023-04-11
Add `defaultTool` prop to `FrontstageConfig`

GerardasB:
This PR adds `defaultTool` prop to `FrontstageConfig` and `StandardFrontstageProps`.

### DIFF
--- a/ui/appui-react/src/appui-react/frontstage/FrontstageConfig.ts
+++ b/ui/appui-react/src/appui-react/frontstage/FrontstageConfig.ts
@@ -10,6 +10,7 @@ import { CommonProps } from "@itwin/core-react";
 import { ContentGroup, ContentGroupProvider } from "../content/ContentGroup";
 import { StagePanelConfig } from "../stagepanels/StagePanelConfig";
 import { WidgetConfig } from "../widgets/WidgetConfig";
+import { ToolItemDef } from "../shared/ToolItemDef";
 
 /** Configuration from which a frontstage is created.
  * @public
@@ -26,6 +27,8 @@ export interface FrontstageConfig extends CommonProps {
    * Increasing the value will make sure to reinitialize App layout instead of restoring to old layout.
    */
   readonly version: number;
+  /** Tool that is started once the Frontstage is activated. */
+  readonly defaultTool?: ToolItemDef;
 
   /** The top-left corner that shows tools typically used to query and modify content. */
   readonly contentManipulation?: WidgetConfig;
--- a/ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
+++ b/ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
@@ -75,6 +75,7 @@ export class FrontstageDef {
   private _contentGroupProvider?: ContentGroupProvider;
   private _floatingContentControls?: ContentControl[];
   private _savedWidgetDefs?: SavedWidgets;
+  private _appDefaultToolId?: string;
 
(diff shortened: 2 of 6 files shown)


ui/appui-react/src/appui-react/frontstage/FrontstageConfig.ts
```diff
@@ -26,6 +27,8 @@ export interface FrontstageConfig extends CommonProps {
    * Increasing the value will make sure to reinitialize App layout instead of restoring to old layout.
    */
   readonly version: number;
+  /** Tool that is started once the Frontstage is activated. */
```
NancyMcCallB, line 30: Original comment:
  /** The defaultTool is is started when then frontstage loads and whenever any other tools exit.
   * Most of the time, this is the Element Selection Tool (CoreTools.selectElementCommand).
   * Your app can specify its own tool or another core tool as default with this property.
   */

It might be worth indicating that it should be a primitive tool

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
```diff
@@ -253,7 +254,12 @@ export class FrontstageDef {
   }
 
   /** Handles when the Frontstage becomes inactive */
-  protected async _onDeactivated() { }
+  protected async _onDeactivated() {
+    if (this._appDefaultToolId) {
+      IModelApp.toolAdmin.defaultToolId = this._appDefaultToolId;
+      this._appDefaultToolId = undefined;
```
NancyMcCallB, line 260: This naming is confusing. appDefaultTool seems to me to be the one defined by the app, not the one that core uses by default.

GerardasB: Renamed to `_toolAdminDefaultToolId`

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
```diff
@@ -316,7 +322,20 @@ export class FrontstageDef {
...
+
+    const defaultTool = this._initialConfig.defaultTool;
+    if (!defaultTool) {
+      void IModelApp.toolAdmin.startDefaultTool();
+      return;
+    }
+
+    this._toolAdminDefaultToolId = IModelApp.toolAdmin.defaultToolId;
+    IModelApp.toolAdmin.defaultToolId = defaultTool.toolId;
+    defaultTool.execute();
```
raplemie, line 337: Instead of requiring a full tool def, could we only require the tool id ? (Set the Id then startDefaultTool?)

GerardasB: Current ToolItemDef allows for custom handlers which support i.e. tool arguments, need to verify if defaultToolId would be enough for now.

jason-crow: AFAIK all the defaultTool's we set are registered and could be invoked by just the toolId

I guess you could do a 
```typescript
defaultTool?: ToolItemDef | string;
```
Then you could have it both ways

GerardasB: For now I've changed the type to `defaultTool?: string;`. Later we can add a union type w/ tool arguments if needed.

raplemie: The change was not pushed.

NancyMcCallB (approved):

PR #268, opened 2023-04-12
Fix duplicate specifiers

GerardasB:
This PR fixes an issue where duplicate specifiers are added by one of transformers failing subsequent transformers.


raplemie (approved):
Seems Ok, but not familiar with anything in there.

PR #274, opened 2023-04-14
Handle abstract UiItemsManager

raplemie:
Update the `@itwin/appui-react ` `UiItemsManager` to include any providers registered to the `@itwin/appui-abstract` `UiItemsManager` while attempting to not force the abstract version to exist (so we can remove it later without breaking...

### DIFF
--- a/ui/appui-react/src/appui-react/ui-items-provider/UiItemsManager.ts
+++ b/ui/appui-react/src/appui-react/ui-items-provider/UiItemsManager.ts
@@ -16,6 +16,7 @@ import { UiFramework } from "../UiFramework";
 import { Widget } from "../widgets/Widget";
 import { ProviderItem } from "./ProviderItem";
 import { UiItemsProvider } from "./UiItemsProvider";
+import { AbstractUiItemsManagerHook } from "./AbstractUiItemsManagerHook";
 
 /** UiItemsProvider register event args.
  * @beta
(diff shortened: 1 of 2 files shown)


ui/appui-react/src/appui-react/ui-items-provider/UiItemsManager.ts
```diff
@@ -220,7 +224,7 @@ export class UiItemsManager {
    * @returns An array of widgets.
    */
   public static getWidgets(stageId: string, stageUsage: string, location: StagePanelLocation, section?: StagePanelSection): ReadonlyArray<ProviderItem<Widget>> {
-    const widgets: ProviderItem<Widget>[] = [];
+    const widgets: ProviderItem<Widget>[] = this._abstractHook.getWidgets(stageId, stageUsage, location, section);
```
GerardasB, line 227: Should we just fallback to using **abstract** `UiItemsManager` if one is defined in `appui-abstract`?
I.e. would following work with current impl?
1. `appui-abstract` 3.7 registers a widget **B**.
2. `appui-react` 4.0 registers a widget **B**.
3. Calling `getWidget` in both `appui-react` and `appui-abstract` should return [**A**, **B**].

raplemie: We can do that as well, but we'll need to handle conversion both ways (When using React, we'll need to convert from and to Abstract, and if it does not exist, not conversion is needed)

raplemie:
Handled by #278

PR #278, opened 2023-04-18
Handle abstract UiItemsManager

GerardasB:


ui/appui-react/src/appui-react/ui-items-provider/AbstractUiItemsManager.ts
```diff
@@ -0,0 +1,548 @@
...
+    case AbstractStagePanelLocation.Left:
+      return StagePanelLocation.Left;
+    case AbstractStagePanelLocation.Top:
+      return StagePanelLocation.Top;
+    case AbstractStagePanelLocation.TopMost:
+      return StagePanelLocation.Top;
+    case AbstractStagePanelLocation.Bottom:
+      return StagePanelLocation.Bottom;
+    case AbstractStagePanelLocation.BottomMost:
+      return StagePanelLocation.Bottom;
```
raplemie, line 495: `fromAbstractStagePanelLocation` is only used when requesting widgets to an adapted UiItemsProvider, it should not return anything for TopMost and BottomMost because if every section is requested, the widget will return in both places. (This is linked to my other comment about the adapted UiItemsManager...)

Same reflexion goes to `fromAbstractStagePanelSection` below.

ui/appui-react/src/appui-react/ui-items-provider/AbstractUiItemsManager.ts
```diff
@@ -0,0 +1,548 @@
...
+    return items;
+  }
+
+  public provideWidgets(stageId: string, stageUsage: string, location: StagePanelLocation, section?: StagePanelSection) {
+    if (!this._adaptee.provideWidgets)
+      return [];
+
+    const abstractLocation = toAbstractStagePanelLocation(location);
+    const abstractSection = section === undefined ? undefined : toAbstractStagePanelSection(section);
+    const abstractWidgets = this._adaptee.provideWidgets(stageId, stageUsage, abstractLocation, abstractSection);
```
raplemie, line 250: Similar to the UiItemsManagerAdapter, if the Start section is requested from this adapter, it would need to also return the Middle section. (Although, I'm not sure I would understand the point of pulling out a provider and then call back its getter functions, this is what this try to provide)

Same thing for location.

raplemie (commented):
Depending on how thorough we want to go with this, given the starting edge case that this whole thing feels already...

ui/appui-react/src/appui-react/ui-items-provider/AbstractUiItemsManager.ts
GerardasB: In that case it would not be possible to query widgets from appui-react provider that were registered as TopMost via abstract provider? We should be handling duplicate widgets and I believe it's not a huge problem to return the same widgets, similar to when the widget provider ignores one of the arguments..

ui/appui-react/src/appui-react/ui-items-provider/AbstractUiItemsManager.ts
```diff
@@ -0,0 +1,548 @@
...
+      };
+    });
+    return items;
+  }
+
+  public getWidgets(stageId: string, stageUsage: string, location: StagePanelLocation, section?: StagePanelSection | undefined): readonly ProviderItem<Widget>[] {
+    const abstractLocation = toAbstractStagePanelLocation(location);
+    const abstractSection = section === undefined ? undefined : toAbstractStagePanelSection(section);
+    const abstractWidgets = this._adaptee.getWidgets(stageId, stageUsage, abstractLocation, abstractSection);
+    const widgets = abstractWidgets
```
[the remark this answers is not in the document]
GerardasB: That's a good point, will try and work it out.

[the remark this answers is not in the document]
GerardasB: Fixed.

ui/appui-react/src/appui-react/ui-items-provider/AbstractUiItemsManager.ts
raplemie: appui-react provider would not be answering to TopMost as it is not in the Enum they are built for. But you are right that we are probably already handling duplicates, so that is probably not a problem.

ui/appui-react/src/appui-react/ui-items-provider/AbstractUiItemsManager.ts
GerardasB: This is similar to Location, so the same argument applies.

raplemie: This is actually different, because this class is adapting an AbstractUiItemsProvider (which might only return values if `Middle` is requested) to a UiItemsProvider, which will never be requested to return such widget.

GerardasB: I'm sorry, this was similar to `AbstractUiItemsManagerAdapterm`, fixed.

raplemie (approved):
LGTM!

PR #281, opened 2023-04-19
Remove 3.7 deprecations

NancyMcCallB:
Remove remaining 3.x deprecations.


raplemie (approved):
Would probably be a good idea to rename the Dialog files to `*DialogRenderer` as this is the only thing remaining in these files, but since the files are quite different, I wonder if Git will get that the file are a rename...

NancyMcCallB:
> Would probably be a good idea to rename the Dialog files to `*DialogRenderer` as this is the only thing remaining in these files, but since the files are quite different, I wonder if Git will get that the file are a rename...

I decided not to do that this late in the release cycle. This change was already more disruptive than I expected. Moving the source files around in the package isn't a breaking change, so I decided to wait.

PR #284, opened 2023-04-21
Add `ErrorBoundary` to widget content renderer

AntanasBuk:
- Errors thrown during rendering of widget content are now caught by `ErrorBoundary`. A fallback UI is rendered instead.
- Added a button to test-apps `Layout Controls` widget that throws a simulated error (during rendering) when clicked.
- Wrote an e2e test.


ui/appui-react/package.json
```diff
@@ -132,11 +132,13 @@
...
     "@itwin/itwinui-variables": "^2.0.0",
     "@itwin/itwinui-icons-react": "^2.1.0",
+    "@itwin/itwinui-illustrations-react": "^2.0.1",
     "@itwin/itwinui-icons": "^1.14.0",
     "classnames": "2.3.1",
     "immer": "9.0.6",
     "lodash": "^4.17.10",
-    "rxjs": "^6.6.2"
+    "rxjs": "^6.6.2",
+    "react-error-boundary": "4.0.3"
```
aruniverse, line 141: do we need to pull in a separate pkg for this? 
this is what we do in the viewer https://github.com/iTwin/viewer/blob/master/packages/modules/viewer-react/src/components/error/ErrorBoundary.tsx

GerardasB: I think we can stick with this instead of defining our own ErrorBoundary, just make sure to sort the dependencies alphabetically in package.json

test-apps/appui-test-app/appui-test-providers/src/ui/widgets/LayoutWidget.tsx
```diff
@@ -486,6 +486,20 @@ function SelectWidgetControls() {
   );
 }
 
+function WidgetContentThrowError() {
+  const [shouldThrow, setShouldThrow] = React.useState(false);
+  if (shouldThrow) {
+    setShouldThrow(false);
```
GerardasB, line 492: Why is this needed?

PR #300, opened 2023-04-26
Docs - TypeScript 5.0 support

ben-polinsky:
Use typedoc 0.23 and TS 5 for generating docs.


common/config/rush/pnpm-lock.yaml
```diff
@@ -1,5 +1,10 @@
 lockfileVersion: 5.3
 
+overrides:
```
raplemie, line 3: Where are these comming from ? Were they added manually ? Will they get wiped next time we `rush update` ?

ben-polinsky: Was a first attempt. They should be gone on update, I can do that and test everything works.

ben-polinsky: Gone.

ben-polinsky: Was an issue with the validation in core... should hopefully be fixed now... let's see

raplemie (approved):
LGTM

raplemie:
@mergifyio backport release/4.0.x

Issue #259, opened 2023-04-10
View cube’s rotation is different in up/down and left/right movements


muthubentley:
View cube’s rotation is different in up/down and left/right movements
When left/right arrows are clicked, The rotation of navcube is simple. If started from Front and right arrow is clicked Front-->Right-->Back-->Left-->Front . 
Whereas if up/down arrows are clicked , The rotation gets inversed once in two time. If start from Front and down arrow is clicked Front-->Bottom-->Back-->Bottom-->Front . Top will never be displayed
This can be seen in any sandboxes. Is this the right behaviour. Is there any background for this?

muthubentley:
@NancyMcCallB  @GerardasB  Request your attention on this issue.

muthubentley:
@NancyMcCallB 
I understand this is targeted for 4.1 release, but user just wants to know what is the background of this behaviour. Is there any justification?

Issue #258, opened 2023-04-10
Tooltip not displayed on arrows around the cube


muthubentley:
Tooltip not displayed on arrows around the cube

**Steps To Reproduce**
Long press on the cube, 4 arrows will be displayed around the cube
When mouse is hovered over the arrow buttons on the cube, tooltip of the position is displayed.
 
![image](https://user-images.githubusercontent.com/107229931/230806894-f2df5858-d233-423d-ad35-280fae7e21d7.png)

Whereas if the cube is facing back, the tooltip doesn’t get display for Up and Down arrow
 
![image](https://user-images.githubusercontent.com/107229931/230806903-41da3c42-0959-4301-8a26-fecb57398ea6.png)

**Expected behavior**
Tooltip should be displayed in all positions.

muthubentley:
@NancyMcCallB  @GerardasB  Request your attention on this issue.

PR #311, opened 2023-05-08
Navigation  Cube Bugs

NancyMcCallB:
Fixes: 
https://github.com/iTwin/appui/issues/259 by bypassing special "heads-up" processing for rotations. A new prop "favorHeadsUpRotation" to the CubeNavigationAid will bypass these changes, in case someone relies on the behavior.

https://github.com/iTwin/appui/issues/258 by explicitly comparing to the Z axis instead of the whole rotation matrix.


common/changes/@itwin/imodel-components-react/nav-cube-bugs_2023-05-08-16-48.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/imodel-components-react",
+      "comment": "Change arrow-click rotation logic to allow toation to rotated faces.",
```
raplemie, line 5:
```suggestion
      "comment": "Change arrow-click rotation logic to allow rotation to rotated faces.",
```

ui/imodel-components-react/src/imodel-components-react/navigationaids/CubeNavigationAid.tsx
```diff
@@ -100,6 +100,7 @@ export interface CubeNavigationAidProps extends CommonProps {
   onAnimationEnd?: () => void;
   /** @internal */
   animationTime?: number;
+  favorHeadsUpRotation?: boolean;
```
raplemie, line 103: Should this be exposed in some way to the `CubeNavigationAidControl` or users already directly use the react component only ?

NancyMcCallB: I think we can wait until someone asks for it. I left the old logic in with a prop in case someone is relying on it, but no reason to propagate that unless it's required.

raplemie (approved):
LGTM!

PR #317, opened 2023-05-12
Reference UiItemsProvider from appui-react

GerardasB:
## Changes

Reference `UiItemsProvider` from `appui-react`.

## Testing

N/A


GerardasB:
Prerequisite for https://github.com/iTwin/itwinjs-core/pull/5503

PR #320, opened 2023-05-16
Handle CSS changes by codemod tool

GerardasB:
## Changes

This PR adds CSS change handling to appui-codemod tool.

## Testing

Manual:
1. Packed the appui-codemod tool
2. Installed appui-codemod globally
3. Ran the tool on test-apps directory (3.x version)

Added new unit tests to appui-codemod package.


tools/codemod/codemod.ts
```diff
@@ -48,6 +60,8 @@ async function run(codemod: string, paths: string[], debug: boolean, optionOverr
   };
   debug && console.log("Running 'jscodeshift' with following configuration:", { transformFile, paths, options });
   await jscodeshift(transformFile, paths, options);
+
+  !skipCss && await runCssCodemod(args, transformFile);
```
raplemie, line 64: If both are promises and works on different files, could they be run at the same time ?

raplemie (commented):
Maybe something that we want to look into later, but should it remove `@imports` as well ?

tools/codemod/codemod.ts
GerardasB: We probably could, but let's keep them separated, easier to navigate output.

tools/codemod/src/v4.0.0/full.ts
```diff
@@ -39,3 +41,72 @@ export default function transformer(file: FileInfo, api: API, options: Options)
...
+
+function renameValue(from: string, to: string): Plugin {
+  return {
+    postcssPlugin: `rename-value ${from} to ${to}`,
+    Declaration(decl) {
+      const parsed = valueParser(decl.value);
+      parsed.walk((node) => {
+        if (node.value === from)
+          node.value = to;
+        if (node.value === `#{${from}}`)
```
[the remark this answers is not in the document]
GerardasB: Variables inside interpolation should get renamed now.

GerardasB:
> Maybe something that we want to look into later, but should it remove `@imports` as well ?

Do you think we should get rid of all SCSS @imports from a specific package? Or just a list of specific partials? Otherwise, we'd have to track all variables/functions defined in the partial.

raplemie:
> Do you think we should get rid of all SCSS @imports from a specific package? Or just a list of specific partials? Otherwise, we'd have to track all variables/functions defined in the partial.

Yes, I'm not sure how we could handle that, we can certainly remove direct imports to file that no longer exists (the ones removed in #204, for example) but I'm not sure if we can make more than this with confidence without making it much more intelligent.

Maybe we could report updated files that were removed variables but still have imports and leave it to the user ?

tools/codemod/src/v4.0.0/full.ts
```diff
@@ -39,3 +41,140 @@ export default function transformer(file: FileInfo, api: API, options: Options)
...
+    }
+  };
+}
+
+function toModulePaths(getPath: (module: string) => string) {
+  return ["cjs", "esm"].map((module) => getPath(module));
+}
+
+const importPathsToRemove = [
+  ...toModulePaths((module) => `~@itwin/core-react/lib/${module}/core-react/button/blue-large`),
```
raplemie, line 84: I think the `~` may not always be there depending on the configuration (it used to be always there, but I think it is no longer necessary, so some may not have it)

GerardasB: Fixed.

raplemie (approved):
Should be helpful!!! :+1:

PR #323, opened 2023-05-18
Add SelectionCountField component

GerardasB:
## Changes

This PR adds a `SelectionCountField` component. This controlled component can be used instead of an existing redux-connected `SelectionInfoField`.

## Testing

Updated unit tests.
Ran the standalone test app to verify.

### DIFF
--- a/ui/appui-react/src/appui-react/statusfields/SelectionInfo.tsx
+++ b/ui/appui-react/src/appui-react/statusfields/SelectionInfo.tsx
@@ -8,50 +8,26 @@
 
 import * as React from "react";
 import type { CommonProps } from "@itwin/core-react";
-import type { IModelConnection } from "@itwin/core-frontend";
 import { SelectionCountField } from "./SelectionCount";
-import { useActiveIModelConnection } from "../hooks/useActiveIModelConnection";
-
-/** Arguments for [[useSelectionSetSize]] hook.
- * @beta
- */
-export interface UseSelectionSetSizeArgs {
-  iModel: IModelConnection | undefined;
-}
-
-/** React hook that returns element count of a selection set.
- * @beta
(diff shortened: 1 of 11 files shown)


ui/appui-react/src/appui-react/statusfields/SelectionInfo.tsx
```diff
@@ -6,60 +6,57 @@
...
-
-// we declare the variable and export that rather than using export default.
 /**
  * SelectionInfo Status Field React component. This component is designed to be specified in a status bar definition.
  * It is used to display the number of selected items based on the Presentation Rules Selection Manager.
  * This React component is Redux connected.
  * @public
- */ // eslint-disable-next-line @typescript-eslint/naming-convention
-export const SelectionInfoField = connect(mapStateToProps)(SelectionInfoFieldComponent);
+ * @deprecated in 4.0. Use [[SelectionCountField]] with [[useSelectionSetSize]] instead.
```
raplemie, line 50: 1. Should we wait for the other to be out of beta (tested and validated as useful) before deprecating this one ?
2. Also, the documentation mentions that this is based on Presentation Rules Selection Manager, I think the issue is that this is no longer true, so we might want to update the documentation.
3. Documentation says that this is Redux connected, but it's no longer, should this component keep the same source of information which must be updated by applications now, and offer the other component as an alternative? The `numItemsSelected` state value is also used in 2 other functions, so it should probably remain updated...

GerardasB: Removed the deprecation. Reverted the changes to keep SelectionInfoField redux connected. Updated the docs.

raplemie (approved):
LGTM!

PR #329, opened 2023-05-23
Fix stage panel transition

GerardasB:
## Changes

Correctly handle `onTransitionEnd` events to fix stage panel transition from ending before actual duration.

## Testing

Ran in the standalone test app.

### DIFF
--- a/ui/appui-layout-react/src/appui-layout-react/widget-panels/Panel.scss
+++ b/ui/appui-layout-react/src/appui-layout-react/widget-panels/Panel.scss
@@ -28,7 +28,7 @@
   }
 
   &.nz-transition {
-    transition-duration: 0.3s;
+    transition-duration: var(--iui-duration-2);
     transition-timing-function: ease-in;
   }
 
@@ -80,7 +80,7 @@
       left: 0;
 
       >.nz-grip.nz-grip {
-        right: calc(#{$nz-grip-width} * -0.5);;
+        right: calc(#{$nz-grip-width} * -0.5);
         transform: translate(0, -50%);
       }
     }
@@ -102,7 +102,8 @@
           }
         }
       }
-      > .nz-grip-container > .nz-grip {
+
+      >.nz-grip-container>.nz-grip {
         right: 0
       }
     }
@@ -135,7 +136,7 @@
         }
       }
 
-      > .nz-grip-container > .nz-grip {
+      >.nz-grip-container>.nz-grip {
         left: 0;
       }
     }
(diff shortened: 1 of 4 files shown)


ui/appui-layout-react/src/appui-layout-react/widget-panels/Panel.tsx
```diff
@@ -233,36 +143,19 @@ export function WidgetPanel() {
         className={className}
         ref={ref}
         style={style}
-        onTransitionEnd={handleTransitionEnd}
+        onTransitionEnd={(e) => {
+          if (e.target !== elementRef.current)
+            return;
+
+          handleTransitionEnd();
```
raplemie, line 150: Shouldnt that condition be in `handleTransitionEnd` itself as it gets the same `ref` anyway ?

ui/appui-layout-react/src/appui-layout-react/widget-panels/PanelSections.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const panel = useLayout((state) => {
+    const p = state.panels[side];
+    const span = isHorizontalPanelState(p) ? p.span : false;
+    return {
+      collapsed: p.collapsed,
+      splitterPercent: p.splitterPercent,
+      widgets: p.widgets,
+      resizable: p.resizable,
+      span,
+    };
```
raplemie, line 109: While a copy of the other version, most member are not used in this context.
```suggestion
    const span = isHorizontalPanelState(p) ? p.span : false;
    return {
      splitterPercent: p.splitterPercent,
      widgets: p.widgets,
    };
```

ui/appui-layout-react/src/appui-layout-react/widget-panels/Panel.scss
```diff
@@ -102,7 +102,8 @@
           }
         }
       }
-      > .nz-grip-container > .nz-grip {
+
+      >.nz-grip-container>.nz-grip {
```
FlyersPh9, line 106: Do we have Stylelint or Prettier setup that could catch these for us?

GerardasB: Not ATM. We have 83 hits for `>.` and 215 hits for `> .` in our `*.scss` files. I'd prefer the variant w/ more spacing.
We could achieve more consistency by setting `scss.format.spaceAroundSelectorSeparator` in workspace settings?

GerardasB: Resolving this for now. I've opened https://github.com/iTwin/appui/issues/331 instead

PR #342, opened 2023-05-31
Add prettier

GerardasB:
## Changes

Add https://prettier.io/ to enforce consistent code formatting in the repository.
Code formatting is enforced by running the prettier in the CI.

This PR adds `rush prettier` command to check for formatting inconsistencies and `rush prettier:fix` to fix the inconsistencies.

It is recommended to install a `prettier` VSCode extension for automatic code formatting.

## Testing

N/A

PR #341, opened 2023-05-31
Prettier fix

GerardasB:
## Changes

The initial run of `rush prettier:fix`.

## Testing

N/A

PR #346, opened 2023-06-02
Clean unused lint comment as well

raplemie:
## Changes

This is removing most useless lint comment we have in the code base that I could find, on top of the prettier fix. The idea is that the prettier fix merge will already affect a ton of files, and this is good to be cleaned as well, and would create another "cleanup" PR when reviewing file history. Having them combined would reduce that history walk, however I'm open to discussion if we think that this could be harmful. (Hence the PR -> PR rather than a commit on the existing branch)

## Testing

Ran `rush lint` without any error locally, will monitor if the CI picks up anything else.

### DIFF
--- a/test-apps/appui-test-app/appui-test-providers/src/store/index.ts
+++ b/test-apps/appui-test-app/appui-test-providers/src/store/index.ts
@@ -33,7 +33,6 @@ const initialState = {
   showCustomViewOverlay: false,
 } as TestProviderState;
 
-// eslint-disable-next-line @typescript-eslint/naming-convention
 export const TestProviderSliceName = "testProviderState";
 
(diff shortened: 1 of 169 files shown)


GerardasB:
No worries, we can merge this in first.
The target branch of https://github.com/iTwin/appui/pull/341 (which already addresses some of the useless lint ignores) is a branch of https://github.com/iTwin/appui/pull/342 so we will end up with a single commit for adding prettier.
I'll just re-run prettier and fix the residual lint errors instead of trying to merge this.

GerardasB (approved):

raplemie:
The branch was taken from *prettier-fix*, so there shouldnt be anything to merge/fix on top of it. (Maybe between prettier-fix and add-prettier though.)

PR #347, opened 2023-06-05
Clear rule instead of removing it when there is only one rule

simasjar:
## Changes
Clear rule instead of removing it when only one rule is left in the rule group.

## Testing
Added two tests for checking if the rule and rule group will change to empty rule when removing it and they are only one rule left in the rule group. Also, removed the test for checking if the rule group is removed when the last rule is removed because with the current change rule group will always have at least one rule.


GerardasB (approved):

ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
```diff
@@ -97,9 +97,17 @@ export class PropertyFilterBuilderActions {
...
       if (itemIndex === -1)
         return;
+      if (parentGroup.items.length === 1) {
+        if (isPropertyFilterBuilderRuleGroup(parentGroup.items[0]))
+          parentGroup.items[0] = createEmptyRule(parentGroup.id);
+        else {
+          parentGroup.items[0].operator = undefined;
+          parentGroup.items[0].value = undefined;
+          parentGroup.items[0].property = undefined;
+        }
```
saskliutas, line 107:
```suggestion
        parentGroup.items[0] = createEmptyRule(parentGroup.id)
```

ui/components-react/src/test/filter-builder/FilterBuilderState.test.ts
```diff
@@ -116,14 +116,52 @@ describe("usePropertyFilterBuilderState", () => {
     const { result } = renderHook(() => usePropertyFilterBuilderState());
     const { actions } = result.current;
 
+    let rootGroup = result.current.state.rootGroup;
+    rootGroup.items.push({ id: "testId", groupId: "groupId" });
```
saskliutas, line 120: You shouldn't mutate object stored in React state. Use `actions` to setup state for tests.

saskliutas (changes requested):

ui/components-react/src/test/filter-builder/FilterBuilderState.test.ts
simasjar: fixed

ui/components-react/src/components-react/filter-builder/FilterBuilderRule.tsx
```diff
@@ -72,9 +73,11 @@ export function PropertyFilterBuilderRuleRenderer(props: PropertyFilterBuilderRu
...
   return <div className="rule">
     <div className="rule-remove-action">
-      <IconButton onClick={removeRule} styleType="borderless" size="small">
-        <SvgDelete />
-      </IconButton>
+      {property || !isOnlyRuleInRuleGroup
+        ? <IconButton onClick={removeRule} styleType="borderless" size="small">
+          <SvgDelete />
+        </IconButton>
+        : undefined}
```
saskliutas, line 80:
```suggestion
        : null}
```

ui/components-react/src/components-react/filter-builder/FilterBuilderRule.tsx
```diff
@@ -28,14 +28,15 @@ export interface PropertyFilterBuilderRuleRendererProps {
   path: string[];
   /** Rule to render. */
   rule: PropertyFilterBuilderRule;
+  isOnlyRuleInRuleGroup?: boolean;
```
saskliutas, line 31:
```suggestion
  isOnlyRuleInGroup?: boolean;
```

saskliutas (changes requested):

ui/components-react/src/components-react/filter-builder/FilterBuilderRule.tsx
```diff
@@ -72,9 +73,11 @@ export function PropertyFilterBuilderRuleRenderer(props: PropertyFilterBuilderRu
 
   return <div className="rule">
     <div className="rule-remove-action">
-      <IconButton onClick={removeRule} styleType="borderless" size="small">
-        <SvgDelete />
-      </IconButton>
+      {property || !isRemovable
```
saskliutas, line 76: `!isRemoveable` looks incorrect. Remove button will be shown when `isRemoveable` is set to `false`.

saskliutas (changes requested):

ui/components-react/src/components-react/filter-builder/FilterBuilderRule.tsx
simasjar: fixed

saskliutas (approved):

PR #352, opened 2023-06-07
Internal api fixes

NancyMcCallB:
## Changes

Removed use of internal itwin.js internal APIs from source code.
Note: alpha and beta APIs are still in use where there are no fully public alternatives.

## Testing

Ran coverage testing to execute all unit tests.


aruniverse:
Did a quick pass didnt see any internal api usage? Most of it was just eslint disable deprecation comments?
Shouldnt this pr fold in changes from https://github.com/iTwin/appui/pull/198 now?

ui/core-react/src/core-react/messagebox/MessageBox.tsx
```diff
@@ -80,6 +81,7 @@ export class MessageBox extends React.PureComponent<MessageBoxProps> {
     modal: true,
   };
 
+  // eslint-disable-next-line deprecation/deprecation
   public override render(): JSX.Element {
```
GerardasB, line 85: Can't we just drop the explicit types from the `render` of class components instead of ignoring the deprecation?

ui/core-react/src/core-react/messagebox/MessageBox.tsx
```diff
@@ -40,6 +40,7 @@ export interface MessageBoxProps extends CommonProps {
   /** List of [[DialogButtonDef]] objects specifying buttons and associated onClick events */
   buttonCluster: DialogButtonDef[];
   /** Title to show in title bar of dialog  */
+  // eslint-disable-next-line deprecation/deprecation
   title?: string | JSX.Element;
```
GerardasB, line 44: Should we replace it with `React.ReactElement` or `React.JSX.Element` instead of ignoring it?

GerardasB (changes requested):
Let's try not to ignore any of JSX's deprecations.

ui/core-react/src/core-react/messagebox/MessageBox.tsx
NancyMcCallB: > Did a quick pass didnt see any internal api usage? Most of it was just eslint disable deprecation comments? Shouldnt this pr fold in changes from #198 now?

Most of the internal APIs we were using were promoted in core. The changes are primarily in the test apps, getting rid of our reliance on MockRender. 
We should only fold in #198 if we can turn it off for the core-3x variant. Also, if it flags alpha and beta APIs it will cause too much noise.

NancyMcCallB: > Should we replace it with `React.ReactElement` or `React.JSX.Element` instead of ignoring it?

The error said to replace with React.JSX.Element, but that would not compile in the 3.x variant. I'll try React.ReactElement.

GerardasB (approved):

PR #356, opened 2023-06-08
Cache IconWebComponent svg content.

raplemie:
## Changes

IconWebComponent now caches its SVG source so if the same url is requested many times, it will reuse the content across multiple icons. (making subsequent display instantaneous)

The fetching will also only occur once even if multiple icon require the same URL at the same time, reducing the network hits even in this case.

## Testing

Tested through changing the "SnapMode.tsx" and assigning all the same icon, and ensuring that all icons were displayed and the network was not hit more than once, and the display time was close to 0 ms on subsequent render of the same icon.
Through this testing I realized that caching the SVG instead of the fetch promise was taking less time on subsequent calls, so I added the `svgCache` and `getList` different caching for this reason (instead of just caching the `getList` and always reusing the promises.

### DIFF
--- a/ui/core-react/src/core-react/utils/IconWebComponent.ts
+++ b/ui/core-react/src/core-react/utils/IconWebComponent.ts
@@ -14,6 +14,101 @@ import { UiCore } from "../UiCore";
 import DOMPurify, * as DOMPurifyNS from "dompurify";
 
 /**
+ * Completed SVG cache
+ * (If we get it once, we use it from there afterwards)
+ */
+const svgCache = new Map<string, HTMLElement>();
+/**
+ * Active "getSource" list
+ * (if multiple icon require the same thing at the same time,
+ * only do the fetch once, and have all icon use the same result,
+ * clear result afterwards, good or bad.)
+ */
+const getList = new Map<string, Promise<HTMLElement>>();
+
+/**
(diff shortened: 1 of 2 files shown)


ui/core-react/src/core-react/utils/IconWebComponent.ts
```diff
@@ -13,6 +13,101 @@ import { Logger } from "@itwin/core-bentley";
 import { UiCore } from "../UiCore";
 import DOMPurify, * as DOMPurifyNS from "dompurify";
 
+/**
+ * Completed SVG cache
+ * (If we get it once, we use it from there afterwards)
+ */
+const svgCache = new Map<string, HTMLElement>();
```
GerardasB, line 20: Do we really need this map? It should be enough to track the promises which would eliminate some conditionals.

GerardasB (changes requested):

ui/core-react/src/core-react/utils/IconWebComponent.ts
raplemie: As I mention in description, it reduce a bit more the time it takes to show on subsequent display, but I'll remove in for code clarity, I doubt that people will notice `4ms` delay and I suppose keeping the function scope in memory is small too compared to the SVG content anyway.

GerardasB (approved):

PR #369, opened 2023-06-16
Add storybook

GerardasB:
## Changes

This PR adds the initial storybook to be deployed to GH pages.
To run the storybook locally:
```
cd docs/storybook
npm run dev
```
## Testing

Ran the dev server. Built and served the static site to verify the stories worked as expected.


.github/workflows/publish-storybook.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
       - name: Setup node
         uses: actions/setup-node@v2
         with:
-          node-version: "14"
+          node-version: 18.14.0
```
aruniverse, line 28:
```suggestion
          node-version: 18.16.0
```

AntanasBuk (commented):
- From storybook feature point of view the only thing I suggest adding is a "Code" tab for every hook and component. I think one of the biggest user use-cases for a storybook would be the "plug and play" functionality of our components where they copy paste the code from storybook into their applications (this is in favor of some components that rely on others to work).
- Code looks good.
- Configuration and publishing needs another pair of eyes.

AntanasBuk (approved):

PR #370, opened 2023-06-16
PropertyFilterBuilder: Focus property selector after adding rule or rule group

simasjar:
## Changes
Now property selector in the filter dialog will be focused after adding the rule or rule group.
 

## Testing
Added one test to check if the property selector is focused after adding a new rule.


ui/components-react/src/test/filter-builder/FilterBuilderRuleGroup.test.tsx
```diff
@@ -248,4 +249,22 @@ describe("PropertyFilterBuilderRuleGroupRenderer", () => {
...
+      <PropertyFilterBuilderRuleGroupRenderer {...defaultProps} />,
+      { actions }
+    );
+
+    const addRuleButton = getByTestId("rule-group-add-rule");
+    addRuleButton.click();
+
+    await waitFor(
+      () =>
+        expect(
```
saskliutas, line 265: I would suggest checking this with `expect(container.querySelector(".rule-property input") === container.ownerDocument.activeElement).to.be.true;`

saskliutas (changes requested):

ui/components-react/src/test/filter-builder/FilterBuilderRuleGroup.test.tsx
```diff
@@ -248,4 +249,22 @@ describe("PropertyFilterBuilderRuleGroupRenderer", () => {
...
   });
+
+  it("focuses rule property after adding new rule", async () => {
+    const actions = new PropertyFilterBuilderActions(sinon.spy());
+    const { getByTestId, container } = renderWithContext(
+      <PropertyFilterBuilderRuleGroupRenderer {...defaultProps} />,
+      { actions }
+    );
+
+    const addRuleButton = getByTestId("rule-group-add-rule");
```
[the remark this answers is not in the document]
simasjar: fixed

ui/components-react/src/test/filter-builder/FilterBuilderRuleGroup.test.tsx
simasjar: fixed

saskliutas (approved):

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
```diff
@@ -206,3 +209,28 @@ const PropertyFilterBuilderGroupOrRule = React.memo(
     );
   }
 );
+
+const useRulePropertyFocus = (
+  currentGroupItemsLength: number,
+  groupRef: React.RefObject<HTMLDivElement>
```
GerardasB, line 215: Usually, I'd expect a hook to return a `React.Ref`. Might be fine in this case, since it is not exported.

GerardasB (approved):

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
saskliutas: Now that you mentioned this maybe it makes sense to change `React.RefObject<HTMLDivElement>` into `HTMLDivElement | null`?

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
```diff
@@ -206,3 +212,27 @@ const PropertyFilterBuilderGroupOrRule = React.memo(
...
+    ) {
+      const ruleProperties = groupRef.querySelectorAll<HTMLInputElement>(
+        ".rule-property input"
+      );
+      ruleProperties[ruleProperties.length - 1].focus();
+      isNewRuleAdded.current = false;
+    }
+    previousGroupItemsLength.current = currentGroupItemsLength;
+  }, [currentGroupItemsLength, groupRef]);
+  return isNewRuleAdded;
```
saskliutas, line 237: Why did you change from returning callback to ref?

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
simasjar: applied @saskliutas suggestion

GerardasB: I think the current solution might lead to some subtle issues where Ref value is changed w/o a re-render (hence useEffect logic is not run until the next render).

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
```diff
@@ -206,3 +212,27 @@ const PropertyFilterBuilderGroupOrRule = React.memo(
     );
   }
 );
+
+const useRulePropertyFocus = (
+  currentGroupItemsLength: number,
+  groupElement: HTMLDivElement | null
+) => {
+  const previousGroupItemsLength = React.useRef<number>(0);
```
GerardasB, line 220:
```suggestion
const useRulePropertyFocus = (
  currentGroupItemsLength: number,
  groupElement: HTMLDivElement | null
): [() => void, React.Ref<HTMLDivElement>] => {
  const ref= React.useRef<HTMLDivElement>(null);
  return [() => {}, ref];
}
```

I was thinking about something like that?

GerardasB (approved):

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
simasjar: I assumed that we don't need `groupElement: HTMLDivElement | null` as a second argument, fixed

GerardasB:
@simasjar please add a note to `NextVersion.md`

simasjar:
> @simasjar please add a note to `NextVersion.md`

Added

Issue #271, opened 2023-04-14
Theming does not work dor Select component after migrating to AppUI 4.x


ImVeryLost:
**Describe the bug**
`Select` dropdown component from `"@itwin/itwinui-react" (v2.8.1) does not fully respect dark theme - the drop down list always uses light theme variables.  Issue for not present for standalone itwinui-react samples (were appUI is not present)
![image](https://user-images.githubusercontent.com/61872067/232001427-3b63d505-9201-4d4a-b40e-374e62370b5b.png)

**To Reproduce**

1. Launch the `test-apps/appui-test-app/standalone` from this repo
2. Set the theme to dark
3.  Go to the 'Exercise Widget Api' fronstage
4. Select the drop down component from the "floating info" widget
5. Drop down list itself uses light theme (see screenshot above)

**Expected behavior**
Dark theme should be used for the list.

**Desktop (please complete the applicable information):**

- iTwin.js AppUI Version - 4.0.0-dev.16
- iTwin.js UI version - 2.x
- iTwin.js Version - 3.7.1

**Additional context**
Not sure if the issue should be filed here or on the iTwinUI side. Since it only appeared after migrating to AppUI 4.x, filed here.
Let me know if I should reach out to iTwinUI team instead.

mayank99:
This issue happens when the dropdown (popover) is placed _outside_ the main ThemeProvider (root) that the select-button (trigger) is in. It might be because there is another ThemeProvider/root on the page which uses light theme.

What is happening (rough pseudo-html):
```html
<ThemeProvider theme='light'>
  <ThemeProvider theme='dark'>
    <select-button />
    ...
  </ThemeProvider>
  ...
  <select-dropdown />
  ...
</ThemeProvider>
```

What should happen:
```html
<ThemeProvider theme='light'>
  <ThemeProvider theme='dark'>
    <select-button />
    ...
    <select-dropdown />
  </ThemeProvider>
  ...
</ThemeProvider>
```

One workaround is to use `appendTo: 'parent'` inside Select's `popoverProps`, but that will likely cause clipping/z-index issues. The proper fix would involve making sure that the popover is always portaled into the main ThemeProvider/root.

@gretanausedaite should be able to provide more details here.

gretanausedaite:
I think this Select is in `FloatingWidgetSelect` component which is appended to Standalone app via `FloatingWidgetsUiItemsProvider()`.

[...]

PR #373, opened 2023-06-19
Add `start` scripts for watch mode build

raplemie:
## Changes

In each packages `start` script will now enter build and copy assets (scss, svg, json...) in watch mode.

## Testing

Ran each one, validated that saving something will make it appear in standalone test app. (Expect connected to behave the same way)


raplemie:
resolves #271

PR #376, opened 2023-06-20
Conditional icon item

### LINKED ISSUE #315: ConditionalIconItem a valid type for IconSpec, but not working?
The [IconSpec](https://github.com/iTwin/appui/blob/7284cd95619addb7fce98abf3d3ee060fa42c2ba/ui/core-react/src/core-react/icons/IconComponent.tsx#L21) type lists [ConditionalIconItem](https://github.com/iTwin/appui/blob/master/ui/core-react/src/core-react/icons/ConditionalIconItem.tsx) as a valid type, but it doesn't appear to be functional for something like a [ToolItemDef](https://github.com/iTwin/appui/blob/master/ui/appui-react/src/appui-react/shared/ToolItemDef.ts) which has an `icon` field that is typed as `IconSpec`

**To Reproduce**
1. Create a ToolItemDef instance for a ToolbarComposer that is set with an icon defined as a `ConditionalIconItem`
2. Notice the icon in the app shows as blank

**Expected behavior**
Expected this to behave exactly like `ConditionalStringValue` except I could use a React.Node for the actual icon, which is the primary use case for attempting to use `ConditionalIconItem` instead

- OS: Windows
- Browser: Chrome
- Version: 112.0.5615.138
- iTwin.js AppUI Version: 4.0.0
- iTwin.js Version: 3.7.0

**Additional context**
I was expecting to find handling for this in [ToolbarItemsManager](https://github.com/iTwin/appui/blob/7284cd95619addb7fce98abf3d3ee060fa42c2ba/ui/appui-react/src/appui-react/toolbar/ToolbarItemsManager.ts#L156) where it checks the `ConditionalStringValue` and `ConditionalBooleanValue`

NancyMcCallB:
## Changes
 Update implementation of IconComponent and IconHelper to suppport ConditionIconItem.

## Testing
Unit testing and setting up the getSplitSingleViewportCommandDef in the ContentLayoutTools (appui-test-uiproviders) to use ConditionalIconItem. 

This fixes #315


ui/core-react/src/core-react/icons/ConditionalIconItem.tsx
```diff
@@ -27,6 +27,16 @@ export class ConditionalIconItem {
     this._value = value;
   }
 
+  /** A work-around for isInstanceOf */
+  public static isConditionalIconItem(item: any): boolean {
```
raplemie, line 31: Do you know why `ConditionalIconItem` required a workaround but `ConditionalStringValue` did not ?

I wonder if this could be related to the same issue I just had with `appui-test-providers` that wasnt build using esm and was breaking some use (I see that our example is in this package) Do we have an example of `ConditionalStringValue` in the providers ?

raplemie (approved):
LGTM, just wondering if the workaround is still needed with the new build of ESM from the appui-test-providers package, because if we start doing these workaround to support that, we might need to dig further into what else is breaking in this particular case (and some will never properly work with 2 different build system at the same time.)

ui/core-react/src/core-react/icons/ConditionalIconItem.tsx
GerardasB, line 31: Do we intend to use this internally only, maybe we could mark this as @internal?
Would this make more sense as a type-guard?
Could we avoid `any` type?
Adding a new public API with "workaround" docs is not ideal.

GerardasB (commented):
A missing section/notes in the `NextVersion.md`

ui/core-react/src/core-react/utils/IconHelper.tsx
```diff
@@ -18,17 +19,31 @@ export class IconHelper {
...
   public static getIconReactNode(
-    icon: string | ConditionalStringValue | React.ReactNode,
+    icon:
+      | string
+      | ConditionalStringValue
+      | React.ReactNode
+      | ConditionalIconItem,
     internalData?: Map<string, any>
-  ): React.ReactNode {
+  ): React.ReactNode | ConditionalIconItem {
```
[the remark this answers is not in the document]
NancyMcCallB: Good question. I thought we could add types but not change existing types. Is that incorrect?

ui/core-react/src/core-react/icons/ConditionalIconItem.tsx
NancyMcCallB: Good idea -- i should have marked it internal.

ui/core-react/src/core-react/utils/IconHelper.tsx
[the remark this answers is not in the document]
GerardasB: We could add optional properties to an interface. To add new values to a union we'd have to be sure that consumer code tests exhaustively for all possible values. In this case, the returned value is probably just rendered, which might cause errors.
There is more info here: https://www.itwinjs.org/learning/api-support-policies/

Btw, this seems to just be an excessive definition, can't we just remove `ConditionalIconItem` from the return type since we never return a `ConditionalIconItem`, it's always a `ReactNode` that's returned.

Issue #364, opened 2023-06-14
Explore current popouts limitations


raplemie:
Our current implementation does not work when the application is within an `iframe`, we need to validate what are the general limitations of popouts "normal" use and also within an iFrame as this is used by internal Bentley applications

```[tasklist]
### Tasks
- [ ] https://github.com/iTwin/appui/issues/379
- [ ] https://github.com/iTwin/appui/issues/383
- [ ] https://github.com/iTwin/appui/issues/384
- [ ] https://github.com/iTwin/appui/issues/202
- [ ] https://github.com/iTwin/appui/issues/273
- [ ] https://github.com/iTwin/appui/issues/380
- [ ] https://github.com/iTwin/appui/issues/381
- [ ] https://github.com/iTwin/appui/issues/402
- [ ] https://github.com/iTwin/appui/issues/498
- [ ] https://github.com/iTwin/appui/issues/494
- [ ] https://github.com/iTwin/appui/issues/536
- [ ] https://github.com/iTwin/appui/issues/524
- [ ] https://github.com/iTwin/appui/issues/563
- [ ] https://github.com/iTwin/appui/issues/564
- [ ] https://github.com/iTwin/appui/issues/612
- [ ] https://github.com/iTwin/appui/issues/627
- [ ] https://github.com/iTwin/appui/issues/632
- [ ] https://github.com/iTwin/appui/issues/649
- [ ] https://github.com/iTwin/appui/issues/650
- [x] Multiple react element trees can not share root context providers. Investigate using the portaling instead. https://github.com/iTwin/appui/pull/820
```

raplemie:
Electron exploration still needed

PR #382, opened 2023-06-22
Fix popout location

### LINKED ISSUE #379: Fix `adjustWindowLocation` issue when used within an `<iframe />`
`adjustWindowLocation` method currently attempts to use `window.top` properties to ensure centering on the parent window, which is prohibited when inside an `iframe`.

Turns out that this is not needed and `window.screenX` for example, will return the "browser" window, not the frame information, so we get the correct information regardless of using `window.top.screenX` or `window.screenX`.

Also, current implementation have a code path for "not centered", however, this is a private method and the "not centered" parameter is never requested, so this code (which contains bugs anyway) is not needed and should be cleaned up.

### LINKED ISSUE #380: Stop using `onbeforeunload`, prefer `pagehide` event
When iframed, `onbeforeunload` will be allowed by firefox only if `allow-modals` is provided, on top of our needed `allow-popups`.
There are a couple of issue with `onbeforeunload` as well and according to https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event, the `pagehide` event is better to use. (in our use, we dont try to block the user from closing so this is a better choice anyway)

### LINKED ISSUE #381: When reopening a popup, it sometimes do not appear where left
When the browser is not full screen, a popup window will often move to the top of left of the screen after its initial loading.

This is caused by an attempt at containing the popouts within the viewer area, but lacks the screen location.
We should remove this `contain` and allow the window to repop where it was last closed.

raplemie:
## Changes

Modified `adjustWindowLocation` private method of `InternalChildWindowManager` to not use `window.top` properties.
Removed dead code from the private function that was never called.
resolves #379 

Removed a `contain` in appui-layout-react that was unexpectedly trying to contain popped out wigets, now the widget will reappear where they were last closed by the user.
resolves #381 

Using `pagehide` event handler instead of `onbeforeunload` for firefox and overall posible better results.
resolves #380 

## Testing

Visual/Manual testing through `test-apps/standalone` (web)

### DIFF
--- a/common/changes/@itwin/appui-layout-react/raplemie-popoutsLocation_2023-06-22-17-07.json
+++ b/common/changes/@itwin/appui-layout-react/raplemie-popoutsLocation_2023-06-22-17-07.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-layout-react",
+      "comment": "Do not limit popped out widget location on the screen",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/appui-layout-react"
+}
--- a/common/changes/@itwin/appui-react/raplemie-popoutsLocation_2023-06-22-17-07.json
+++ b/common/changes/@itwin/appui-react/raplemie-popoutsLocation_2023-06-22-17-07.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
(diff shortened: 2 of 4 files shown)


raplemie:
@jason-crow @simnorm This fixes a couple of initial issues with iFramed AppUI attempting to use popouts, if the iFrame is sandboxed, it will still require the `allow-popups` rule but that should allow you guys to test.

Be aware that there are still issues so this should probably not used in prod yet, see #364 for a list, but mainly #384 would probably be a show stopper.

NancyMcCallB (approved):

PR #390, opened 2023-06-27
Add Component Examples to test app

kconner1:
## Changes
Added component examples to test app

## Testing
Run test app and navigate to "Component Examples" frontstage


AntanasBuk (changes requested):
- Description should be provided.
- We now also use "prettier" extension in VSCode, you should add it and run it on the files you've created or modified, else the tests will fail.

test-apps/appui-test-app/appui-test-providers/src/ui/providers/ComponentExamplesProvider.tsx
```diff
@@ -0,0 +1,756 @@
...
+
+class TestContentControl extends ContentControl {
+  constructor(info: ConfigurableCreateInfo, options: any) {
+    super(info, options);
+
+    this.reactNode = <div>Test</div>;
+  }
+}
+
+class TestUiDataProvider extends DialogLayoutDataProvider {
```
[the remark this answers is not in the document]
kconner1: To do this, my understanding is that I would have to make a file with the classes in the appui package, export that, and the import in the test app. I am not sure that is the best idea considering we would only want this for use internally. I could import the file directly without exporting, but that would break the import within package lint rule which I could override. However, I am not sure that is the best idea. The way that I currently have this set up is parallel to how classes from test files were reused for component examples previously in the twins-core repo.

tools/codemod/src/v4.0.0/__testfixtures__/full.input.tsx
```diff
@@ -1,7 +1,7 @@
 /*---------------------------------------------------------------------------------------------
-* Copyright (c) Bentley Systems, Incorporated. All rights reserved.
-* See LICENSE.md in the project root for license terms and full copyright notice.
-*--------------------------------------------------------------------------------------------*/
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
```
GerardasB, line 2: In addition to installing the extension you can fix current formatting inconsistencies for all files by running:
`rush prettier:fix`
This should get rid of changes like these :)

common/changes/@itwin/appui-codemod/kayla-componentExamples_2023-06-29-20-20.json
```diff
@@ -0,0 +1,10 @@
+{
```
GerardasB, line 1: `rush change` only generated this due to an (unnecessary) formatting change in codemod package. I think you should get rid of the change files, run `rush prettier:fix` then `rush change` again.

ui/appui-layout-react/CHANGELOG.json
```diff
@@ -1,21 +1,6 @@
 {
   "name": "@itwin/appui-layout-react",
   "entries": [
-    {
```
GerardasB, line 4: We don't want to remove these changes from changelog files :)

docs/changehistory/4.2.0.md
```diff
@@ -1,57 +0,0 @@
----
```
GerardasB, line 1: We want to keep this..

GerardasB (changes requested):

common/changes/@itwin/appui-layout-react/changelogs-4.2.0_2023-06-27-15-06.json
GerardasB, line 1: We shouldn't remove these either.

test-apps/appui-test-app/appui-test-providers/src/ui/providers/ComponentExamplesProvider.tsx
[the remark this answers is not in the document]
GerardasB: I agree, let's avoid exporting test-specific things from our packages. It might indicate poor unit test quality if providers used for visual testing match our unit test providers.

GerardasB:
I couldn't get "Cursor Popup" example to run. "Configurable Ui Content" looked fun since I could open the same example recursively multiple times (Examples within examples etc..).
Looks like there are some styling issues as well for some of these components (i.e. status bar stuff) when rendered in isolation (which we should address in a separate PR). 
I think we will need to consider moving these examples to the storybook sometime in the future, but this will give us a big head start.
Other than that the `Component Examples` page looks great, good job!

tools/codemod/src/v4.0.0/__testfixtures__/full.input.tsx
GerardasB: You are correct, files under `__testfixtures__` are supposed to be ignored by prettier (see `.prettierignore`), not sure why this got formatted (for now you can revert this by hand).

kconner1: Git is not allowing me to commit the changes of removing the spaces. I keep getting this error: 
```
⚠ lint-staged prevented an empty git commit.
  Use the --allow-empty option to continue, or check your task configuration
```

GerardasB: You are correct, `lint-staged` runs a git hook to update the copyright headers of changed files.
For now I've used `-n` https://git-scm.com/docs/git-commit#Documentation/git-commit.txt--n to get rid of this change, but we'll want to update the headers of `__testfixtures__` files in a separate PR.

GerardasB (approved):

AntanasBuk (approved):

PR #391, opened 2023-07-03
ControlledTree: cleanup copiled SparseTree code

saskliutas:
## Changes

`SparseArray` class has two symbol properties defined: `immerable` and `Symbol.iterator`. For some reason compiled javascript looks like this:

```typescript
var _b;

export class SparseArray {
  constructor() {
      this[_b] = true;
      // ...
  }
 
  [(_b = immerable, Symbol.iterator)]() {
    // ...
  }
}
```

This causes some problems for terser when minimizing code and removing dead code to reduce bundle size. Minimized code looks like this:

```javascript
(_b = lC), Symbol.iterator;
```

Which under `"use strict"` mode throws runtime error because `_b` is not defined: https://github.com/iTwin/viewer/issues/283

Setting `immerable` property value in constructor seems to result in cleaner compiled javascript code which does not cause problems for terser:

Compiled javascript code:
```typescript
export class SparseArray {
  constructor() {
      // ...
      this[immerable] = true;
  }
  [Symbol.iterator]() {
    // ...
  }
}
```

Terser output:
```javascript
Symbol.iterator;
```

This probably should be fixed at `terser` or `typescript` level (maybe some changes to `tsconfig` would solve this). But at least for now this provides a workaround to avoid runtime error when building apps with `@bentley/react-scripts`

## Testing

Building locally with different configurations.

### DIFF
--- a/common/api/components-react.api.md
+++ b/common/api/components-react.api.md
@@ -2544,6 +2544,7 @@ export class SparseArray<T> implements Iterable<T | undefined> {
     // (undocumented)
     [immerable]: boolean;
     [Symbol.iterator](): IterableIterator<T | undefined>;
+    constructor();
     get(index: number): T | undefined;
     getIndex(lookupValue: T): number | undefined;
     getLength(): number;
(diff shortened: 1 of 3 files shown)


aruniverse (approved):

simasjar (approved):

GerardasB (approved):

GerardasB:
@Mergifyio backport release/4.2.x

PR #395, opened 2023-07-06
Empty toolsettings message

NancyMcCallB:
Fix for #345 

## Changes

Update the message displayed for the empty tool settings state so that the tool name is included.
Compose the message with before and after strings and the tool name to allow for different grammatical structure when translating to other languages.
Use the string "Active Tool" in the unlikely event that the tool has no registered name.

## Testing

Tested in appui-test-app


ui/appui-react/src/appui-react/widget-panels/ToolSettings.tsx
```diff
@@ -30,12 +30,20 @@ export interface ToolSettingsEntry {
...
+  const labelStringStart = IModelApp.localization.getLocalizedString(
+    "UiFramework:tools.noToolSettingsStart"
+  );
+  const labelStringEnd = IModelApp.localization.getLocalizedString(
+    "UiFramework:tools.noToolSettingsEnd"
   );
   const labelNode = (
     <div className="uif-toolsetting-label-docked-horizontal-empty">
-      {labelString}
+      {labelStringStart}{toolName ? toolName : toolPlaceholderName}{labelStringEnd}
```
GerardasB, line 46: Could we work on **styling** and render an **icon** as suggested in the concept image?

GerardasB (changes requested):

ui/appui-react/src/appui-react/widget-panels/ToolSettings.tsx
GerardasB: Need a note in `NextVersion.md` as well.

NancyMcCallB: I believe rendering the tool icon as an inactive picture in the tool settings is confusing.

PR #396, opened 2023-07-07
Update sinon to 15.2.0 and @types/sinon to 10.0.15

anmolshres98:
Update sinon to 15.2.0 and @types/sinon to 10.0.15


common/config/rush/pnpm-lock.yaml
```diff
@@ -9,7 +9,7 @@ importers:
     specifiers:
       '@playwright/test': ^1.24.1
     devDependencies:
-      '@playwright/test': 1.31.2
```
GerardasB, line 12: Seems like dependencies other than `sinon` were updated as well, did you by any chance run `rush update --full`? Was that intentional?

GerardasB (approved):

common/config/rush/pnpm-lock.yaml
aruniverse: i doubt doing the full update was intentional; but it might be a good thing; now that you guys have broken off from the core repo, how often are you guys doing full updates? core used to do it whenever there was a new addon

anmolshres98: It wasn’t intentional but yes I did run full rush update. Let me know if I should revert. Thanks

anmolshres98:
> why are there no ts changes?

The tests were passing so didn’t see where to make ts changes

aruniverse:
> > why are there no ts changes?
> 
> The tests were passing so didn’t see where to make ts changes

take a look at this build failure; im expecting some ts changes to fix this:
https://github.com/iTwin/nightman/actions/runs/5502644897/jobs/10027049674

anmolshres98:
> > > why are there no ts changes?
> > 
> > 
> > The tests were passing so didn’t see where to make ts changes
> 
> take a look at this build failure; im expecting some ts changes to fix this: https://github.com/iTwin/nightman/actions/runs/5502644897/jobs/10027049674

I think that error might have been from core-appui sinon type version mismatch. Here's a nightman build with this branch checked out for appui and master/main for all other repos: https://github.com/iTwin/nightman/actions/runs/5511425916/jobs/10046960042

`pnpm build` passing on all for this run. Ignore the lint error for now

anmolshres98:
Need help fixing the `extract-api` run failure. When I run locally, the changes made are all similar to this: 
image

I don't think this is right though 🤔

aruniverse:
> Need help fixing the `extract-api` run failure. When I run locally, the changes made are all similar to this: image
> 
> I don't think this is right though 🤔

revert the `lock file` and then just run `rush update` the only deps updated should be sinon and types/sinon

PR #397, opened 2023-07-09
Remove the usage of internal TelemetryManager from Core

calebmshafer:
## Changes

Removing the usage of the `internal` Core telemetry manager APIs. This API has been marked as `internal` and `alpha`  since the 3.0 but the usage of it has not been cleaned up everywhere since then. At the time, the purpose of moving it internal was that the telemetry, which was being used more as feature tracking, should be the responsibility of the application and not the core library.

The suggested way for the libraries to enable applications to handle the feature tracking workflow moving forward is raising events, or other observable state. This will allow the app to decide how to handle it and in many cases an event is already being raised.

The API is sparsely used throughout this repo and a few others to indicate when a significant change happens. In the cases within the AppUI repository and is directly coupled with either an event being raised or a state change.

## Testing

N/A


GerardasB (approved):
LGTM, would be great to add a section to `NextVersion.md` based on your PR description :)

PR #401, opened 2023-07-11
Upgrade to eslint 8 and new flat config system

### LINKED ISSUE #360: Move to eslint 8
Followup to #231 

- `@itwin/eslint-pluging`: 4.dev should be safe to use.
- `eslint` ^8

### LINKED ISSUE #413: Fix "@itwin/no-internal-barrel-imports" linting errors
**Describe the bug**

Because of [this PR I created](https://github.com/iTwin/appui/pull/401), there has been introduced a new linting error, specifically `@itwin/no-internal-barrel-imports : Do not consume barrel imports within the same package`. I have set the rule to "warn" for now in this base config file [here](https://github.com/iTwin/appui/pull/401#discussion_r1260243922) so that the PR can be merged. This "warn` value should be removed and the linting errors should be fixed instead.

**To Reproduce**

1. Go to base of repo
2. Run `rush lint`
3. See all the linting problems reported as warning
4. Remove the line with ` "@itwin/no-internal-barrel-imports": "warn",` in `ui/eslint.config.base.js`
5. Fix all the linting errors

**Expected behavior**
No linting problems reported when running `rush lint`

**Screenshots**
Example run of `rush lint`:
image

anmolshres98:
Change motivated from Nightman build failures owing to eslint-plugin version mismatch:https://github.com/iTwin/nightman/actions/runs/5514259302/jobs/10053301272

Specifically, newer eslint-plugin uses a [new config system](https://eslint.org/blog/2022/08/new-config-system-part-2/) and our packages need to be updated to use this new system

resolves #360 
resolves #413


ui/eslint.config.base.js
```diff
@@ -0,0 +1,35 @@
...
+        "error",
+        {
+          args: "after-used",
+          argsIgnorePattern: "^_",
+          varsIgnorePattern: "^_",
+          ignoreRestSiblings: true,
+        },
+      ],
+      "no-duplicate-imports": "off",
+      "@itwin/no-internal-barrel-imports": "warn",
```
anmolshres98, line 32: This is a new addition I made, will come back to this after making this upgrade to presentation repo as well. Here to prevent lint failure for now.

anmolshres98:
@Jake-Screen could you review this for me please, thanks!

common/changes/@itwin/appui-layout-react/as-eslint-upgrade_2023-07-11-20-20.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-layout-react",
+      "comment": "Upgrade @itwin/eslint-plugin to 4.0.0-dev.43 and eslint to 8.44.0",
```
aruniverse, line 5: for all changelogs
these are what the users see, they shouldnt have to know which exact version or see mention of a pre-release 

```suggestion
      "comment": "Upgrade to eslint@8 and @itwin/eslint-plugin accordingly",
```

.vscode/settings.json
```diff
@@ -96,5 +96,6 @@
   "cSpell.words": ["Discoverability"],
   "typescript.tsdk": "common/temp/node_modules/typescript/lib",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "files.insertFinalNewline": true
+  "files.insertFinalNewline": true,
+  "eslint.experimental.useFlatConfig": true
```
aruniverse, line 100: want to make sure eslint warnings shows up correctly in vscode

ui/imodel-components-react/package.json
```diff
@@ -106,8 +106,7 @@
     "ts-node": "^10.8.2",
     "typemoq": "^2.1.0",
     "typescript": "~5.0.2",
-    "eslint-config-prettier": "~8.8.0",
-    "@rushstack/eslint-patch": "^1.3.0"
+    "eslint-config-prettier": "~8.8.0"
```
aruniverse, line 109: can you alphabetize deps

test-apps/appui-test-app/standalone/eslint.config.js
```diff
@@ -0,0 +1,44 @@
...
+    ecmaVersion: "latest",
+    ecmaFeatures: {
+      jsx: true,
+      modules: true,
+    },
+  },
+};
+
+module.exports = [
+  {
```
aruniverse, line 18: youll want to make sure to add an entry for the uiconfig for src/frontend/**

aruniverse (commented):
jake is out for the rest of the week
i think things look fine
take the pr out of draft to see if lint works in the ci

ui/imodel-components-react/package.json
anmolshres98: sure, will do. I used `rush add` to add this deps, i'm surprised it didn't do it automatically. Am I doing it wrong? 🤔

aruniverse: i typically edit the pkg.json manually so never noticed

.vscode/settings.json
anmolshres98: Honestly can't tell since eslint plugin even on a normal basis is broken for my vscode. Can verify linting is working from the cli. I added this because of the README instruction here: https://github.com/iTwin/eslint-plugin. Lmk if you want me to remove, will do so. Thanks.

aruniverse: you shoudl see squiggly lines when theres a linter error
make sure you have the vscode eslint extension installed
change some code so you have a linter error

image

if you go to the output tab, change the task to eslint and you should hopefully see some errors in there

anmolshres98: Gotcha, it looks like it cannot find the config file it is looking for. I imagine if this is because of the monorepo rush setup we have: 
image

aruniverse: Plz get that working then. Might need to mess with vscode eslint settings

test-apps/appui-test-app/standalone/eslint.config.js
aruniverse: this is applicable to all test apps, i just didnt want to repeat myself and make multiple comments

anmolshres98: I see what you mean but the older config does not have the uiconfig included. Are you suggesting that I should make this addition despite the older version not having these? 
image

anmolshres98: The ones importing in `eslint.config.base.js` should have that config btw

ui/eslint.config.base.js
anmolshres98: @aruniverse Should I be working on fixing the errors created by this. This seems to be newer (originating from newer eslint-plugin) and there are a ton of these ([build](https://dev.azure.com/bentleycs/iModelTechnologies/_build/results?buildId=2268891&view=logs&jobId=9f3a3068-7880-5994-433d-b92c1f9d379b&j=9f3a3068-7880-5994-433d-b92c1f9d379b&t=83bcf1fc-5ebb-5f5c-0841-3e5c18e8878e)). I don't know if I should leave it to warn/off and have someone else on appui working on fixing it or if I should do it myself.

The lint errors seem to be coming from test files only (I think) and the meaning of error is this: 
image

.vscode/settings.json
anmolshres98: working now

ui/eslint.config.base.js
```diff
@@ -0,0 +1,36 @@
+// eslint plugin on vscode cannot resolve with just `require("@itwin/eslint-plugin")`
+const iTwinPlugin = require("../common/temp/node_modules/@itwin/eslint-plugin");
```
anmolshres98, line 2: importing like this so that vscode eslint works properly. @aruniverse lmk what you think. Thanks!

ui/eslint.config.base.js
aruniverse: make a separate issue and have the appui team work on it, just suppress it for now

ui/eslint.config.base.js
aruniverse: seems hacky, theres prob a better way. plz investigate

anmolshres98: Here are some approaches I tried using:
1. Using the `@rushstack/eslint-patch/modern-module-resolution` library, which was also being used before but it might be unsupported now in newer ESLint version so getting error message both in VS Code ESLint extension and cli run: 
image Hence, this cannot be a solution
1. Using `eslint.options` in `.settings.json` like so: 
image
This was being used in the core repo. However, this option seems to have been removed entirely and does not work at all. Output from ESLint VS Code extension: 
image
So this is not an option either.
1. We can have the base config use only custom rules and each config file for each package import itwin/eslint-plugin every time. The change would look like so: 
image
Note that each `eslint.config.js` importing `eslint.config.base.js` would have to make those additions, essentially having multiple repetitions (violates DRY). This solution works for both cli runs and VS Code ESLint extension but it has the tradeoff of repetition like I mentioned above. 

At this point, any workaround I think will be somewhat hacky or need to have repetition like I mentioned. There might be a better solution in the future or I could have just missed something completely obvious, which is always a possibility.

Let me know if you wanna go with Option 3 or we should go with a different approach @aruniverse . Thanks!

ui/eslint.config.base.js
anmolshres98: Issue created: #413 , Should I assign to someone specific?

aruniverse:
@yashincontrol @Jake-Screen plz review

ui/eslint.config.base.js
aruniverse: try moving this based config to the root of the repo
and then at the repo root package.json add the itwin eslint plugin pkg
and then try updating the require to as it should be, require(itwin/eslint-plugin), and see if that works

aruniverse: > VS Code Extension still showing the error message: Cannot find module '@itwin/eslint-plugin'

did you restart the eslint server? use the command pallette to do so
theres no vscode eslint extension settings to point to the correct installation?

anmolshres98: Yup, I have been restarting the ESLint server using command pallette with every smallest change I make and verifying by adding an unused var on a tsx file. (Error shows up for unused var if ESLint working on VSCode). 

Yes, I have tried different variations with `eslint.nodePath` and `eslint. workingDirectories` settings and also the `resolvePluginsRelativeTo` as mentioned above with no luck so far. Most errors showing up when I tweak these settings are errors mentioning how the config file could not be found

.vscode/settings.json
```diff
@@ -96,5 +96,11 @@
   "cSpell.words": ["Discoverability"],
   "typescript.tsdk": "common/temp/node_modules/typescript/lib",
   "editor.defaultFormatter": "esbenp.prettier-vscode",
-  "files.insertFinalNewline": true
+  "files.insertFinalNewline": true,
+  "eslint.experimental.useFlatConfig": true,
+  "eslint.workingDirectories": [
```
Jake-Screen, line 101: What is the workingDirectories setting here for?

test-apps/appui-test-app/connected/eslint.config.js
```diff
@@ -0,0 +1,44 @@
...
+    ecmaVersion: "latest",
+    ecmaFeatures: {
+      jsx: true,
+      modules: true,
+    },
+  },
+};
+
+module.exports = [
+  {
```
Jake-Screen, line 18: add the entry for the uiConfig like in arun's comment

.vscode/settings.json
anmolshres98: VSCode ESLint extension does not function properly without it. If removed, the ESLint server spits out this error:
image

test-apps/appui-test-app/connected/eslint.config.js
anmolshres98: uiConfig doesn't seem to be part of the older eslint config, so this addition would be entirely new correct? And I should add uiConfig to all configs inside `test-apps/appui-test-app`?

Jake-Screen: Yes, to all of the test apps. For connected and standalone it would be the src/frontend folder. for appui-test-providers it looks like it would be the ui folder

ui/eslint.config.base.js
Jake-Screen: Another potential option would be to create a private package in the ui folder that exports the base config. Making it private would have it not be published and allow you to set eslint-plugin as a dependency, helping to fix the issue of having to point to ../common/temp/node_modules

anmolshres98: That might very well solve the issue at hand but creating an entirely new private package with a single dep and no code just to solve a vs code es lint issue seems pretty hacky to me and imo isn't the most simple and right solution here

ui/components-react/src/test/TestUtils.ts
```diff
@@ -33,7 +33,8 @@ import { ITwinLocalization } from "@itwin/core-i18n";
 import { prettyDOM } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import type { AsyncValueProcessingResult } from "../components-react";
-import { DataControllerBase, UiComponents } from "../components-react";
+import { UiComponents } from "../components-react/UiComponents";
+import { DataControllerBase } from "../components-react/editors/PropertyEditorManager";
```
yashincontrol, line 37: Was this change done to resolve the no-internal-barrel-imports error or just code cleanup?

ui/eslint.config.base.js
Jake-Screen: The biggest issue seems to be placement in reference to node_modules.The change to a flat config means that directly requiring configs and other items is now the way to do it. We would have to place it in a location that has access to a node_modules folder otherwise we would have to do the referencing to the common folder. 

From the above options i like 3 the most since you could place it anywhere and avoid the reference issues. it does require some repetition like you say but it seems like the best option due to the removal of the other settings that we used before

test-apps/appui-test-app/connected/eslint.config.js
anmolshres98: gotcha, all 3 packages should have uiConfig in their eslint config now

ui/eslint.config.base.js
anmolshres98: I somewhat agree as well. Since each of these packages already have `@itwin/eslint-plugin` in their list of dependencies already.

I pushed this implementation on my recent commit, we can always revert this commit if we choose to not go with it

ui/components-react/src/test/TestUtils.ts
anmolshres98: This was for a lint error fix but a rule I updated should take care of this so I have reverted this change, good catch. Thanks

yashincontrol: Sounds good, thanks. Some of the auto fixes were breaking the build pipelines on core, so I used the --fix option only for the comma dangle error.

.vscode/settings.json
Jake-Screen: Is this still needed after the change to the location of the base config?

ui/components-react/src/test/TestUtils.ts
anmolshres98: Yup, I have also found that fixing no barrel imports breaks code coverage

Jake-Screen (approved):

yashincontrol (approved):

ui/imodel-components-react/package.json
```diff
@@ -68,7 +68,7 @@
     "@itwin/core-orbitgt": "^3.7.0",
     "@itwin/core-quantity": "^3.7.0",
     "@itwin/core-react": "workspace:*",
-    "@itwin/eslint-plugin": "^3.7.0",
+    "@itwin/eslint-plugin": "^4.0.0-dev.43",
```
aruniverse, line 71: can you pull in dev45?

test-apps/appui-test-app/connected/src/frontend/appui/oidc/SignIn.tsx
```diff
@@ -134,6 +134,7 @@ export class SignIn extends React.PureComponent<SignInProps, SignInState> {
           {this.props.onRegister !== undefined && (
             <span className="components-signin-register">
               {this.state.profilePrompt}
+              {/* eslint-disable-next-line jsx-a11y/anchor-is-valid */}
```
aruniverse, line 137: why not just fix this instead of disabling it?

test-apps/appui-test-app/connected/src/frontend/appui/imodelopen/ITwinDropdown.tsx
```diff
@@ -96,6 +96,7 @@ export class ITwinDropdown extends React.Component<
         <ul style={ulStyle}>
           {iTwins &&
             iTwins.map((iTwin: ITwin, i: number) => (
+              // eslint-disable-next-line jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-element-interactions
```
aruniverse, line 99: why not fix these as well?

aruniverse (approved):

common/changes/@itwin/appui-layout-react/as-eslint-upgrade_2023-07-11-20-20.json
raplemie: @aruniverse I fell there should be no comment at all, this is internal dev changes only, not for consumers, any reason to add a changelog comment at all for this ?

raplemie (changes requested):

common/changes/@itwin/appui-layout-react/as-eslint-upgrade_2023-07-11-20-20.json
aruniverse: good point, i'm ok not having a comment at all for the changelogs

anmolshres98: updated

ui/eslint.config.base.js
```diff
@@ -0,0 +1,30 @@
+module.exports = [
```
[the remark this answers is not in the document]
anmolshres98: done, thanks

raplemie (approved):
LGTM!

Issue #406, opened 2023-07-12
PropertyGrid and Tree might become blank when switching widgets


saskliutas:
VirtuallizedPropertyGrid and ControlledTree might become blank when switching widgets. This happens when PropertyGrid/Tree is scrolled to the bottom, other widget is opened and then widget containing PropertyGrid/Tree is opened again.

This issue is encountered because when switching widgets components inside aren't unmouted. As both PropertyGrid and Tree uses virtualized list we end up in a situation where list is scrolled to the bottom but scroll returns to the top after opening widget again without scroll event. Need to update virtualized list scroll position in this case.

raplemie:
Fix for this was included in 4.3.0.

PR #410, opened 2023-07-13
Persist scroll position when using PropertyGrid or ControlledTree in Widget

saskliutas:
Added utility hooks that used with `useTransientState` allows to persist PropertyGrid and ControlledTree scroll position when switching between widgets.

Closes https://github.com/iTwin/appui/issues/406


grigasp (approved):

common/api/components-react.api.md
```diff
@@ -3411,6 +3411,13 @@ export class UrlPropertyValueRenderer implements IPropertyValueRenderer {
 // @public
 export function useAsyncValue<T>(value: T | PromiseLike<T>): T | undefined;
 
+// @public
+export function useControlledTreeLayoutStorage<T extends Element>(): {
+    ref: React_2.RefObject<T>;
```
GerardasB, line 3416: Could we change the `ref` type of these hooks to `Ref<T>`?

GerardasB (changes requested):

common/api/components-react.api.md
saskliutas: Changed to `Ref<T>`

GerardasB (approved):

PR #420, opened 2023-07-20
Fix selection count on BlankConnection

GerardasB:
## Changes

Add `BlankConnection` handling to `SyncUiEventDispatcher`. Fix issues with store initialization and reset when iModelConnection is set.

## Testing

Tested via standalone test-app. Added a decorator with a couple of pickable graphics to validate that those are added to a selection set.


raplemie (approved):
LGTM

PR #423, opened 2023-07-21
Add `hasDocs` on release version bumps commits

raplemie:
## Changes

Change the pipeline to add `hasDocs` tag on the docs if these should be published (only when releasing a new package)

## Testing

None

### DIFF
--- a/common/config/azure-pipelines/jobs/docs-ci.yaml
+++ b/common/config/azure-pipelines/jobs/docs-ci.yaml
@@ -64,26 +64,6 @@ stages:
               PathtoPublish: "$(Build.StagingDirectory)/docs/"
               ArtifactName: "AppUI Docs"
 
-          - powershell: |
-              $commitMsg = @"
-              $(Build.SourceVersionMessage)
-              "@
-              if ($commitMsg -match '^(\d+.\d+.\d+)$') {
-                Write-Host `'$commitMsg`'` is a major/minor/patch version bump
-                Write-Host '##vso[task.setvariable variable=ShouldPublish;]True'
-              } else {
-                Write-Host `'$commitMsg`'` is not a major/minor/patch version bump
-                Write-Host '##vso[task.setvariable variable=ShouldPublish;]False'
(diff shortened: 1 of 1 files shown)


common/config/azure-pipelines/jobs/docs-ci.yaml
```diff
@@ -64,6 +64,26 @@ stages:
...
+                Write-Host '##vso[task.setvariable variable=ShouldPublish;]False'
+              }
+            displayName: Publish if major/minor/patch version bump
+            condition: and(succeeded(), in(variables['Build.Reason'], 'IndividualCI', 'Schedule', 'Manual'), eq(variables['Agent.OS'], 'Windows_NT'))
+
+          - task: tagBuildOrRelease@0
+            inputs:
+              type: "Build"
+              tags: "hasDocs"
+            condition: and(succeeded(), in(variables['Build.Reason'], 'IndividualCI', 'Schedule', 'Manual'), eq(variables['Agent.OS'], 'Windows_NT'), eq(variables['ShouldPublish'], 'True'))
```
raplemie, line 85: @ben-polinsky I added this with the idea that we only want to publish the new docs when there is an actual release (so, no RC or Nightly will get tagged), however, the build will be tagged _before_ the docs gets validated below, so I'm not sure it make sense... Any idea on this ? I was starting to think to add this bits after the templated job in "Validate_Docs" below, but I dont know if that would tag the right thing at all. (And I'm not even sure how to test this without publishing a bunch of npm packages by running the pipelines)

ben-polinsky: If you set it up after on a different stage, it will tag correctly. That's sort of how presentation and transformer have their builds set up.

raplemie: Thanks, changed for a different stage.

GerardasB (approved):

ben-polinsky (approved):

PR #425, opened 2023-07-25
Fix desynced element hiding and clearing

AntanasBuk:
## Changes

Buttons `clearHideIsolateEmphasizeElements` and `hideElements` now display correctly* when individual elements' visibility in the view is toggled.

*`clearHideIsolateEmphasizeElements` button should not be displayed if no element in the current view is hidden. `hideElements` button should only be displayed if elements in the active selection (if one exists) are not hidden.

## Testing

Set up a cospace with viewer-components-react & appui and tested button updating their states while toggling individual element visibility from Models Tree.

### DIFF
--- a/ui/appui-react/src/appui-react/syncui/SyncUiEventDispatcher.ts
+++ b/ui/appui-react/src/appui-react/syncui/SyncUiEventDispatcher.ts
@@ -221,65 +221,27 @@ export class SyncUiEventDispatcher {
         if (undefined === args.previous) {
           void IModelApp.toolAdmin.startDefaultTool();
         } else {
-          // istanbul ignore next
-          if (
-            args.previous.onViewChanged &&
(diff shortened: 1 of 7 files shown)


NancyMcCallB (approved):

ui/appui-react/src/appui-react/syncui/SyncUiEventDispatcher.ts
```diff
@@ -253,6 +271,15 @@ export class SyncUiEventDispatcher {
             args.current.onFeatureOverridesChanged.addListener(
               SyncUiEventDispatcher._dispatchFeatureOverridesChange
             );
+          // istanbul ignore next
+          if (
```
GerardasB, line 275: Could we please get rid of these workarounds for the tests?

PR #429, opened 2023-07-27
Remove file write mode from test app

NancyMcCallB:
Brien pointed out that when creating the new test apps, we brought along part of the write-mode, but none of the testing. For now, removing write mode from the standalone app, since file mode does not figure into our testing.

### DIFF
--- a/test-apps/appui-test-app/connected/src/frontend/appui/ExternalIModel.ts
+++ b/test-apps/appui-test-app/connected/src/frontend/appui/ExternalIModel.ts
@@ -127,7 +127,7 @@ export class ExternalIModel {
           SampleAppIModelApp.loggerCategory(this),
           `openIModel (external): iTwinId=${this.iTwinId}&iModelId=${
             this.iModelId
-          } mode=${SampleAppIModelApp.allowWrite ? "ReadWrite" : "Readonly"}`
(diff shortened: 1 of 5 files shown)


GerardasB:
By testing we mean e2e tests or something is not working correctly in write-mode?

NancyMcCallB:
> By testing we mean e2e tests or something is not working correctly in write-mode

We used to have a frontstage that exposed tools and features for write mode. Since that functionality moved to another team, we had no need for a write mode -- it was primarily added for itwin.js core developers to use in testing. I did not see any reference to write mode in the e2e tests, but I may have missed it. Is there something about AppUI that you think requires write mode testing?

GerardasB:
> > By testing we mean e2e tests or something is not working correctly in write-mode
> 
> We used to have a frontstage that exposed tools and features for write mode. Since that functionality moved to another team, we had no need for a write mode -- it was primarily added for itwin.js core developers to use in testing. I did not see any reference to write mode in the e2e tests, but I may have missed it. Is there something about AppUI that you think requires write mode testing?

Ah, I see. Sure if there is another way for core developers to test editing we can get rid of this.

GerardasB (approved):

Issue #428, opened 2023-07-27
Remove NineZone term from iTwinjs 4.x public docs and property descriptions


YashodipD:
When I was reading about ToolbarUsage. I came across Ninezone term. It is very confusing if you are new developer and looking at it. NineZone UI pattern is used very long back in iTwinjs 2.x

Please remove / hide all the Ninezone related details in iTwinjs 4.x docs / property description.

Take a look at - https://www.itwinjs.org/reference/appui-react/toolbar/toolbarusage/

raplemie:
Documentation site was just updated to fix this!

PR #430, opened 2023-07-28
Fix toolbar doc

NancyMcCallB:
A colleague from User Success pointed out that we are using the obsolete term "ninezone" in our toolbar documentation. Changed the comment to be harvested into the API docs to read "AppUI."


YashodipD:
https://github.com/iTwin/appui/issues/428 - this issue was raised here.

GerardasB (approved):

PR #431, opened 2023-07-28
Visual testing for components

kconner1:
## Changes
Added visual testing for all component examples.

## Testing
To test:
1. Follow [README.md](https://github.com/iTwin/appui/blob/master/full-stack-tests/ui/README.md) instructions to run iTwin.js UI End-To-End Tests
2. Make changes to components and run the tests to make sure that the tests for the components you changed fail. When viewing a failed test's results, you should be able to see the expected, actual, and diff snapshots.


full-stack-tests/ui/test-files/04_Plant.i.ibim-hierarchies
grigasp, line 1: You probably want to add the `*-hierarchies` and `*.Tiles` files to `.gitignore` rather than commit them.

GerardasB (commented):
Looks great!
Could you let me know if you are using the prettier extension since there are some failures due to formatting?
You will need to update `playwright.config.ts` to start the backend as well (to use the standalone connection and open an iModel) OR use a blank connection instead (it should be enough to specify a `frontstage` URL param).

kconner1:
I think I forgot to run prettier because I just followed the Source Code Edit Workflow in the repository readme.

Could you explain why I need to start the backend in playwright.config.ts? I just ran the `npm run start:servers` script instead of the `npm run start:webserver` script in the [readme](https://github.com/iTwin/appui/blob/86dc3a2ee2d6bc586bcd7a402de0c99ad6c61610/full-stack-tests/ui/README.md?plain=1#L13). Couldn't I just update the readme?

full-stack-tests/ui/test-files/04_Plant.i.ibim-hierarchies
grigasp: @kconner1 I don't see a `.gitignore` update. Running the tests will generate these cache files and without them in `,gitignore` it's going to be very annoying for anyone running them locally.

GerardasB:
> I think I forgot to run prettier because I just followed the Source Code Edit Workflow in the repository readme.
> 
> Could you explain why I need to start the backend in playwright.config.ts? I just ran the `npm run start:servers` script instead of the `npm run start:webserver` script in the [readme](https://github.com/iTwin/appui/blob/86dc3a2ee2d6bc586bcd7a402de0c99ad6c61610/full-stack-tests/ui/README.md?plain=1#L13). Couldn't I just update the readme?

Yes, but you'd want to do that on the CI as well.

kconner1:
Changes to .gitignore and playwright.config.ts have been added

GerardasB (approved):

raplemie (approved):
I think masking the Components list in the two test would reduce the flakiness of these tests, otherwise, seems all good!

PR #432, opened 2023-07-31
Improve instructions for apps to use icons webfont

tcobbs-bentley:
## Changes

Updated the @bentley/icons-generic-webfont section of the 4.0.0 changehistory file to provide improved instructions for applications that need to continue to use the webfont.


tcobbs-bentley:
@NancyMcCallB The wording in this PR is different from my suggestion in Teams, but I feel it is an improvement.

NancyMcCallB (approved):

PR #434, opened 2023-08-02
Enable conditional properties in all Toolbars

### LINKED ISSUE #314: Make the AppUI's toolbar support showing/hiding a ToolbarItem with isHidden property set

We would like to have a toolbar that contains the items which are visible/hidden based on a specific condition (like the image below). We use the AppUI's Toolbar to make the toolbar but it seems that it still does not support showing/hiding a ToolbarItem with isHidden property set yet. 
Screenshot 2023-05-10 at 4 10 12 PM

Note: We have tried to use the ToolbarComposer but it does not meet our use case because it keeps trying to load/display the toolbar items from the UI providers only.

Make the Toolbar works similarly to the ToolbarComposer.

**Describe alternatives you've considered**
We have copied some pieces of code which manage the visibility of the toolbar items from the ToolbarComposer (e.g. copied the hook useToolbarItemSyncEffect, and use some internal APIs like ToolbarItemsChangedArgs, ToolbarItemsManager...) and used them in our code base. It works but we don't think it is a good solution though.

### LINKED ISSUE #424: Option for ToolbarComposer not to bring in addon items
The pop-up action toolbar in measure-tools-react was recently refactored and uses the ToolbarComposer. There's a regression with the old functionality where we get non-measurement items appearing in the popup.

I see the [ToolbarComposer ](https://github.com/iTwin/appui/blob/ff744cc48a55051ccf482481c20cc0c40b232ede/ui/appui-react/src/appui-react/toolbar/ToolbarComposer.tsx#L291C40-L291C62) queries for more items from UI providers, can we get an option to not do this and have our toolbar remain "private"?

For reference, the usage in measure-tools is in [PopupToolbar](https://github.com/iTwin/viewer-components-react/blob/master/packages/itwin/measure-tools/src/widgets/PopupToolbar.tsx) and used [here](https://github.com/iTwin/viewer-components-react/blob/080f31089da1c6b4b45915fc0b6383d43ef9f965/packages/itwin/measure-tools/src/widgets/MeasurementActionToolbar.tsx#L436).

### LINKED ISSUE #445: Measure tools action toolbar needs new ToolbarUsage type
Currently, the measurement actions toolbar is also populated with the frontstages group because the ToolbarUsage type ContentManipulation is too general pupose.

To solve this, it would be helpful to have a new ToolbarUsage type like 'Measurements'.

MicrosoftTeams-image (2)

raplemie:
## Next version
[docs/changehistory/NextVersion.md](https://github.com/iTwin/appui/compare/raplemie/toolbarHidden?expand=1#diff-4a7785af3b56bc69d2ad9f34257cd4a96cba53855aa039abeb761613d0f81988)

## Changes

* Moved conditional toolbar item properties handling to the `Toolbar` component, instead of the `ToolbarComposer`.
  * Deprecated `useUiItemsProviderToolbarItems`, replaced by internal `useActiveStageProvidedToolbarItems`.
  * Deprecated `useDefaultToolbarItems`, not needed.
* Moved the `ToolbarWithOverflow` overflow logic to the `Toolbar` component (with `enableOverflow` props to enable it) to consolidate the 2 toolbars.
  * Deprecated `ToolbarWithOverflow`.
* Split the behavior of `ToolbarItemsManager` into 2 different hooks. 
  * `useActiveToolIdSynchedItems`, ensure tool with id of activeToolId is the only active toolbar item.
  * `useConditionalSynchedItems`, ensure that conditionals are updated with SynchEvents.
  * Deprecated `ToolbarItemsManager`. (Kept because of public dependencies, which are deprecated as well)
* Provided default values for `ToolbarItemContext` (not directly related to the overall PR)
  * Removed _internal_ `ToolbarItemContext` in example code as the defaults are now provided and the elements can work without explicitly providing the context.
* Refactored many duplicate recursive function with generics to reduce number of lines. (like `addItemToSpecifiedParentGroup`, `nestedAddItemToSpecifiedParentGroup`...)
## Testing

* Validated that all unit tests passed.
* Created new unit tests for new code/refactor.
* Added to toolbar section of `ComponentsExampleProvider` so we can visual tests later on, and for immediate comparison between `ToolbarComposer`, `Toolbar` and `ToolbarWithOverflow`.

closes #314 
closes #424 
closes #445


common/api/appui-react.api.md
```diff
@@ -4665,7 +4667,7 @@ export enum ToolbarUsage {
     ViewNavigation = 1
 }
 
-// @beta
+// @beta @deprecated
```
GerardasB, line 4670: Need a codemod to change `ToolbarWithOverflow` to `Toolbar` with `enableOverflow` prop.

common/api/components-react.api.md
```diff
@@ -2938,14 +2938,8 @@ export interface ToolbarPopupContextProps {
...
 // @public @deprecated
-export interface ToolbarProps extends CommonProps, NoChildrenProps {
-    expandsTo?: Direction;
-    items: CommonToolbarItem[];
-    onItemExecuted?: OnItemExecutedFunc;
-    onKeyDown?: (e: React_2.KeyboardEvent) => void;
-    panelAlignment?: ToolbarPanelAlignment;
-    toolbarOpacitySetting?: ToolbarOpacitySetting;
-    useDragInteraction?: boolean;
+export interface ToolbarProps extends ToolbarWithOverflowProps {
```
GerardasB, line 2941: We are adding `overflowExpandsTo` prop to `ToolbarProps` this way.
It would make more sense to extend `ToolbarWithOverflowProps` from `ToolbarProps`.

common/api/appui-react.api.md
```diff
@@ -4650,10 +4650,12 @@ export interface ToolbarPopupProps extends PopupPropsBase {
 
 // @beta
 export interface ToolbarProps extends CommonProps, NoChildrenProps {
+    enableOverflow?: boolean;
```
GerardasB, line 4653: Should we move `expandsTo` option to `enableOverflow` to make API non-ambiguous, since it doesn't make sense to provide `expandsTo` if `enableOverflow` is `false`?

common/api/components-react.api.md
GerardasB: Seems like a feature addition to the already deprecated API, can't we avoid that?

ui/components-react/src/components-react/toolbar/Toolbar.tsx
```diff
@@ -48,15 +374,50 @@ export interface ToolbarProps extends CommonProps, NoChildrenProps {
   onItemExecuted?: OnItemExecutedFunc;
   /** Optional function to call on any KeyDown events processed by toolbar */
   onKeyDown?: (e: React.KeyboardEvent) => void;
+  /**
+   * @internal Use [Toolbar]($appui-react) from appui-react instead of using this property.
+   */
+  syncUiEvent?: BeEvent<(args: { eventIds: Set<string> }) => void>;
```
GerardasB, line 380: Could we avoid adding internal members to public interfaces as those never make sense?

GerardasB: How about existing consumers that have NOT moved away from this deprecated component? Do they need to use the internal `syncUiEvent` to correctly handle conditional stuff?

GerardasB (changes requested):

docs/changehistory/NextVersion.md
```diff
@@ -7,22 +7,40 @@ Table of contents:
...
 - Fixed `clearHideIsolateEmphasizeElements` and `hideElements` button display when view changes happen outside AppUI's API.
+- Fixed `Toolbar` and `ToolbarWithOverflow` to correctly handle `ConditionalStringValue`, `ConditionalBooleanValue` and `ConditionalIconValue` tool item properties.
+
+### Deprecations
+
+- `ToolbarWithOverflow` is now deprecated in favor of the `Toolbar` with `enableOverflow` prop set to `true`.
+- `useUiItemsProviderToolbarItems` is now deprecated in favor of the fixed `Toolbar` [^1].
+- `useDefaultToolbarItems` is now deprecated in favor of the fixed `Toolbar` [^1].
+
+[^1]: These hooks were exposed as **@public** and have `ToolbarItemsManager` **@internal** class as an argument, which contained the `Conditional*Value` refreshing logic. The logic have been cleaned up and now use simpler React hooks to handle the same behavior. As such, the `ToolbarItemsManager` is no longer used internally and will not be maintained. As the `Toolbar` component correctly handle the Conditionals, it should no longer be needed in any case.
```
GerardasB, line 33: It feels like this would benefit from rewording.

ui/components-react/src/components-react/toolbar/Toolbar.tsx
raplemie: We could avoid adding this internal members if we move the whole implementation to appui-react and make a copy of it, which I hope was the plan once we can actually get rid of the components-react implementation, but at this time, I think it's cleaner to add an internal prop. (One other option would be to simply NOT declare it at all and use it anyway, this is only a type and as it is internal, we could cast it so we know its potentially there.)

Existing consumers that have NOT moved away from this deprecated stuff shouldnt expect the new feature, and as such the component will behave like it did before. (which is, no conditionals are handled, which is what happen if you dont provide the event)

common/api/appui-react.api.md
raplemie: Are we planning to codemod every changes now ?

common/api/components-react.api.md
raplemie: Like I said, we can avoid that in 2 ways, duplicating code, or not using types at all.

If we extend the other way around, we will have an "enableOverflow" props to a component that is only meant to have that to true all the time, which wouldnt make much sense.

ui/components-react/src/components-react/toolbar/Toolbar.tsx
raplemie: A third option would be to have a completely internal toolbar implementation in components-react, which the public implementations from component-react and appui-react would use, maybe that's the cleanest.

docs/changehistory/NextVersion.md
raplemie: Feel free to suggest

ui/components-react/src/components-react/toolbar/Toolbar.tsx
raplemie: Moved the common implementation to `InternalToolbarComponent` which is now used by `Toolbar`, `ToolbarWithOverflow` in components react (reverted all props changes) and `Toolbar` from appui-react, using the new props. I'll update the enableOverflow in a separate commit but this comment should be resolvable now.

common/api/components-react.api.md
raplemie: Reverted all the changes to the components-react component props.

common/api/appui-react.api.md
GerardasB: I assume ideally we'd want to deliver a codemod for each deprecation.
We could then mention in the NextVersion.md to run the codemod to facilitate the upgrade :)
ATM we only have one version v4.0.0 so just go ahead and create a separate `v4.4.0` dir. 
Additionally, I was thinking it would make sense for 5.0 codemod to call all 4.x codemods :)

docs/changehistory/NextVersion.md
GerardasB, line 33:
```suggestion
[^1]: `useUiItemsProviderToolbarItems` and `useDefaultToolbarItems` hooks were using **@internal** `ToolbarItemsManager` class as an argument to facilitate the `Conditional*Value` refresh logic. Since the `Toolbar` component now correctly handles the conditionals, the deprecated hooks should no longer be needed in any case.
```

GerardasB: Added a suggestion below.

common/api/appui-react.api.md
```diff
@@ -3299,6 +3299,11 @@ export interface OpenChildWindowInfo {
 export class OpenMessageCenterEvent extends UiEvent<{}> {
 }
 
+// @beta
+export interface OverflowProps {
```
GerardasB, line 3303: I'd prefer to rename this to `ToolbarOverflowProps` to avoid a potential name clash in the future.

common/api/appui-react.api.md
raplemie: Created #446 to enable this support, learning this with a basic level of confidence for this PR will delay a bit too much.

common/api/appui-react.api.md
raplemie: Renamed to `OverflowToolbarOptions` to follow `CanFloatWidgetOptions` pattern.

GerardasB (approved):

Issue #426, opened 2023-07-26
ContextSubMenu's right icon looks huge


AAugustBentley:
`ContextSubMenu`'s `SvgCaretRight` icon looks huge compared to before.

4.2.0 version of AppUI:
![image](https://github.com/iTwin/appui/assets/129863171/546bedc0-844e-451d-80e7-ee58710a6916)

How it looked in 3.7.x version:
![image](https://github.com/iTwin/appui/assets/129863171/247a3dc0-709a-490d-bb85-42c917dc79c5)

It is possible that it looked alright in 4.1.0 version but I do not have screenshots from that time to check.

raplemie:
Hi, we'll look into this! Thanks for the report!

raplemie:
Hi @AAugustBentley, a fix for this was provided in 4.4.0 which was released today, let us know if the result is correct!

PR #438, opened 2023-08-03
Decrease and center ContextSubMenuArrow

joehenry9498:
## Changes
- Decrease arrow size and center it in ContextSubMenu

## Testing
- I tested the changes by running the storybook and seeing how the ContextMenu and ContextSubMenu looked in there.

Had to create a new pull request for this because there were git problems with the CLA

resolves https://github.com/iTwin/appui/issues/426

### DIFF
--- a/common/api/components-react.api.md
+++ b/common/api/components-react.api.md
@@ -192,6 +192,9 @@ export class BooleanTypeConverter extends TypeConverter {
     sortCompare(a: Primitives.Boolean, b: Primitives.Boolean, _ignoreCase?: boolean): number;
 }
 
+// @internal (undocumented)
+export function buildPropertyFilter(groupItem: PropertyFilterBuilderRuleGroupItem): PropertyFilter | undefined;
+
 // @public
 export interface CategorizedPropertyItem extends FlatGridItemBase {
     // (undocumented)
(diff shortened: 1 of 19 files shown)


ui/core-react/src/core-react/contextmenu/ContextMenu.scss
```diff
@@ -221,6 +221,8 @@ $uicore-menu-top-bottom-padding: calc(#{$uicore-menu-padding} * 0.6667);
         font-size: calc(var(--iui-font-size-2) * 0.5);
         line-height: var(--iui-size-l);
         text-align: center;
+        position: relative;
```
raplemie, line 224: From what I see, most of the tweaks from this class were for the previous "webfont" mode, I think we can simplify all of this now by removing everything but 
```scss
{
  margin-left: $uicore-unbordered-padding;
  margin-top: var(--iui-size-3xs);
}
```
and get the same result with less "What's going on" when looking at this class.

raplemie (changes requested):

ui/core-react/src/core-react/contextmenu/ContextMenu.scss
joehenry9498: Looks good in storybook with that simplified styling. Changes pushed

raplemie:
Please add an entry in NextVersion.md (should be really similar to what was entered in the changes, but we will likely release that as part of 4.4.0 and NextVersion is the release document in this case.

raplemie (approved):
LGTM! Thanks!

PR #444, opened 2023-08-07
Clear rule values when rule operator is changed

simasjar:
## Changes

Keep the rule value the same when the operator changes between `Less`, `LessOrEqual`, `Greater`, `GreaterOrEqual` and `IsEqual`, `IsNotEqual`, otherwise clear it.

closes iTwin/presentation/issues/203
## Testing

Added tests to check if the rule value changes to undefined or is the same after the rule operator changes.


ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
```diff
@@ -158,6 +158,15 @@ export class PropertyFilterBuilderActions {
       const rule = findRule(state.rootGroup, path);
       if (!rule) return;
       if (isUnaryPropertyFilterOperator(operator)) rule.value = undefined;
+      if (
+        (!isOperatorEqualOrIsNotEqual(rule.operator) ||
+          !isOperatorEqualOrIsNotEqual(operator)) &&
+        (!isInequalityOperator(rule.operator) ||
+          !isInequalityOperator(operator)) &&
+        operator !== rule.operator
+      ) {
```
saskliutas, line 167: This condition is quite hard to understand with all these `!` operators. Maybe you could add one more function that would check if operators are similar (both operators are `equal/notEqual` or both operators are one of the `>`, `>=`, `<`, `<=`). That way you could simplify this condition:
```suggestion
      if (operator !== rule.operator && !areOperatorsSimilar(operator, rule.operator)) {
```

saskliutas (changes requested):

ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
simasjar: fixed

NancyMcCallB (approved):

PR #452, opened 2023-08-10
openCursorMenu now works with React 18 Strict Mode

### LINKED ISSUE #375: Context menu is not opened when using React 18 `<StrictMode>`
**Describe the bug**
`IModelApp.uiAdmin.showContextMenu` doesn't open context menu with React 18 when using `<StrictMode>`. `IModelApp.uiAdmin.showContextMenu` return `true` but context menu is not visible.

- OS: Windows 11
- Browser: Electron
- Version: 24
- iTwin.js AppUI Version: 4.0.0
- iTwin.js Version: 4.0.0

joehenry9498:
## Changes
- Fixed openCursorMenu function so that it works with React 18 Strict Mode. The initial complaint was about the showContextMenu function but that uses openCursorMenu so more stuff was broken than initially thought.
- Variable wasn't being set back to it's original state after it was done being used. This caused Strict Mode to not allow the cursor menu to open due to the new double rendering introduced to Strict Mode in React 18.

## Testing
- First I updated the project to React 18. This involves manually changing the index file to use "createRoot" instead of ReactDOM.render. Once you rebuild, you can test on the Component Examples frontstage under Cursor. Before, if strict mode was wrapped around the app like it normally is, clicking 'Open Cursor Popup Menu' would do nothing. However, if you took strict mode out, clicking it would open up the context menu. Now the context menu should open regardless if the app is wrapped in strict mode or not.

Resolves #375


ui/appui-react/src/appui-react/cursor/cursormenu/CursorMenu.tsx
```diff
@@ -77,6 +77,7 @@ export class CursorPopupMenu extends React.PureComponent<
   public override componentWillUnmount() {
     this._componentUnmounting = true;
     SyncUiEventDispatcher.onSyncUiEvent.removeListener(this._handleSyncUiEvent);
+    this._componentUnmounting = false;
```
raplemie, line 80: It seems like this would pretty much cause the `this._componentUnmounting` to be quite useless if it is `false` only in the fraction of a second while we remove the listener, as the need for this would be for a race condition in the first place (the listener is triggered before we unmount, but executed after we unmounted, in which case, the value would still be `false` and cause the same issue as if it wasnt checked at all)

I think the `... = false` would make more sense in `componentDidMount` (and probably rename it to `this._isMounted` and invert the logic)

joehenry9498: Resolved

PR #455, opened 2023-08-14
Push back deprecation for codemod availability

raplemie:
## Changes

Removed deprecation tag on `ToolbarWithOverflow` component, (we should add it back when we have the proper codemod available, until then, both component do the same thing and will not cause issues to the users anyway.

Removed line in NextVersion as the component is no longer deprecated.

## Testing

N/A

### DIFF
--- a/common/api/appui-react.api.md
+++ b/common/api/appui-react.api.md
@@ -4671,7 +4671,7 @@ export enum ToolbarUsage {
     ViewNavigation = 1
 }
 
-// @beta @deprecated
+// @beta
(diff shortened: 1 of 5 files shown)


GerardasB (approved):
Should we revert hook deprecation as well for now?

raplemie:
> Should we revert hook deprecation as well for now?

These hooks have no replacement except the toolbar and should not be used (their only potential usage was to attempt to replicate what the Toolbar is now doing automatically) I dont intend to provide codemod for their removals as the how/why they are used have no direct replacement, so I think they can be deprecated right away.

Issue #462, opened 2023-08-23
Browser locks up when TreeSelectionManager gets a shift-click for a large number of tree nodes (>200)


wilmaier:
**Describe the bug**
The application freezes and the tab needs to be refreshed.
Selecting everything by parent node works without issues. This could be related to lazy loading because if I scroll down very slowly and let every node get loaded before selecting the last element then the app does not freeze.

**To Reproduce**

1. Open IModel that would have a few hundred elements
2. Type 'a' into the search widget field
3. Expand node with a large amount of elements
4. Select the first one and scroll down to the bottom and shift+click the last element.

**Expected behavior**
Clicking on the parent node causes all of the child nodes to be added to selection set.
Using shift click to select all child nodes should work the same.

**Desktop (please complete the applicable information):**

- OS: Windows 11
- Browser: Chrome 
- Version 116.0.5845.110 (Official Build) (64-bit)
- iTwin.js AppUI Version 4.4.0
- iTwin.js Version 4.1.0

grigasp:
Adding repro steps that are easier to understand:

1. Go to [presentation rules editor](https://presentationruleseditor.bentley.com/open-imodel?iTwinId=b27dc251-0e53-4a36-9a38-182fc309be07&iModelId=f30566da-8fdf-4cba-b09a-fd39f5397ae6#editor/N4IgTgrgNgpgzjALiAXCYAdAdgAhxkASwBMCV8QAlaeJARgIBps8DJY4ycBtFvHTLn6twNACoBPAA4wuBSgHsFiAHILi8Jn2EE4MgMaEAZoX0BDRIQVZOqHtuECHjinpj7JMuSACSNxGZY+jBqGnAA8kYAygbGpgDCUGZwCLbMQi4i+kkpmnaCmY66+gAWMAC2Zipm5bJ2BABChHDxCmB16YU6INnJcNW1tuS8GV0iAKKwtViIBM6ZALrzwgC+nV0EZu0ACgpQEuVtUiWmXIiQMOuFBADmYAoQUg0SiX1cRmZQCFeZt-ePzwAMmYAEYwKDvT4IZY4FbzJajOFCBErEArIA).
2. Select the first node in the tree on the right.
3. Click on the scrollbar handle and quickly drag it to the bottom.
4. While holding the `Shift` key, click on one of the nodes.
5. Notice the frontend is hung.

[...]

PR #463, opened 2023-08-25
ControlledTree: avoid hanging browser with range selection

saskliutas:
Closes https://github.com/iTwin/appui/issues/462

Browser hang was caused by infinite loop in TreeEventDispatcher:
```
generate({
  initialState: deselectedItems,
  iterate: () => [],
  scheduler: asapScheduler,
})
```

Changed it to not use `generate` and determine if `deselectedItems` array should be emitted based on `selectedNodeItems` observable value index.

Made some additional changes:
* Bumped `rxjs` version to the latest. Refactored usage of deprecated operators and changed imports to the suggested approach `import { operator } from "rxjs";`
* Refactored `TreeEventDispatcher` tests because they were hard to understand and using internal `TreeEventDispatcher` parts.
* Refactored `TreeEventDispatcher` access to `VisbileTreeNodes`. Now `VisibleTreeNodes` getter is always passed to `TreeEventDispatcher` contructor and is always defined. This allows to remove a lot of `undefined` checks in `TreeEventDispatcher` and `TreeSelectionManager`
* Reduced amount of `if` statements and nesting in `TreeSelectionManager` keyboard navigation handling code.

### DIFF
--- a/ui/components-react/src/components-react/tree/controlled/TreeEventDispatcher.ts
+++ b/ui/components-react/src/components-react/tree/controlled/TreeEventDispatcher.ts
@@ -12,14 +12,12 @@ import {
   concat,
   concatAll,
   concatMap,
-  defaultIfEmpty,
   defer,
   distinctUntilChanged,
   EMPTY,
   filter,
   finalize,
   from,
-  generate,
   map,
   merge,
   mergeAll,
@@ -28,7 +26,6 @@ import {
   share,
   subscribeOn,
   toArray,
-  zip,
 } from "rxjs";
 import type { CheckBoxState } from "@itwin/core-react";
 import type { SelectionMode } from "../../common/selection/SelectionModes";
@@ -283,21 +280,16 @@ export class TreeEventDispatcher implements TreeActions {
   }> {
     const deselectedItems = this.collectNodeItems(deselection);
     if (isRangeSelection(selection)) {
-      return zip(
-        this.collectNodesBetween(selection.from, selection.to).pipe(
-          defaultIfEmpty([] as TreeNodeItem[])
-        ),
-        // Without a scheduler `generate` would block the main thread indefinitely
-        generate({
-          initialState: deselectedItems,
-          iterate: () => [],
-          scheduler: asapScheduler,
+      let firstEmission = true;
+      return this.collectNodesBetween(selection.from, selection.to).pipe(
+        map((selectedNodeItems) => {
+          if (firstEmission) {
+            firstEmission = false;
+            return { selectedNodeItems, deselectedNodeItems: deselectedItems };
+          }
+
+          return { selectedNodeItems, deselectedNodeItems: [] };
         })
-      ).pipe(
-        map(([selectedNodeItems, deselectedNodeItems]) => ({
-          selectedNodeItems,
-          deselectedNodeItems,
-        }))
       );
     }
 
--- a/ui/components-react/src/components-react/tree/controlled/TreeEventDispatcher.ts
+++ b/ui/components-react/src/components-react/tree/controlled/TreeEventDispatcher.ts
@@ -6,28 +6,30 @@
  * @module Tree
  */
 
-import type { Observable } from "rxjs/internal/Observable";
-import { concat } from "rxjs/internal/observable/concat";
-import { defer } from "rxjs/internal/observable/defer";
-import { EMPTY } from "rxjs/internal/observable/empty";
-import { from } from "rxjs/internal/observable/from";
-import { generate } from "rxjs/internal/observable/generate";
-import { merge } from "rxjs/internal/observable/merge";
-import { of } from "rxjs/internal/observable/of";
-import { zip } from "rxjs/internal/observable/zip";
-import { concatAll } from "rxjs/internal/operators/concatAll";
-import { concatMap } from "rxjs/internal/operators/concatMap";
-import { defaultIfEmpty } from "rxjs/internal/operators/defaultIfEmpty";
-import { distinctUntilChanged } from "rxjs/internal/operators/distinctUntilChanged";
-import { filter } from "rxjs/internal/operators/filter";
-import { finalize } from "rxjs/internal/operators/finalize";
-import { map } from "rxjs/internal/operators/map";
-import { mergeAll } from "rxjs/internal/operators/mergeAll";
-import { publishReplay } from "rxjs/internal/operators/publishReplay";
-import { refCount } from "rxjs/internal/operators/refCount";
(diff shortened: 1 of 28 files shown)


ui/components-react/src/test/tree/controlled/TreeEventDispatcher.test.ts
```diff
@@ -3,604 +3,823 @@
...
+        .to.have.lengthOf(2)
+        .and.containSubset([
+          {
+            selectedNodeItems: [{ id: "A" }, { id: "B" }],
+            deselectedNodeItems: [],
+          },
+          {
+            selectedNodeItems: [{ id: "C" }],
+            deselectedNodeItems: [],
+          },
```
grigasp, line 417: interesting - why C is separate from A and B?

ui/components-react/src/test/tree/controlled/TreeEventDispatcher.test.ts
```diff
@@ -3,604 +3,823 @@
...
+        ]);
 
-      expect(spy).to.be.calledOnce;
-      const changes = spy.args[0][0] as TreeCheckboxStateChangeEventArgs;
-      const results = await extractSequence(rxjsFrom(changes.stateChanges));
-      expect(results).to.not.be.empty;
-      const affectedNodeItems = results[0].map((change) => change.nodeItem);
-      expect(affectedNodeItems).to.be.deep.eq(expectedAffectedNodeItems);
+      expect(treeEvents.onSelectionModified).to.be.calledTwice;
+      const [args1] = treeEvents.onSelectionModified.args[1];
```
grigasp, line 421: what causes `onSelectionModified` to be called the second time?

ui/components-react/src/test/tree/controlled/TreeEventDispatcher.test.ts
```diff
@@ -3,604 +3,823 @@
...
+      const { dispatcher } = setupTreeEventDispatcher(
+        SelectionMode.Multiple,
+        (model) => {
+          model.setChildren(undefined, [createTreeNodeInput("A")], 0);
+        }
+      );
 
-      dispatcher.onNodeCheckboxClicked(selectedNodes[0].id, CheckBoxState.On);
+      dispatcher.onNodeCheckboxClicked("B", CheckBoxState.On);
+      expect(treeEvents.onCheckboxStateChanged).to.not.be.calledOnce;
```
grigasp, line 490:
```suggestion
      expect(treeEvents.onCheckboxStateChanged).to.not.be.called;
```

ui/components-react/src/test/tree/controlled/TreeEventDispatcher.test.ts
```diff
@@ -3,604 +3,823 @@
...
+      (model) => {
+        model.setChildren(undefined, [createTreeNodeInput("A")], 0);
+      }
+    );
+    dispatcher.onNodeCollapsed("A");
+    expect(treeEvents.onNodeCollapsed).to.be.calledOnceWith({ nodeId: "A" });
+  });
+
+  describe("keyboard navigation", () => {
+    it("forward events to `TreeSelectionManager", () => {
```
grigasp, line 778:
```suggestion
    it("forwards events to `TreeSelectionManager`", () => {
```

ui/components-react/src/test/tree/controlled/TreeEventDispatcher.test.ts
grigasp: I see similar situations in other tests as well - could you please just describe when/why `onSelectionModified` are fired on drag selection?

grigasp (changes requested):

ui/components-react/src/components-react/tree/controlled/TreeEventDispatcher.ts
```diff
@@ -97,8 +97,12 @@ export class TreeEventDispatcher implements TreeActions {
...
           ),
           concatAll(),
-          publishReplay(),
-          refCount()
+          share({
+            connector: () => new ReplaySubject(),
+            resetOnError: false,
+            resetOnComplete: false,
+            resetOnRefCountZero: true,
+          })
```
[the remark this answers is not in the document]
saskliutas: Change to `shareReplay`. Interesting that these docs didn't suggest using `shareReplay` https://rxjs.dev/deprecations/multicasting#publishreplay

ui/components-react/src/test/tree/controlled/TreeEventDispatcher.test.ts
saskliutas: This is happening because `SelectionHandler` does not invoke `updateSelection` if there are no selection changes: https://github.com/iTwin/appui/blob/71ab57c9416bb598688d13f47bff3563489d30ac/ui/components-react/src/components-react/common/selection/SelectionHandler.ts#L454

And it detects changes only when `updateDragAction` is called with different item that previous or initial. So first emission will always contains the initial item for which `DragAction` was created and first different item that is handled by `DragAction`.

ui/components-react/src/test/tree/controlled/TreeEventDispatcher.test.ts
saskliutas: When drag selection is in progress `SelectionHandler` is accumulating all changes in `BatchSelectionOperation` https://github.com/iTwin/appui/blob/71ab57c9416bb598688d13f47bff3563489d30ac/ui/components-react/src/components-react/common/selection/SelectionHandler.ts#L461
and when drag selection is finished items are `selected/deselected` based on changes accumulated in that operation https://github.com/iTwin/appui/blob/71ab57c9416bb598688d13f47bff3563489d30ac/ui/components-react/src/components-react/common/selection/SelectionHandler.ts#L395

I guess this stands as source of truth of what was `selected/deselect` during drag operation. I see it makes separate calls for selected and deselected items so in other test `onSelectedModified` ends up called three times.

ui/components-react/src/test/tree/controlled/TreeEventDispatcher.test.ts
saskliutas: Yes

ui/components-react/src/test/tree/controlled/TreeEventDispatcher.test.ts
grigasp: thanks, I think I understand this now (comments that you added to the tests are really helpful)

grigasp (approved):

raplemie (approved):
Glad we got rid of all those "/internal/" imports ! :)

PR #467, opened 2023-08-29
Filter builder: Ability to build filter ignoring errors

saskliutas:
Add ability to get filter without persisting encountered errors in a state. This is useful if component want to do some actions with filter but does not want to show errors in UI.

Prerequisite for https://github.com/iTwin/presentation/issues/221


grigasp (approved):

raplemie (approved):
LGTM!

PR #477, opened 2023-09-05
ControlledTree: Do not add loaded nodes to model after cancellation

saskliutas:
Closes https://github.com/iTwin/presentation/issues/161

After `SubscriptionScheduler` subscribes to observable it waits until it completes even if there are no more subscribers. Because of this `updateModel` call should be done after observable is scheduled to make sure that `loadNode` could be cancelled when all subscribes unsubscribes and nodes from this cancelled request does not end up in `TreeModel`


grigasp (approved):

raplemie (approved):
LGTM!

PR #479, opened 2023-09-05
Tool Settings widget can now be set to no docking allowed

joehenry9498:
## Changes
- allowPanelTargets now returns a blank array instead of undefined when a blank array is inputted into the setter. This allows the logic to work properly for docking to not be allowed

## Testing
- Tested with the full stack tests and looking at the widgets results in the standalone


joehenry9498:
More work needs to be done on this, noticing a lot of bugs and inconsistencies with how the widgets are docking with and without these changes. Working on it now

wilmaier:
Since the "Send back" button is not associated with any of the allowed panels, it should remain no matter what that is set to.  The tool settings dock to a top panel which is separate from the docking locations.

raplemie:
> Since the "Send back" button is not associated with any of the allowed panels, it should remain no matter what that is set to. The tool settings dock to a top panel which is separate from the docking locations.

The tool settings is a unique case in this and have a "Dock to top" button, the other widgets have a "Send Back" button which will send the widget to its last known panel, and in this case, it wouldnt have any, so it shouldnt be shown.

joehenry9498:
Changes have been implemented. You create a widget with an empty array for panelsAllowed to not allow it to dock. Changes have been documented in the props. If undockable widget is combined with dockable widget, send to back button disappears for both.

raplemie (changes requested):

ui/appui-layout-react/src/appui-layout-react/widget/Buttons.tsx
```diff
@@ -18,12 +18,32 @@ import {
...
 /** @internal */
 export function TabBarButtons() {
   const isToolSettings = useIsToolSettingsTab();
   const floatingWidgetId = useFloatingWidgetId();
+  const canBeDocked = useWidgetAllowedToDock();
+  const allTabsCanBeDocked = useLayout((state) => {
+    let noDockingCount = 0;
+    if (floatingWidgetId) {
+      const widget = getWidgetState(state, floatingWidgetId);
+      widget.tabs.forEach((tab) => {
```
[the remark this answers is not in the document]
joehenry9498: Removed this function

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -286,6 +290,10 @@ export class WidgetDef {
       this._defaultState = config.defaultState;
     }
 
+    if (config.allowedPanels && config.allowedPanels.length === 0) {
+      this._defaultState = 3;
+    }
+
```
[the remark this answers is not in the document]
joehenry9498: Changes pushed. Let me know if the closed logic is as expected

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -470,7 +478,11 @@ export class WidgetDef {
     targets: ReadonlyArray<StagePanelLocation> | undefined
   ) {
     this._allowedPanelTargets =
-      targets && targets?.length > 0 ? targets : undefined;
+      targets && targets?.length > 0
+        ? targets
+        : targets?.length === 0
+        ? []
+        : undefined;
```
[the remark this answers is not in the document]
joehenry9498: Fixed

joehenry9498:
In the process of fixing playwright tests that are messing the windows build up

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -213,7 +213,15 @@ export class WidgetDef {
...
 
-    const canFloat = config.canFloat;
+    let canFloat = config.canFloat;
+    if (config.allowedPanels && config.allowedPanels.length === 0) {
+      if (config.canFloat !== true) {
+        canFloat = config.canFloat;
+      } else {
+        canFloat = true;
+      }
+    }
```
raplemie, line 223: The logic appears to be backwards here, this code says: if we have `allowedPanels` an empty array, and `canFloat` is not `true`, leave it as is, otherwise (it is already `true`, force it to `true) ?

Is this setting even read anywhere if your tests worked ?

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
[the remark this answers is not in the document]
raplemie: Seems to be good, are the "closed" appearing correctly ?

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
joehenry9498: config.canFloat can also be an object that is the container id of the floating object. So the logic is as this, if we have an empty allowed panels array check if the config.canFloat is not a true boolean. If its false, we want to set it to true anyway, but if a config.canFloat object was passed in, we want to set can float to that

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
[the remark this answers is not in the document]
joehenry9498: appears so to me

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
raplemie: Ok, then there is still an issue, because `false` or `undefined` will use the `canFloat !== true` true branch. The test should probably `typeof  config.canFloat === "object"`

NancyMcCallB: Currently, the documented behavior is that an empty allowedPanels array means that all the panels are allowed. This was for backward compatibility.

joehenry9498: Yeah I see what you mean now, on it

joehenry9498: > Currently, the documented behavior is that an empty allowedPanels array means that all the panels are allowed. This was for backward compatibility.



> Currently, the documented behavior is that an empty allowedPanels array means that all the panels are allowed. This was for backward compatibility.

Passing in nothing at all so that it is undefined means that all panels are allowed now. Whilst passing in a blank array will restrict docking. Does something need to be changed to account for backwards compatibility? Or does the documentation need changed further? @NancyMcCallB

joehenry9498: > Ok, then there is still an issue, because `false` or `undefined` will use the `canFloat !== true` true branch. The test should probably `typeof config.canFloat === "object"`

This logic has been fixed

raplemie (approved):
LGTM!
Thanks!

PR #491, opened 2023-09-14
Fix toolsetting dropdown closing on UI hides

### LINKED ISSUE #461: Dropdown closes when UI auto-hides in Tool Settings when widget is docked
When tool settings are docked, clicking on a dropdown shows its choices but they disappear after a short while when the UI auto hides. The dropdown works just fine when tool settings are undocked/in a pop-up.

image

To create the dropdown, we send DialogItem with enum.choices in the property object.

image

raplemie:
## Changes

Removed a `key={Date.now()}` prop that forced the full refresh of the Docked toolbar component whenever react was testing if it needed to. Changed that to a key that only updates when the "settings" array change (as we dont have better way of guessing that the content should be updated)
Without the key, the refreshing is not happening when triggered by the events as we use index for keys in the underlying components, but we dont have anything else to base our "refresh" requirement on since we are given React node, and not component. There might be a better solution, but this one should at the very least fix the issue and reduce the number of full refresh of the tool settings.

## Testing

Added new tool with dynamic settings in AppUiTestProviders (this is what required the removed `key` props)
Added new full-stack-tests ensuring that the dynamic tool was refreshing correctly.

## Issue
resolves #461


ui/appui-react/src/appui-react/widget-panels/ToolSettings.tsx
```diff
@@ -74,17 +74,23 @@ export function WidgetPanelsToolSettings() {
...
+    setForceRefreshKey(Date.now());
+  }, [settings]);
   // for the overflow to work properly each setting in the DockedToolSettings should be wrapped by a DockedToolSetting component
   return (
     <DockedToolSettings
       itemId={
         InternalFrontstageManager.activeToolSettingsProvider?.uniqueId ?? "none"
       }
-      key={Date.now()}
+      key={forceRefreshKey}
```
GerardasB, line 89: Can we just get rid of the key or do we really need to re-mount the component?

ui/appui-react/src/appui-react/widget-panels/ToolSettings.tsx
```diff
@@ -74,17 +74,23 @@ export function WidgetPanelsToolSettings() {
...
     >
       {settings &&
         settings.map((entry, index) => (
-          <DockedToolSetting key={index}>
+          <DockedToolSetting key={`${index}${entry.editorNode?.toString()}`}>
```
GerardasB, line 93: Why is this needed? It will be `[object Object]` most of the time..

GerardasB (changes requested):

ui/appui-react/src/appui-react/widget-panels/ToolSettings.tsx
raplemie: Yep, it was an attempt at not requiring the topmost key, but shouldnt have been left (I tried JSON.stringify() to get the actual content, but since these are rendered react node, I get circular objects...)

ui/appui-react/src/appui-react/widget-panels/ToolSettings.tsx
raplemie: We need to remount the component because we are only placing pre rendered component in there, and since they are only indexed by key underneath, if the node changes they are not rerendered. So the content of the node change, if the event that change them is known by the node (by a Hook for example), but if the whole node is changed, nothing happens because they all have the same key, and since we dont have any insight of what is in that node, there isnt much we can do about it.

Maybe we could redesign the whole system, and build a key out of the toolsettings values/editor, but that would likely break more stuf that fix, and require dev updating their stuff, so I went with this as we have 2 people waiting on this.

common/changes/@itwin/appui-react/raplemie-toolSettingsRefreshOnHideUi_2023-09-14-20-32.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "Fix an issue with dropdown closing when main UI hides.",
```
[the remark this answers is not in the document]
raplemie: Added

ui/appui-react/src/appui-react/widget-panels/ToolSettings.tsx
raplemie: removed unneeded part

PR #492, opened 2023-09-15
react to core/appui-abstract deprecations

ben-polinsky:
Reacting to the following deprecations in @itwin/appui-abstract:

- `PointProps`
- `UiEvent`
- `UiEventDispatcher`

The `isLetter` utility method has been removed (it was @alpha and did not need to be deprecated)

In service of this:
- `UiEventDispatcher` has been folded into `SyncUiEventDispatcher`
- `UiSyncEventArgs` and `UiSyncEvent` have also been moved to appui-react.
- `UiEvent` has been deprecated in favor of `BeUiEvent` since the two are equivalent.
- appui packages already consuming @itwin/core-geometry now use its `XAndY` interface instead of `PointProps` as they are equivalent
- appui-layout-react has an inline implemnetation of XAndY to avoid the breaking peer dependency change which adding core-geometry would require.

Accompanying PR in core is here: https://github.com/iTwin/itwinjs-core/pull/5998


ui/core-react/src/core-react/utils/Point.ts
```diff
@@ -6,38 +6,38 @@
  * @module Utilities
  */
 
-import type { PointProps } from "@itwin/appui-abstract";
+import type { XAndY } from "@itwin/core-geometry";
```
raplemie, line 9: There is something off in the `package.json` because `core-react` do not depend on `core-geometry`, yet it is there in `devDependencies`. On top of that, these are internal classes, and we already have defined `PointProps` in `popup/Popup.tsx`, can you move the definition in a `uils/PointProps.tsx`, but do not put that file in the barrel export, so it remains internal.

ui/appui-react/src/appui-react/syncui/SyncUiEventDispatcher.ts
```diff
@@ -66,91 +64,182 @@ export enum SyncUiEventId {
...
 /** This class is used to send eventIds to interested UI components so the component can determine if it needs
  * to refresh its display by calling setState on itself.
  * @public
  */
 export class SyncUiEventDispatcher {
-  private static _uiEventDispatcher = new UiEventDispatcher();
   private static _unregisterFuncs = new Array<() => void>();
   private static _connectionUnregisterFuncs = new Array<() => void>();
   private static _iModelConnection?: IModelConnection;
+  private static _syncEventTimerId: number | undefined;
```
raplemie, line 88: This is going backwards though, we want to move toward less statics, you can move the code to something like `InternalSyncUiEventDispatcher.tsx` and likely leave it as it is currently in `appui-abstract`, we will want to move more of the logic to that class and remove the specifics from the static class rather that the other way around.

raplemie (changes requested):

ui/appui-react/src/appui-react/syncui/SyncUiEventDispatcher.ts
ben-polinsky: I will take a look, thanks.

ui/core-react/src/core-react/utils/Point.ts
```diff
@@ -6,19 +6,19 @@
...
 /** Describes and provides methods to work with 2d points.
  * @internal
  */
 export class Point implements PointProps {
-  /** Creates point from [[PointProps]]. */
-  public static create(pointProps: PointProps) {
-    return new Point(pointProps.x, pointProps.y);
+  /** Creates point from PointProps */
+  public static create(PointProps: PointProps) {
+    return new Point(PointProps.x, PointProps.y);
```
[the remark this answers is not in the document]
ben-polinsky: Whoops - yeah bad find and replace.

ui/appui-react/src/appui-react/syncui/InternalSyncUiEventDispatcher.ts
```diff
@@ -0,0 +1,151 @@
...
+ */
+export interface UiSyncEventArgs {
+  eventIds: Set<string>;
+}
+
+/** UiSync Event class.
+ * @public
+ */
+export class UiSyncEvent extends BeUiEvent<UiSyncEventArgs> { }
+
```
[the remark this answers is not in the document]
ben-polinsky: Yes yes

ben-polinsky:
> You'll need to run `rush update --variant core-3x` to update the other `pnpm-lock.yaml` file.

Done. Thanks

common/api/appui-react.api.md
```diff
@@ -3705,7 +3705,7 @@ export enum SelectionScope {
 }
 
 // @public
-export const SelectionScopeField: ConnectedComponent<typeof SelectionScopeFieldComponent, Omit_3<SelectionScopeFieldProps, "availableSelectionScopes" | "activeSelectionScope">>;
+export const SelectionScopeField: ConnectedComponent<typeof SelectionScopeFieldComponent, Omit_3<SelectionScopeFieldProps, "activeSelectionScope" | "availableSelectionScopes">>;
```
raplemie, line 3708: Try reverting that specific change, we've had issues with that declaration for a long time, where it does something in one place and a different one in a different place. (This is usually "solved" by going directly to that package folder,
* `npm run clean`
* `npm run build`
* `npm run extract-api`

ui/appui-react/src/appui-react/syncui/InternalSyncUiEventDispatcher.ts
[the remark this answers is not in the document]
ben-polinsky: Thanks fixed.

common/api/appui-react.api.md
ben-polinsky: Thanks, fixed.

raplemie (approved):
LGTM!

PR #493, opened 2023-09-15
Bump iTwin UI package to get ProgressRadial fix

joehenry9498:
## Changes
- Bumped iTwinUI package version so that fix for ProgressRadial goes in. This will effect the controlled tree components so that the spinner aligns when loading.

## Testing
- Test with the Controlled Tree component in Component Examples Provider. I set the node to always be loading in Node.tsx Line 174 and then colored the background of it red, to really see the changes. The spinner should stay inside container in tree component

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -45,6 +45,9 @@ common/autoinstallers/*/.npmrc
 # storybook specific
 storybook-static
 
+# playwright specific
+test-results
+
 .env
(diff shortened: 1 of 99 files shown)


raplemie (approved):
There is an issue with ToggleSwitch contour, but that do not appeart to be new, I asked iTwinUI for insights.

PR #499, opened 2023-09-20
Correctly set `resizable` flag of a floating widget

### LINKED ISSUE #497: Widget setting "canFloat: { isResizable: false }" not doing anything
**Describe the bug**
When configuring widget and setting `canFloat: { isResizable: false }` nothing happens.

**To Reproduce**

1. Go to https://itwin.github.io/appui/495/?path=/docs/widget-canfloat-options--docs
1. Change setting `isResizable` to "False"
1. Move widget around (don't dock it)
1. Resize widget

**Expected behavior**
When configuring widget and setting `canFloat: { isResizable: false }` widget should not be resizable. 

**Screenshots**

- OS: Windows 11
- Browser: Edge
- Version: 117
- iTwin.js AppUI Version: 4.5.0 (but reproducible in storybook)
- iTwin.js Version 4.1.6

GerardasB:
## Changes

Fixes #497 by correctly configuring `resizable` flag of a floating widget after it is docked and undocked.

## Testing

Tested via storybook. Added a unit test.


raplemie:
There is something off when we merge a non resizable widget with a resizable widget. When you do that, and move to the non-resizable tab, the tab is resizable but the size is kept for the tab if it is poped out again, but it is now unresizable, with the new size.

https://github.com/iTwin/appui/assets/1904889/cf4c4e12-c208-4e79-883e-89d97a150621

I think when the window is not resizable, we should always keep the initialConfig floating size (when unmerging)

A larger question would be, does it make sense that a non resizable widget can merge in a container of a different size than its own and/or is resizable ?

GerardasB:
> I think when the window is not resizable, we should always keep the initialConfig floating size (when unmerging)

Made a fix for that: floating widget size will not persist if a widget is not resizable.

> A larger question would be, does it make sense that a non resizable widget can merge in a container of a different size than its own and/or is resizable ?

Alternatively, we could toggle the size of a floating widget when switching tabs. But that has its own sub-set of issues/cases to discuss.

raplemie:
The fix is working well when you explicitly drag out the non resizable tab, but if it remains alone in its container, it will not "switch" back to his non resizable size.

https://github.com/iTwin/appui/assets/1904889/3b303314-dba3-4211-bc18-344278fb954c

>Alternatively, we could toggle the size of a floating widget when switching tabs. But that has its own sub-set of issues/cases to discuss.

Yeah that would be confusing I think...

GerardasB:
> The fix is working well when you explicitly drag out the non resizable tab, but if it remains alone in its container, it will not "switch" back to his non resizable size.

Just resetting the size will have similar issues to the previously proposed "toggling the size of a floating widget when switching tabs", i.e. unexpectedly changing the size and position (in case the widget is contained on the screen).

We should probably think more about the use case with multiple widgets. I.e. we could:
1. Disallow merging if a widget is not resizable. Might be considered as a side effect...
2. Disallow merging if a widget is not resizable AND the target floating widget size doesn't match. We'd still want to allow programmatically changing the size of a floating widget (which could be similar to user-resizing).
3. Provide a way to define allowed floating widgets (similarly to allowed stage panels).

Juliakas:
> We should probably think more about the use case with multiple widgets. I.e. we could:
> 
> 1. Disallow merging if a widget is not resizable. Might be considered as a side effect...
> 2. Disallow merging if a widget is not resizable AND the target floating widget size doesn't match. We'd still want to allow programmatically changing the size of a floating widget (which could be similar to user-resizing).
> 3. Provide a way to define allowed floating widgets (similarly to allowed stage panels).

In our case, we have some widgets that are floating by default and would benefit from being set as non resizable. I think we could have option 2 and developers could make non resizable floating widget same size groups if they so choose, but option 1 is good as well.

Another thing - our use case also requires that non resizable floating widgets should be able to be docked regardless of size.

GerardasB:
Merging in as is now, opened https://github.com/iTwin/appui/issues/502 to track edge cases of restriction quirks.

raplemie (approved):

PR #501, opened 2023-09-26
Fix Message Center messages icon colors

### LINKED ISSUE #488: MessageCenterField message icon has no colour based on severity
**Describe the bug**
Messages in MessageCenter have no colouring on their icons.

As far as I know, message icons were coloured based on severity back in Appui 3.x. 

**Expected behavior**
Icons have colours based on severity.

**Screenshots**
Current: 
image

Expected: 
image

raplemie:
## Changes

Brought back some classnames changes that were lost when we removed the webfont.

## Testing

Added new tests to complete the Message Center Field to include messages so we can validate that we do have the proper colors.

Moved the Message Center component in component examples to the bottom of the list so all the content can be seen on the snapshot, which caused a need to update the toolassistance field snapshot.

## Issue

resolves #488


GerardasB (approved):
Nice to have those visual tests :)

Issue #91, opened 2022-10-13
Give application control over UiItems display (UiProviders refactor discussions)


raplemie:
Current UiItemsManager functionality is missing some valuable control for applications that might want to control the final resolution of providers toolbar items, and widgets.
"overrides" were provided as an initial solution for this issue, but the solution is only filtering and need to be revisited along the effort to see what we can do.

## Requirements:
- [ ] Be able to modify priority of items. (IE: Always have a specific order, or always have civil tools[^1] at the bottom of the list)
- [ ] Be able to control display (on/off) of elements in specific frontstages. (IE: Dont display civil tools in Markup frontstage)
- [ ] Be able to position items in specific locations (IE: Move widget default location from right panel to bottom panel)
- More to come...

## Issues/considerations:
- `UiItemsManager` is static and can be called from anywhere, thus any solution based on the register function like the current "override" will not allow an application to control this provider.
- `UiItemsManager` currently call each providers asking for specific locations, which are then filtered by the provider implementation, thus, moving things around cannot be handled by simply changing a property.
- More to come...

Initiated internally by DR.

[^1]: Civil tools are used as an example, the desired solution is not to "simply ask civil to change their provider" ;)

PR #504, opened 2023-09-28
Update `UiItemsProvider` to facilitate overriding

GerardasB:
## Changes

Add alternative provider functions to `UiItemsProvider` to facilitate overriding scenarios.
Currently existing functions like `provideToolbarItems` return items based on stage and item location arguments which cause complications when overriding i.e. provider `A` returns `ItemA` for `StageA` only. To override provider `A`, the application needs to know stage id `StageA` (implementation detail) to get the item and map it to a different frontstage.
New provider functions are argument-free to avoid this. 
Additionally, the proposal is to group providers around frontstages: `stageIds` and `stageUsages` properties are added to `UiItemsProvider`.
`Widget` and `ToolbarItem` types are appended with a `location` property that allows to specify location in the layout(s). It replaces the previous `provideToolbarItem` arguments `StagePanelLocation` and `StagePanelSection`. This has additional benefits in extensibility, i.e. with current location/section combinations applications are limited to 4 toolbars (with semantically pre-defined locations).

## Testing

Added additional e2e, unit tests and ran a standalone test app.


ui/appui-react/src/appui-react/ui-items-provider/AbstractUiItemsManager.ts
```diff
@@ -247,188 +243,175 @@ class AbstractUiItemsManagerAdapter implements Target {
...
-
-  public provideToolbarButtonItems(
-    stageId: string,
-    stageUsage: string,
-    usage: AbstractToolbarUsage,
-    orientation: AbstractToolbarOrientation,
-    _stageAppData?: any
-  ) {
-    if (!this._adaptee.provideToolbarItems) return [];
+function createUiItemsProviderToAbstractAdapter(
```
GerardasB, line 246: Previous adapter used to lose additional object properties.

ui/appui-react/src/appui-react/ui-items-provider/UiItemsManager.ts
```diff
@@ -211,42 +214,57 @@ export class UiItemsManager {
...
+        const providerItems = uiProvider.provideToolbarItems
+          ? uiProvider
+              .provideToolbarItems(stageId, stageUsage, usage, orientation)
+              .map((item) => ({ ...item, providerId }))
+          : [];
+        items.push(...providerItems);
+      });
+    }
+
+    // Note: items returned by `getToolbarItems` are not returned when using UiItemsManager from @itwin/appui-abstract@^3.0.0
```
GerardasB, line 241: Should we support new providers from appui-abstract 4.0.0? (when running w/o adapters). With the current implementation calling `AbstractManager.getWidgets()` and `Manager.getWidgets()` will return different results.

ui/appui-react/src/appui-react/widgets/Widget.tsx
```diff
@@ -52,4 +52,8 @@ export interface Widget {
   readonly label?: string | ConditionalStringValue;
   readonly priority?: number;
   readonly tooltip?: string | ConditionalStringValue;
+  /** Describes to which container the item is added.
+   * @alpha TODO: rename due to CanFloatWidgetOptions["containerId"]
```
GerardasB, line 56: Suggestions welcome.

ui/appui-react/src/appui-react/widgets/WidgetUtilities.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import type { StagePanelLocation } from "../stagepanels/StagePanelLocation";
+import type { StagePanelSection } from "../stagepanels/StagePanelSection";
+
+/** Helper class to create widget items.
+ * @alpha
+ */
+export namespace WidgetUtilities {
+  /** @alpha */
+  export function toContainerId( // TODO: rename?
```
GerardasB, line 17: Same reasoning as `ToolbarItemUtilities` OR might need a rename either way due to `containerId` rename.

ui/appui-react/src/appui-react/widgets/WidgetUtilities.tsx
```diff
@@ -0,0 +1,38 @@
...
+    location: StagePanelLocation,
+    section: StagePanelSection
+  ) {
+    return JSON.stringify({ location, section });
+  }
+
+  /** @alpha */
```
GerardasB, line 24: Missing docs.

ui/appui-react/src/test/ui-items-provider/UiItemsManager.test.tsx
```diff
@@ -770,6 +773,345 @@ describe("UiItemsManager", () => {
...
+      });
+
+      it.skip("should provide only for specified stageUsages", () => {
+        UiItemsManager.register({
+          ...provider,
+          stageUsages: ["custom"],
+        });
+
+        // TODO:
+        expect(UiItemsManager.getBackstageItems()).lengthOf(0);
```
GerardasB, line 1000: Should we add optional `stageId` and `stageUsage` arguments to `UiItemsManager.getBackstageItems`? Or should we continue to treat backstage items as global (which IMHO complicates API unnecessarily).

ui/appui-react/src/appui-react/toolbar/ToolbarItem.ts
```diff
@@ -56,6 +56,10 @@ export interface CommonToolbarItem {
   readonly groupPriority?: number;
   /** Priority within a toolbar or group. */
   readonly itemPriority: number;
+  /** Describes to which toolbar the item is added.
```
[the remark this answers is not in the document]
jason-crow: is toolbarId just a way for the the item to constrain itself to a particular toolbar?

ui/appui-react/src/appui-react/widgets/Widget.tsx
jason-crow: what exactly is a container in this sense?

ui/appui-react/src/appui-react/widgets/WidgetUtilities.tsx
```diff
@@ -0,0 +1,38 @@
...
+  export function fromContainerId(toolbarId: string):
```
jason-crow, line 25: toolbarId copy paste? shouldn't be containerId?

ui/appui-react/src/appui-react/toolbar/ToolbarItem.ts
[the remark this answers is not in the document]
GerardasB: Yes, previously it was done through `ToolbarUsage` and `ToolbarOrientation` args, but this allows us to cover more general cases (i.e. additional toolbars or different layouts/meanings).

ui/appui-react/src/appui-react/widgets/Widget.tsx
GerardasB: Similar to `toolbarId` it would map to a section of a stage panel (replaces `StagePanelLocation` and `StagePanelSection` args)

jason-crow: I think its dangerous to expect this field which is typed string to be of the form `JSON.stringify({ location, section })`

I would think it'd be easier to manage by just having a `container` field that maps to `{ location: SPL; section: SPS; }`

A part of me thinks adding a layer with the widget and this would help for future support of layouts that have different location/section concepts (you might not want to have those on the actual widget) but there's already allowedPanels so probably not a big deal

GerardasB: It would be easier to manage, but not as extensible. I.e. suppose we'd want to provide a separate layout that supports previously removed "topMost" location or "middle" section. On the other hand it would probably make sense to have this strongly typed instead (and i.e. allow any object for customization?).

raplemie: If we do that, then all providers needs to be updated to support all layouts, and we cannot allow new layouts to be defined as we still stick to a specific sematic. I dont think that `JSON.strignify` is a solution however, but we could provide wellknowns enums to be used, on top of any arbitrary strings.

``` ts
enum AppUIWidgetTargets {
  TopStart: "appuidefaultlayout-top-start",
  TopEnd: "appuidefaultlayout-top-end",
  ...
}
```

Then we can use these in the `containerId` the well known ones, and if you want something else, you are free to do so.

``` ts
containerId: AppUIWidgetTargets.TopEnd;
```

Then the layout can ask for widgets that are directed to itself.

test-apps/appui-test-app/appui-test-providers/src/ui/providers/UpdatedUiItemsProvider.tsx
```diff
@@ -0,0 +1,77 @@
...
+        () => undefined,
+        "Updated backstage item"
+      ),
+    ];
+  },
+  getStatusBarItems: () => {
+    return [
+      StatusBarItemUtilities.createActionItem(
+        `${id}:statusbar-item`,
+        StatusBarSection.Center, // TODO: map by sectionId?
```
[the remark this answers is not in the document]
raplemie: I think what we are currently doing is rather change toolbar item and widgets to behave like statusbar Items, where the item knows where it should go, but that can be changed.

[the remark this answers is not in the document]
GerardasB: Yes, this was more applicable for custom "layouts" similar to `toolbarId` and `containerId`. Removing the TODO.

ui/appui-react/src/appui-react/widgets/Widget.tsx
GerardasB: I tend to agree with @jason-crow on the union types, however, since we want to support "custom" layouts we might not want to add SpatialWidget and PanelsWidget on the same type and instead consider a separate `SpatialUiItemsProvider` type... As discussed, it would make sense to extract layout-agnostic Widget properties. The biggest challenge is backward compatibility.

ui/appui-react/src/appui-react/ui-items-provider/UiItemsProvider.ts
```diff
@@ -24,21 +24,59 @@ export interface UiItemsProvider {
   /** Id of provider. */
   readonly id: string;
 
-  /** Provides toolbar items. */
+  /** Describes stages for which the items should be provided.
+   * @note Ignored for `provideToolbarItems`, `provideStatusBarItems`, `provideBackstageItems`, `provideWidgets` methods.
+   * @alpha
+   */
+  readonly stageIds?: ReadonlyArray<string>;
```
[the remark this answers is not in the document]
GerardasB: These are "ands" between the arrays, i.e. if `Provider1` is allowed in `Frontstage1` and `Usage1` it will not load items to `Frontstage2` `Usage1` or `Frontstage1` `Usage2`.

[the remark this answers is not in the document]
GerardasB: With this change, we aim to group the providers not only features but also frontstages.
The problem with a separate array is that it:
- Requires twice as many APIs and overrides.
- Enables "global" providers which provide items to multiple stages. This complicates overriding, i.e. need to know item IDs to map correctly or other implementation detail...

common/api/appui-react.api.md
```diff
@@ -3328,6 +3334,24 @@ export interface PanelStateChangedEventArgs {
...
+};
+
+// @alpha (undocumented)
+export interface PanelsUiItemsProvider extends UiItemsProvider {
+    readonly getToolbarItems?: () => ReadonlyArray<PanelsToolbarItem>;
+    readonly getWidgets?: () => ReadonlyArray<PanelsWidget>;
+}
+
+// @alpha (undocumented)
+export interface PanelsWidget extends Widget {
```
GerardasB, line 3350: @jason-crow, @raplemie does this make more sense? How about naming?

ui/appui-react/src/appui-react/toolbar/ToolbarItem.ts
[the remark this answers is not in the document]
GerardasB: This is changed as well to match the `containerId` changes of a Widget.

ui/appui-react/src/appui-react/widgets/Widget.tsx
GerardasB: Introduced `PanelWidget` with specific properties to replace the `containerId`.

ui/appui-react/src/test/ui-items-provider/UiItemsManager.test.tsx
raplemie: If that changes anything, I guess we can add them, since they can be empty anyway.

ui/appui-react/src/appui-react/ui-items-provider/UiItemsProvider.ts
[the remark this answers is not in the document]
raplemie: How is overriding working with this API ? If I create a measure tool provider, I will probably not put the stage id and stage usage info, how will an application decide where to put things ? #91

common/api/appui-react.api.md
raplemie: I think its a good direction, that will certainly allow us to have different layout in the future, and have more extensible options to them, I'm only wondering if we won't risk have clashes in interpreting these without a clear discriminator, or how would someone provide information to multiple layouts simultaneously if ever 2 have clashing names with different semantics.
Could something like this be a good idea ?
```ts
interface Widget {
  ...(current existing widget props),
  layoutConfig: {[layoutId: string]: {} },
}

const myWidget: Widget = {
  id: "my-widget",
  ...props,
  layoutConfig: {
    panels: {
      location: StagePanelLocation.Top,
      section:...,
    },
    spatial: {
      location: SpatialLocation.DocumentPanel,
    }
  }
}
```

Or we dont expect the provider creator to provide default values, so there wouldnt be any clashes ?

ui/appui-react/src/test/ui-items-provider/UiItemsManager.test.tsx
GerardasB: Added.

ui/appui-react/src/appui-react/ui-items-provider/UiItemsProvider.ts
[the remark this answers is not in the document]
GerardasB: The application would override the `stageIds` property of a provider when registering.

ui/appui-react/src/appui-react/ui-items-provider/UiItemsManager.ts
```diff
@@ -126,6 +127,7 @@ export class UiItemsManager {
   /**
    * Registers a UiItemsProvider with the UiItemsManager.
    * @param uiProvider the UI items provider to register.
+   * @param overrides the UI items provider to register.
    */
   public static register(
```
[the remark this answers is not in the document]
GerardasB: One idea to avoid this issue is to avoid global/static register functions in packages/modules and instead export a provider itself (which can later be overridden by an application when registering).

common/api/appui-react.api.md
GerardasB: Applied this proposal to both `Widget` and `ToolbarItem` types. How do we want to name the `panels` layout?

GerardasB: Updated `UiItemsManager` to return widgets/toolbar items w/o a defined location (for consistency with other item types).

GerardasB:
Update NextVersion.md

common/changes/@itwin/appui-react/items-provider_2023-10-06-10-54.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "Add alternative provider methods to UiItemsProvider.",
```
GerardasB, line 5: Update `NextVersion.md`

a-gagnon:
Looks good. I think the change from `UiItemsProvider.provideXX` methods in favor of parameterless `UiItemsProvider.getXX` methods.
Can't say I fully understand the benefits of putting `stageIds`/`stageUsages` directly on the `UiItemsProvider`, or even `Widgets.layouts`. It'll probably be clearer when it gets used...

My understanding is that stageId and stageUsage is determined by the app, and uiitemsprovider packages shouldn't be aware of that.

jason-crow (approved):

common/api/appui-react.api.md
```diff
@@ -5265,6 +5297,8 @@ export interface Widget {
     readonly id: string;
     // (undocumented)
     readonly label?: string | ConditionalStringValue;
+    // @alpha
+    readonly layouts?: WidgetLayouts;
```
raplemie, line 5301: Looking at writing the Content documentation and I wonder if we shouldnt name this differently, we already have Content Layouts (and StandardContentLayouts in appui-abstract) Adding another `layout` name in the mix will add to the already confusing naming we have. 
Sadly, I dont have a great idea to replace this name..
`uiOrganisation`
`uiConfig`
`destinations`
`where`
`home`
`slot`
`at`
`in`
`showIn`

ui/appui-react/src/appui-react/ui-items-provider/UiItemsManager.ts
raplemie: Is there even a way to run without adapters? UiItemsManager exist in iTwin.js 4.0, so we are always using adapters. (Most of the code is probably never reached as _abstractAdapter always exist, until we reach iTwin.js 5.0)

Even so, I'm not sure I understand the interrogation, at this point, I feel we are only changing the input, but give the same output from the Manager, so we probably have to tweak it so it returns the same thing regardless of was it requested by `get` or `provide` methods, it should still return the same, so the handling will be in the manager itself (So we should have some map that convert the "layouts" properties to the same "widget" we used to have, for example.)

The `layouts` property appearing in the next version would require adding an alpha method to the Manager itself, so "any" layout can request the widgets, and get all of them, and decide where to put them, but I understood that we werent doing that yet. (as such, when I provide a widget via `get`, the result should be the same as if I had provided it with `provide`)

I finally think that if `get` methods exists, we should only call those, and not the `provide` ones, as we expect whoever to have opted in the `get` methods to want to test them, and the way it is implemented right now, we will get both, and drop one of them (so, it will be difficult to test if it actually works)

GerardasB: That's correct - we only use the adapters at the moment.
These are valid points, I'll work on it.

GerardasB: Applied the following changes which simplified code quite a bit:
- Use the adapter to map items returned by `get` properties to `provide` properties
- Prefer `get` properties instead of `provide` in `UiItemsProvider`.
- Removed arguments from `getBackstageItems`.
- Removed `stageIds` and `stageUsages`. These were optional either way, we can add them in the future if needed.
- Enhanced the tests to actually test both variants: w/ and w/o abstract adapters

GerardasB: Tests added.

common/changes/@itwin/appui-react/items-provider_2023-10-06-10-54.json
GerardasB: Updated.

common/api/appui-react.api.md
GerardasB: I don't really like the proposed alternative names or could come up with a better alternative. My vote is to keep the `layouts` property name as is, since the `ContentLayout` type seems unique enough.

raplemie: Fine, we can always revisit if needed and feedback reveals any confusion.

raplemie (commented):
I like this improvement and simplification.

common/api/appui-react.api.md
```diff
@@ -3315,6 +3317,18 @@ export interface PanelSizeChangedEventArgs {
     size: number | undefined;
 }
 
+// @alpha
+export interface PanelsLayoutToolbarItem {
```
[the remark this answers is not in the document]
GerardasB: `section` is already a required property of status ba

ui/appui-react/src/appui-react/widgets/Widget.tsx
```diff
@@ -47,4 +48,28 @@ export interface Widget {
...
+}
+
+/** Describes widget configuration specific for each layout.
+ * @alpha
+ */
+export interface WidgetLayouts {
+  /** Widget configuration in a panels layout. */
+  readonly standard?: PanelsLayoutWidget;
+  /** Widget configuration in a custom layout. */
+  readonly [layoutId: string]: Object | undefined;
```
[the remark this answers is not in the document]
GerardasB: I think we could remove the key-value pair for now. Making `standard` an optional property in the future would be a breaking change...

ui/appui-react/src/appui-react/ui-items-provider/UiItemsProvider.ts
```diff
@@ -24,21 +24,47 @@ export interface UiItemsProvider {
   /** Id of provider. */
   readonly id: string;
 
-  /** Provides toolbar items. */
+  /** Provides toolbar items.
```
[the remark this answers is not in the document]
GerardasB: Added a note.

ui/appui-react/src/appui-react/widgets/Widget.tsx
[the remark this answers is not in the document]
GerardasB: Alternatively, we could use a default location for widgets/toolbar items if `standard` is not provided. But then we will probably want to add a flag to the `frontstage` to disable that behavior in the future.

common/api/appui-react.api.md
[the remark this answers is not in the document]
GerardasB: Maybe it would make sense to add an empty object at least to other items: status bar item, backstage item. That would allow you to conditionally provide an item based on a layout (unless we decide to go with default locations for widgets, toolbar items).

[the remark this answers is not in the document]
raplemie: I was under the impression that the items returned by `get` methods were different than the one by `provide`, but yeah, if we intend to return the same ones, then we can't do that. And at the moment, I guess just like we removed the custom ones, we shouldnt add unused properties for now, so let's leave it as is, we can always change it later when we add Custom stuff management.

ui/appui-react/src/appui-react/widgets/Widget.tsx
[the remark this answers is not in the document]
GerardasB: Since it is a property like any other, making it optional is a breaking change.
I.e. in v1 (where property is required), a developer is accessing that w/o any checks. After we update our definitions it will either result in TS errors or runtime errors.
Also, this will not solve the original issue of forcing required location values, since we have to keep the `layouts` property  _optional_. TS will not complain about `standard` property unless `layouts` is set...

raplemie (approved):
I think that's a good starting point, we'll probably want to clarify that if `layouts.standard` is not provided, the only way to reach the related items will be by directly calling the registered providers `get*` methods (instead of using the Manager `get` methods.

PR #505, opened 2023-09-28
Widget No Longer Disappears When Popout Fails

joehenry9498:
## Changes
- Whenever widget is popped out and fails for some reason, widget will no longer disappear and toast will appear showing widget failed to pop out. This enables you to continue using the widget, previously you had to refresh to get the widget back. Changed the tests accordingly.

## Testing
- I was able to reliably to get to the popup to fail by doing the following. I ran the standalone on localhost and had an iframe point to that local host. Here is the configuration I used for the iframe.

- <iframe src= "http://localhost:3000/" width="1200px" height="500px" sandbox="allow-scripts allow-same-origin allow-modals"></iframe>

- Widget should not disappear and toast should appear saying the widget pop out failed. A widget pop out on the normal standalone project where popouts are allowed should work as usual. Also check widget popouts work as expected on Electron


ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
```diff
@@ -274,7 +280,10 @@ export class FrontstageDef {
     }
 
     for (const popoutId of popoutsToOpen) {
-      this.openPopoutWidgetContainer(popoutId);
+      if (oldState) {
+        const result = this.openPopoutWidgetContainer(popoutId, oldState);
+        return result;
```
GerardasB, line 285: Do we assume only a single popout can be opened at a time (return in a loop)?

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
```diff
@@ -806,7 +826,10 @@ export class FrontstageDef {
   /** Opens window for specified PopoutWidget container. Used to reopen popout when running in Electron.
    * @internal
    */
-  public openPopoutWidgetContainer(widgetContainerId: string) {
+  public openPopoutWidgetContainer(
+    widgetContainerId: string,
+    oldState: NineZoneState
+  ): boolean | void {
```
GerardasB, line 832: Just return a `boolean` att.

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
```diff
@@ -287,13 +296,24 @@ export class FrontstageDef {
...
     this._nineZoneState = state;
-    this.handlePopouts(oldState);
+    const popoutResult = this.handlePopouts(oldState);
     if (oldState) {
       this.triggerStateChangeEvents(oldState);
     }
 
     const isClosing = this._isStageClosing || this._isApplicationClosing;
     if (isClosing && !this.isReady) return;
+    if (popoutResult === false) {
```
GerardasB, line 306: We are not emitting an event, but we are not resetting the state either. `frontstageDef.nineZoneState` will still contain a popout widget.

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
```diff
@@ -287,13 +296,24 @@ export class FrontstageDef {
 
     const oldState = this._nineZoneState;
     this._nineZoneState = state;
-    this.handlePopouts(oldState);
+    const popoutResult = this.handlePopouts(oldState);
     if (oldState) {
```
GerardasB, line 300: I wonder if `onWidgetStateChangedEvent` is emitted when the popout widget fails to open.

GerardasB (changes requested):

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
joehenry9498: I was unsure about the mechanics of this. It seemed like it is setup to be able to open multiple at a time, but I wasn't sure if it was possible for the user to open multiple at a time. Because they have to click the button for the widget to popout, so it didn't seem like it to me. Can change if it is possible

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
joehenry9498: Looking into further

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
joehenry9498: Looking into further

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
```diff
@@ -274,7 +280,10 @@ export class FrontstageDef {
     }
 
     for (const popoutId of popoutsToOpen) {
-      this.openPopoutWidgetContainer(popoutId);
+      if (oldState) {
```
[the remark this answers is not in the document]
GerardasB: Haven't tried reproducing, but seems like that's the case while looking at the code.

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
GerardasB: Looks like we need to set up some testing for electron...
This should be possible with electron (once again on initialization, but in any handler as well).

GerardasB: I.e. by calling https://www.itwinjs.org/reference/appui-react/frontstage/frontstagedef/popoutwidget/

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
joehenry9498: The state is being reset in FronstageDef.tsx line 873. The state appears fine to me when debugging whenever a popout fails. The widget is contained as floating and not as a popout

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
joehenry9498: It is emitted

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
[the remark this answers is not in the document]
joehenry9498: Added a fix to push the old state requirement deeper into the code. An electron popout should still popout on original state. I just worry about the edge case with it where the popout fails on the initial state on Electron, that would lead to the widget disappearing again

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
GerardasB: With what `WidgetState`? Will we emit a correct `WidgetState` once again when restoring to the "old nine zone" state?

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
GerardasB: Yeah, I think we could get away with a couple of unit tests:
- Simulate popout button click (i.e. just dispatch an action)
- Simulate "electron" initialization w/ multiple popouts (not sure if there are cases where the window would fail to open on the electron)

Assert on opened windows, widget state changes, and frontstageDef.nineZoneState.

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
joehenry9498: It is emitting onWidgetStateChangedEvent whenever it is setting the failed popout back to the old state. It is not emitting the onWIdgetStateChangedEvent with the incorrect state

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
joehenry9498: We already have a simulated popout button dispatch test. The test is located in `FrontstageDef.test.tsx`. I'll create a new one for initialization for multiple popouts

joehenry9498: New test has been created to cover the second scenario

GerardasB (approved):

PR #507, opened 2023-10-02
Revert iTwinUI dependency upgrade.

raplemie:
## Changes

Reverted packages changes from #493 because it causes issues with the new stylesSheets in some of our tests, iTwinUI V3 uses a different system so it might be only temporary, not sure it's worth adding the error overrides suggested for this to work. JSDOM parser have issues with advanced CSS features and output the css in the console, making tests really difficult to read, it is possible to override the console to drop these messages, but when I tried it causes other problems.

## Testing

Removed ChildMessageManager tests that were trying to test style copy, because `document.styleSheets` and `document.adoptedStylesheets` are always empty (Because of JSDOM ?)

Added visual tests for "View Attributes" popout tests, which have iTwinUI styling. Should be more reliable that unit tests.

Ran tests, validated that the warnings are gone.


raplemie:
@joehenry9498 FYI.

docs/changehistory/NextVersion.md
```diff
@@ -18,6 +18,7 @@ Table of contents:
 - Fix message center messages overflow behavior breaking layout.
 - Fix message center messages icon colors based on message severity.
 - Correctly configure `resizable` flag of a floating widget after docking and undocking.
+- Fix missing styles in child window by also copying `adoptedStylesheets`.
```
GerardasB, line 21: The actual change is in #493

GerardasB (approved):

PR #508, opened 2023-10-02
Fix grips flickers + AUDIT FIX

### LINKED ISSUE #443: UI: Unpinned panel handle starts to flicker on hover and fails to open the panel
#### Environment

- __Package version(s)__: 4.3.0
- __Operating System__: Windows 11
- __Browser name and version__: Google Chrome 115.0.5790.110 and Firefox 115.0.2

#### Steps to reproduce

1. Unpin any of the panels
1. Make sure that it's closed and you see the panel handle
1. Move your mouse to the end of the window, near the panel handle

#### Actual behavior

Notice that unpinned panel handle starts to flicker on mouse hover and fails to open the panel

#### Expected behavior

In the case of unpinned and collapsed panel, it should just always open on hover

#### Additional information

Quick demo of the issue from Pineapple side:

93d16df8-65ec-4cbf-afbb-05288ae8dc92

raplemie:
## Changes

Add a delay to the grip movement so the grips stop flickering.
Update chai.js to fix npm audit issue.

## Testing

Ran unit test, tested solution in DR manually (could not reproduce in test apps)

## Issue
resolves #443

### DIFF
--- a/ui/appui-layout-react/src/appui-layout-react/widget-panels/Grip.scss
+++ b/ui/appui-layout-react/src/appui-layout-react/widget-panels/Grip.scss
@@ -16,13 +16,23 @@ $line-grip-width: calc(#{$nz-grip-width} - 2px);
   justify-content: center;
   position: relative;
   display: flex;
-  transition: background-color 0.2s ease-out;
+  transition: 0.2s ease-out;
 
   > .nz-handle {
     position: absolute;
     top: 50%;
     left: 50%;
     transform: translate(-50%, -50%);
+  }
+
+  &.nz-left {
+    transition-property: background-color, right;
+  }
+  &.nz-right {
+    transition-property: background-color, left;
+  }
+  &.nz-top {
+    transition-property: background-color, bottom;
   }
 
   &.nz-left,
--- a/common/changes/@itwin/appui-layout-react/raplemie-gripTransitionIssue_2023-10-02-19-37.json
+++ b/common/changes/@itwin/appui-layout-react/raplemie-gripTransitionIssue_2023-10-02-19-37.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-layout-react",
+      "comment": "Fix auto collapsing panels grips flickering in some cases.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/appui-layout-react"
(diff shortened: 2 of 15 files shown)


ui/appui-layout-react/src/appui-layout-react/widget-panels/Grip.scss
```diff
@@ -25,6 +25,16 @@ $line-grip-width: calc(#{$nz-grip-width} - 2px);
...
 
+  &.nz-left {
+    transition-property: background-color, right;
+  }
+  &.nz-right {
+    transition-property: background-color, left;
+  }
+  &.nz-top {
+    transition-property: background-color, bottom;
+  }
```
GerardasB, line 36: How about `nz-bottom`?

GerardasB: This is also not adding a `delay`, but a `duration` to "position properties"? I don't see where we would change one of `right/left/bottom` properties though for the transition to kick in?

raplemie: 1. `nz-bottom` is not changing its `top` like this on `nz-collapsed`, and anyway this issue really only occured on the right panel, but I'm adding safeguard in case it would ever happen in the other ones.
2. I'm not adding a `delay` but simply a duration to these properties because that is enough to fix the issue (adding a delay would get weird as the blue bar wouldnt appear and the panel would move around.
3. This property is actually changed in https://github.com/iTwin/appui/blob/2af8649373b55aedc4cd05d59f507243ecdf9eb2/ui/appui-layout-react/src/appui-layout-react/widget-panels/Panel.scss#L107 (and other lines for other sides), but the `transition` needs to be there before the `nz-collapsed` is added, so I cant put it in the same place for this to work correctly.

GerardasB: Well this new CSS should be added to `Panel.scss` OR the `right: 0;` (and related CSS) should be moved to `Grip.scss`.

GerardasB: Approving for now just to get the fix out, but would be great to add a todo at least..

GerardasB (approved):

PR #509, opened 2023-10-04
ControlledTree: Fix tree rendering with stale data

saskliutas:
## Changes

Fixed `useTreeModel` hook to make one more render when `TreeModelSource` changes to make sure that tree is rendered with latest data in case tree model was modified between `TreeModelSource` change and `onModelChanged` event listener was added.

Replaced `useRerender` with `useState` because `rerender` function had new reference on each render and caused `onModelChanged` event listener to be removed and added on each render.

## Testing

Added unit test.


saskliutas:
@Mergifyio backport release/4.5.x

ui/components-react/src/test/tree/controlled/TreeHooks.test.tsx
```diff
@@ -18,74 +18,125 @@ import {
...
-  beforeEach(() => {
-    modelSourceMock.reset();
-    onModelChangeEvent = new BeUiEvent<[TreeModel, TreeModelChanges]>();
-
-    modelSourceMock
-      .setup((x) => x.onModelChanged)
-      .returns(() => onModelChangeEvent);
-    modelSourceMock.setup((x) => x.getModel()).returns(() => testModel);
-  });
+  beforeEach(() => {});
```
grigasp, line 35: seems unnecessary

ui/components-react/src/test/tree/controlled/TreeHooks.test.tsx
```diff
@@ -18,74 +18,125 @@ import {
...
+
+    before(async () => {
+      await TestUtils.initializeUiComponents();
+    });
+
+    after(() => {
+      TestUtils.terminateUiComponents();
+    });
+
+    it("until subscription to `onModelChangedEvent`", async () => {
```
grigasp, line 129: could you clarify the test name?

grigasp (commented):
> Replaced `useRerender` with `useState` because `rerender` function had new reference on each render and caused `onModelChanged` event listener to be removed and added on each render.

Isn't that a bug in `useRerender`?

ui/components-react/src/test/tree/controlled/TreeHooks.test.tsx
saskliutas: Updated test name

saskliutas:
> > Replaced `useRerender` with `useState` because `rerender` function had new reference on each render and caused `onModelChanged` event listener to be removed and added on each render.
> 
> Isn't that a bug in `useRerender`?

Not necessarily. If `useRerender` was added to be used in other places than `useEffect` there might be no need to memoize `rerender` function. Although I did a quick search and it looks like `useRerender` is not used anywhere anymore. So I will remove it.

GerardasB (approved):

grigasp (approved):

PR #513, opened 2023-10-04
Backport 4.6.0 changelogs

raplemie:


.vscode/settings.json
raplemie, line 1: This is an oversight, I forgot to remove this change, but... Can we commit this ? I often use node_modules and these default exclusions are rather annoying, it would better be in the user settings, rather than being commited to the workspace settings like that. Any objections ?

joehenry9498: I like being able to see the node modules as well

GerardasB: I agree, however, we should probably keep excluding node_modules from the search.

GerardasB (approved):

PR #516, opened 2023-10-06
adjust to more appui-abstract utils changes

ben-polinsky:
Guiding principles for these changes:

Remove functionality which is not in appui's domain and which can better supported outside of iTwin and AppUI. (i.e. Keyboard Key enums, retrieving class names of objects or their prototype.)
 
Proposed `@itwin/appui-abstract` deprecations

[utils/misc.ts](https://github.com/iTwin/itwinjs-core/blob/master/ui/appui-abstract/src/appui-abstract/utils/misc.ts) - `getClassName`

This function returns the class name of an object or its constructor. Renamed to `getObjectClassName` to reduce confusion with HTML/React element class names and moved internal.  However, because AppUI was previously consuming this internal API, it will be deprecated in `@itwin/appui-abstract`. The suggestion to consumers will be migrate to a custom implementation.
 
[utils/IconSpecUtilities.ts](https://github.com/iTwin/itwinjs-core/blob/master/ui/appui-abstract/src/appui-abstract/utils/IconSpecUtilities.ts) - `IconSpecUtilities`
This utility seems unnecessary to me. We can determine the type of the icon string with a few checks and simplify usage. To ensure core 4.x compatibility with previously released appui versions, this will not be deprecated in core until version 5.x, but a note will be added.
 
[utils/UiError.ts](https://github.com/iTwin/itwinjs-core/blob/master/ui/appui-abstract/src/appui-abstract/utils/UiError.ts) - `UIError` - Will be deprecated in core. Delay switching to `BentleyError` next major version of appui comes out as it is a breaking change.
 
[utils/filter/*](https://github.com/iTwin/itwinjs-core/tree/master/ui/appui-abstract/src/appui-abstract/utils/filter) - `IMatch` and `matchesWords`
Both consumed by appui-* even though the latter is internal. They'll be deprecated in core, and have been moved to `appui-react`.  One usage of `IMatch` occurs in core-react which has been in-lined. 
	
[common/KeyboardKey.ts](https://github.com/iTwin/itwinjs-core/blob/master/ui/appui-abstract/src/appui-abstract/common/KeyboardKey.ts) - `KeyboardKey`
Usage of the two enums and one method in this module are sprinkled throughout appui and consuming applications. In my opinion, this isn't our responsibility. From what I can see we never use `SpecialKey` or `FunctionKey` as a type in any public APIs (even the HTML `keyboardEvent.key` property is typed as a string) - they're just for convenience. Furthermore, strings are used in place of enum values often. I'm deprecating this entire module and recommending using the `ts-key-enum` package. Though I've replaced usage throughout unit tests, I verified that the appui-abstract types work as well.

Proposed `@itwin/appui-abstract` @internal removals

[utils/misc.ts](https://github.com/iTwin/itwinjs-core/blob/master/ui/appui-abstract/src/appui-abstract/utils/misc.ts) - `LoggerCagtegory`
Removed and will not be replaced as a public API in AppUI repo. In-lined where necessary.

[utils/filter/*](https://github.com/iTwin/itwinjs-core/tree/master/ui/appui-abstract/src/appui-abstract/utils/filter) - `*`
Nearly all of these filter utilities and types are internal and not used in core, AppUI, or any of our popular applications. The initial implementation was taken from VS code. What remains is the `IMatch` interface (public) and the `matchesWords` method (internal). These are detailed above.

Companion PR in core is here: https://github.com/iTwin/itwinjs-core/pull/6179


ui/appui-react/src/appui-react/utils/UIError.ts
```diff
@@ -0,0 +1,15 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import type { GetMetaDataFunction } from "@itwin/core-bentley";
+import { BentleyError, BentleyStatus } from "@itwin/core-bentley";
+
+/** iTwin.js UI UiError class is a subclass of BentleyError, extended with a category. Errors are logged.
+ * @public
+ */
+export class UiError extends BentleyError {
```
GerardasB, line 11: How is the `category` used? Is the expectation for `category` to be used while logging the errors (i.e. logger checks if an error is `instanceof UiError`)? Having multiple classes isn't a good approach if that's the case. Maybe we should consider either keeping a single class OR the need for category/UiError altogether?

ui/appui-react/src/appui-react/framework/FrameworkKeyboardShortcuts.ts
```diff
@@ -5,7 +5,7 @@
 /** @packageDocumentation
  * @module KeyboardShortcut
  */
-import type { FunctionKey, SpecialKey } from "@itwin/appui-abstract";
+import type { FunctionKey, SpecialKey } from "../keyboardKey/KeyboardKey";
```
GerardasB, line 8: Could you run `extract-api`, this seems like a breaking change (or just invalid), since `SpecialKey` is used in a `@public` type `KeyboardShortcutProps`.

ui/appui-react/src/appui-react/utils/UIError.ts
ben-polinsky: Right. My initial thinking was to eliminate the `UiError` altogether, but thought "it had to be used for something." However, I've found no evidence of loggers or other error handlers checking if the error is `UiError` or if the error has a category property. Certainly not used in core, and I found no usage in our most popular apps. 

I can ask around to see if anyone remembers the intent, but that is often a fruitless task.

ui/appui-react/src/appui-react/framework/FrameworkKeyboardShortcuts.ts
GerardasB: I think tooling doesn't complain since the newly added `@internal SpecialKey` is not in the barrel file.

ui/appui-react/src/appui-react/utils/UIError.ts
ben-polinsky: Remembered history is that it was just used in appui for all error types, but once logging categories were introduced, the `UiError.category` was essentially abandoned. I'll look to remove it altogether if possible.

ben-polinsky: If we want to be safe, we cannot remove `UiError`. It's public and consumers could possible be using the category property even though I've found no evidence so far. I feel uncomfortable removing it from appui.

What do you think about deprecating it in `core/appui-abstract` (recommending `BentleyError` as a replacement), continuing to use it in appui until core@6 comes out, then remove it completely?

ui/appui-react/src/appui-react/framework/FrameworkKeyboardShortcuts.ts
ben-polinsky: Thinking through this more, I don't see an issue. `SpecialKey` is still available, just deprecated. It can continue to be used. The property in question `KeyboardShortcutProps.key` is of type `string | FunctionKey | SpecialKey` - with both `FunctionKey` and `SpecialKey` being `string enums`, the set of acceptable types is the entire set of strings. Even if we removed `FunctionKey` and `SpecialKey` from the union type, the set of acceptable types is the same.

ui/appui-react/src/appui-react/utils/UIError.ts
GerardasB: Yes, I think that makes more sense, instead of re-introducing it once again.

ui/appui-react/src/appui-react/framework/FrameworkKeyboardShortcuts.ts
GerardasB: Oh, good catch, `SpecialKey` enum is a subset of a string.

common/api/imodel-components-react.api.md
```diff
@@ -333,7 +333,7 @@ export interface CubeProps extends React_2.AllHTMLAttributes<HTMLDivElement>, Co
 }
 
 // @public
-export class CubeRotationChangeEvent extends UiEvent<CubeRotationChangeEventArgs> {
+export class CubeRotationChangeEvent extends BeUiEvent<CubeRotationChangeEventArgs> {
```
GerardasB, line 336: The same arguments apply to "removal of UiEvent" since `category` property is not on `BeUiEvent` and a different subclass is used which would be breaking in the case of `instanceof` checks.

ui/appui-react/src/appui-react/utils/IconSpecUtilities.ts
raplemie, line 1: It does not make sense to put that in AppUI-React package. The sole reason this exists is because AppUI-Abstract do not allow React icons, if that was supported (as it is in AppUI-React, people would simply import directly as React and not bother with this)

In essence, if you need this, it's because you dont want to use AppUI-React and want to use the Abstract version (think extensions)

ben-polinsky: Why do we need this utility at all? There are two cases it's currently used for:

- Using `data uris` - we can detect if it has a `data:` prefix
- Using another url to be fetched - we can detect if the url is valid and/or if the url ends in .svg

ui/appui-react/src/appui-react/utils/UIError.ts
ben-polinsky: Reverted. I believe we can move to `BentleyError` at the next major for `AppUi`, however, since we wouldn't be deprecating any public API first.

ui/appui-react/src/appui-react/utils/IconSpecUtilities.ts
ben-polinsky: I've removed the need for `IconSpecUtilities` - let me know your thoughts.

raplemie: A bit worried about "other" implementations, but I admit that this is very unlikely that anyone actually built an "IconSpecUtilities" parser. We will need to augment the defintion of the icon spec to validate what type of string is actually handled, currently, this is not clear at all. (Even the description of the `createWebComponentIcon` method does not help anyone to know what it should feed the method...) I'll make a separate comment for this in IconComponent file.

ui/core-react/src/core-react/icons/IconComponent.tsx
```diff
@@ -49,10 +55,17 @@ export function Icon(props: IconProps) {
...
 
   if (iconString) {
-    const webComponentString =
-      IconSpecUtilities.getWebComponentSource(iconString);
-    if (webComponentString) {
-      const svgLoader = `<svg-loader src="${webComponentString}"></svg-loader>`;
+    const webComponentString = getWebComponentSource(iconString);
+
+    if (
+      iconString.startsWith("data:") ||
```
raplemie, line 61: Since we are adding "allowed" values to `iconSpec`, can you add in the `IconSpec` type above the description of what is supported for the strings ?
IE:
```ts
/** Prototype for an IconSpec which can be a string, ReactNode or ConditionalStringValue.
 * Strings are expected to be the following:
 *  - Created by AppUi Abstract `IconSpecUtilities.createWebComponentIconSpec` (deprecated)
 *  - A URL to a SVG file (must end with ".svg").
 *  - A dataURI with MimeTypes `image/svg+xml` (ensure valid dataURI encoding) or `image/svg+xml;base64`;
 *  - Anything else will be treated as a class name and be directly added to the `class` attribute.
 * @public
 */
export type IconSpec =
  | string
  | ConditionalStringValue
  | React.ReactNode
  | ConditionalIconItem;
```

ui/appui-react/src/appui-react/keyboardKey/KeyboardKey.ts
```diff
@@ -0,0 +1,50 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+/** Enumeration for Function Keys
+ * @internal
+ */
+export enum FunctionKey {
```
raplemie, line 9: Instead of redefining those in each package, use the one in core-react (leave it internal, but do not duplicate the defintion)

ui/appui-react/src/appui-react/UiFramework.ts
```diff
@@ -428,12 +428,16 @@ export class UiFramework {
...
   /** @internal */
   public static loggerCategory(obj: any): string {
-    const className = getClassName(obj);
+    const className = UiFramework.getObjectClassName(obj);
     const category =
       UiFramework.packageName + (className ? `.${className}` : "");
     return category;
   }
 
+  private static getObjectClassName(obj: any): string {
```
raplemie, line 437: Use this function from `core-react` (keep it internal), do not duplicate it everywhere.

ui/appui-react/src/appui-react/content/ContentLayout.tsx
```diff
@@ -15,14 +15,14 @@ import { SplitPane } from "./split-pane/SplitPane";
...
+import {
+  type ContentLayoutProps,
+  type LayoutFragmentProps,
+  type LayoutHorizontalSplitProps,
+  type LayoutSplitPropsBase,
+  type LayoutVerticalSplitProps,
+  UiEvent,
 } from "@itwin/appui-abstract";
-import { UiEvent } from "@itwin/appui-abstract";
 import type { ActiveContentChangedEventArgs } from "../framework/FrameworkContent";
```
raplemie, line 26: Although I find import reordering annyoing to review, I understand it, but here we are changing the way we import stuff, but only for one package, is there a reason for that ? Should prettier catch that ? I feel there is a lot of reorder in this PR in files that are not otherwise modified, that makes it a bit lenghty to review.

ben-polinsky: I agree. My prettier settings were automatically "correcting" to mixing type and value imports, and then I was undisciplined and sometimes followed the new syntax. I'll revert changes to files not otherwise modified which I most likely saved while reading.

ui/appui-react/src/appui-react/UiFramework.ts
ben-polinsky: @raplemie @GerardasB Did that by importing the function directly - however having a tough time getting this to work with both mocha and storybook. Was the idea to keep internal but still export via barrel?

ui/core-react/src/core-react/icons/IconComponent.tsx
ben-polinsky: This is done despite GH not showing the update here.

ui/appui-react/src/appui-react/content/ContentLayout.tsx
ben-polinsky: Done

ui/appui-react/src/appui-react/UiFramework.ts
raplemie: Yes, export it via barrel so we dont mix esm and cjs since this is not only used in tests.

raplemie:
> [common/KeyboardKey.ts](https://github.com/iTwin/itwinjs-core/blob/master/ui/appui-abstract/src/appui-abstract/common/KeyboardKey.ts) - `KeyboardKey` Usage of the two enums and one method in this module are sprinkled throughout appui and consuming applications. In my opinion, this isn't our responsibility. From what I can see we never use `SpecialKey` or `FunctionKey` as a type in any public APIs (even the HTML `keyboardEvent.key` property is typed as a string) - they're just for convenience. Furthermore, strings are used in place of enum values often. I'm deprecating this entire module and recommending using a custom implementation or a more-specific library.  I've in-lined constants where string literals seemed awkward.

We used to have a dependency on `ts-key-enum` in `component-react` package, which I removed since `appui-abstract` was offering the enum, we should probably return to this package for simplicity and reducing our overall maintenance.

common/api/imodel-components-react.api.md
raplemie: Category is added to `UIError`, not `UiEvent`, maybe that's the confusion :)

ben-polinsky: Yes, nonetheless, the subclass instanceof check would break users. I did look into https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance which could be used to keep those checks in tact, but I think it's easier to wait for the next major release.

ui/appui-react/src/appui-react/UiFramework.ts
ben-polinsky: Done, thanks.

ben-polinsky:
> We used to have a dependency on `ts-key-enum` in `component-react` package, which I removed since `appui-abstract` was offering the enum, we should probably return to this package for simplicity and reducing our overall maintenance.

Yeah good to have a suggestion for users to switch to as well. Thanks

full-stack-tests/ui/tests/backstage/backstage.test.ts-snapshots/backstage-test-with-snapshotPath-1-chromium-win32.png
ben-polinsky, line 1: Previously the playwright test this snapshot relates to was attempting to output an icon via the already deprecated `svg:` prefix and`?sprite`. As you can see, it was not rendering the icon. My original changes added padding/margin to the top list item here (Simple custom frontstage) which caused the snapshot test to fail. To me, the extra padding made sense because the item attempted to load an icon. 

I think this solution is better: I've updated the `ExampleStagesBackstageProvider` file to correctly specify icons and committed the updated snapshot files.

ben-polinsky:
@raplemie let me know if you could take another pass when you have a chance

ui/imodel-components-react/src/test/viewport/ViewportComponentEvents.test.ts
```diff
@@ -16,7 +16,7 @@ import { ViewportComponentEvents } from "../../imodel-components-react/viewport/
 import { TestUtils } from "../TestUtils";
 
 describe("ViewportComponentEvents", () => {
-  const onSelectedViewportChanged = new BeUiEvent();
+  const onSelectedViewportChanged = new UiEvent();
```
raplemie, line 19: This one had not changed in 2 years and was already using the correct BeUiEvent ;)

raplemie (changes requested):
Imports and debug console, other are minor.

test-apps/appui-test-app/connected/src/frontend/index.tsx
```diff
@@ -78,6 +77,7 @@ import {
   ToolAdmin,
   ViewClipByPlaneTool,
 } from "@itwin/core-frontend";
+import { getObjectClassName } from "@itwin/core-react/lib/cjs/core-react/utils/getObjectClassName";
```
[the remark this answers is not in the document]
ben-polinsky: Ah sorry thought I caught all.

ui/imodel-components-react/src/test/viewport/ViewportComponentEvents.test.ts
ben-polinsky: Dah... thanks

ui/appui-react/src/appui-react/statusfields/SectionsField.tsx
```diff
@@ -17,12 +17,11 @@ import {
   ViewClipDecoration,
   ViewClipDecorationProvider,
 } from "@itwin/core-frontend";
-import type { CommonProps } from "@itwin/core-react";
+import { type CommonProps } from "@itwin/core-react";
```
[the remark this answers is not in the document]
ben-polinsky: My vscode auto format was doing it initially. I changed most back but didn't catch this one.

raplemie:
Looks good, please add an entry in the NextVersion about the iconSpec change.

ben-polinsky:
> Looks good, please add an entry in the NextVersion about the iconSpec change.

Ah yes, thanks - done.

raplemie (approved):
LGTM!
Thanks!

PR #520, opened 2023-10-11
Notifications - #471 and #519

### LINKED ISSUE #471: mention in docs that we cannot send notifications to users, until status bar is enabled.
Lack of information in docs, it would be helpful if you could mention that enable status bar to send user notifications.
* * *

_⚠ Do not edit this section. It is required for imodeljs.github.io ➟ GitHub issue linking_

*   LINK: Notifications-and-Messages---iTwin.js-link-to-docs
*   TAG: imodeljs-docs-question
*   SOURCE: https://www.itwinjs.org/learning/ui/appui-react/notifications/

### LINKED ISSUE #519: HTMLElement does not work with ReactNotifyMessageDetails
**Describe the bug**
`ReactNotifyMessageDetails` accepts `MessageType = string | HTMLElement | ReactMessage` as valid types for its `detailedMessage` prop. However, an error occurs when attempting to output a HTMLElement as a message using `ReactNotifyMessageDetails` with `MessageManager.outputMessage`.

**To Reproduce**

This sandbox reproduces the behavior:
[ReactNotifyMessageDetails-does-not-support-html](https://www.itwinjs.org/sandboxes/AlfonsoMartello/ReactNotifyMessageDetails-does-not-support-html).
Please go to this sandbox and follow the instructions in the widget.

**Current behavior**
Error is thrown:
`Objects are not valid as a React child (found: https://imjsdocumentprodeussa01.z13.web.core.windows.net/sandbox/compiler/index.html?lng=en&itwinid=402f1a92-c7b1-4012-b787-7fa45e2e7fe4&imodelid=b55dec38-d9b7-4029-9b9c-6b899151328f#). If you meant to render a collection of children, use an array instead.`

**Expected behavior**
HTMLElement should be displayed in toast.

raplemie:
### Toast and Sticky
Moved `Toast` and `Sticky` type message handling directly into `MessageManager` instead of it being in `StatusBar`, as it didnt really make any sense since this component is not optional and part of the implementation.

### Activity Messages
Moved use of `ActivityMessageRenderer` in `ConfigurableUiContent` and is now always active, instead of being used in `StatusBar`. No other implementation details were changed, so the `ActivityCenterField` can be used as before, if present.

### HTMLElement in messages
Brought back `MessageRenderer` that was lost when we moved to new iTwinUI toast messages. (This is what handles `HTMLElement` and put it in react tree)

### MessageCenter
removed a `<br/>` and used `MessageDiv` instead of `MessageSpan` to get the same behavior but no wasted space as is done everywhere else. (We should refactor this part, but I'll do that in a different PR)

### AnimateOutToElement *update*
setAnimateOutToElement now remembers previous entries and will fallback to previous elements to animate to, when the MessageCenter was displayed multiple time, only the last one was kept, once it was removed, the animation was broken and went to the top left corner. (This was apparent in the "ComponentExample", but would have the same effect if it was displayed in a nested frontstage for example.)

## Testing

Added visual tests for HTML and React messages with overflow for MessageCenter AND sticky toasts.

## Issues
resolves #471
resolves #519


docs/changehistory/NextVersion.md
```diff
@@ -10,3 +10,5 @@ Table of contents:
 ### Fixes
 
 - Whenever widget is popped out and `window.open` fails, widget no longer disappears.
+- `AppNotificationManager` no longer requires `StatusBar` to be rendered in the active frontstage to show messages.
```
GerardasB, line 13: This would make more sense under "Changes". And let's tag this as a `minor`.

GerardasB (approved):

docs/changehistory/NextVersion.md
raplemie: Moved to changes

PR #543, opened 2023-10-19
Add frontstage stories

GerardasB:
## Changes

Add stories to showcase frontstages.

## Testing

N/A


docs/storybook/src/VirtualCursor.ts
```diff
@@ -0,0 +1,154 @@
...
+            return;
+          }
+
+          onMove(to);
+          currentPosition = to;
+          resolve();
+        };
+        requestAnimationFrame(move);
+      });
+    },
```
raplemie, line 149: Could we replace that by simply putting a `transition: transform 1s ease-in-out` ? (Or maybe have the 1s a parameter of the move)

GerardasB: I've tried that idea initially, however imagine dragging a

raplemie (approved):

PR #546, opened 2023-10-20
Fix custom hooks not working in React 18 strict mode

saskliutas:
## Changes

Fixed `useAsyncValue` hook not working in React 18 strict mode because `useEffect` was not symmetrical.
Fixed `useDisposable` hook not working in React 18 strict mode because `useEffect` was not symmetrical.
Cleaned up `useDebouncedAsyncValue` hook.
Bumped `react-window` and `react-highlight-words` versions to the ones that has React 18 as peerDependency.

## Testing

Manual testing in viewer app running with React 18 in strict mode.


raplemie (approved):
Not familiar with `rxjs`, but I'll take Grigas's review as rxjs part is valid!

PR #547, opened 2023-10-20
Add iTwinUI high contrast theme support

### LINKED ISSUE #332: Add Support to High Contrast provided by iTwinUI
For contract accessibility facilitation, we want to enable the existing High Contrast variant of iTwinUI from within AppUI.

(Possibly simply a prop to the `ThemeManager`)

Need to review what is required if more than that.

### LINKED ISSUE #407: ThemeManager bleeding out, if not used in root component
When ThemeManager is used not in the root component, it is bleeding out. 

For example, when ThemeManager is being used just on the route for viewing iModel and the theme is set to "dark":

**Expected behavior:** the theme is dark only in the iModel view, and not outside it in the application.
**Actual behavior:** the theme is partially or fully dark in the whole application.

- iTwin.js AppUI Version 4.2.0
- iTwin.js Version 4.0.2

### LINKED ISSUE #409: Change ThemeManagerProps "theme" property to string literals with predefined options
Currently, ThemeManagerProps "theme" property is set to type string:
image

It should be changed to string literals with predefined options instead to avoid possible errors (due to typos like "liht" instead of "light", etc.). For example, there is a type ThemeType, which includes all the possible options:
image

- iTwin.js AppUI Version 4.2.0
- iTwin.js Version 4.0.2

raplemie:
## Changes

* Added new themes values to support high contrast (resolves #332)
* Made the themes now a union of the enum values and (string & {}) to have IDE better support. (resolves #409)
* Updated documentation of `ThemeManager` to clarify scope (resolves #407)
* Removed `gradient` div in `SheetViewNavigationAid` which was `display:none` since 4 years, and the only place where we actually used `$buic-button-gradient-1` and `-2`, (even when removing display:none, the gradient was not actually working)
* The colors are no longer used, but I kept the vars in case someone actually use that outside of us, but we no longer make a difference between light and dark themes (If anyone actually notices that, we'll knwo that someone is using it, and we can tell them to stop using it as we will remove this sooner or later)
* Updated settings to allow selection of the new high contrast mode as well as explicitly requesting iTwinUI inheritance of the ThemeProvider (We no longer have any specific colors of our own, everything is derived from iTwinUI, making our theme manager somewat obsolete, but we can't get rid of it for now.)

## Testing

Ran `cover` and `playwright` tests, updated snapshot.


ui/appui-react/src/appui-react/theme/ThemeManager.tsx
```diff
@@ -13,18 +13,51 @@ import { ThemeProvider } from "@itwin/itwinui-react";
...
+/**
+ * List of valid theme values
  * @public
  */
-export const SYSTEM_PREFERRED_COLOR_THEME = "SYSTEM_PREFERRED";
+export type ThemeValues =
+  | keyof {
+      [P in keyof typeof ColorTheme as `${(typeof ColorTheme)[P]}`]: "";
+    }
+  | (string & {});
```
GerardasB, line 41:
```suggestion
export type ThemeType = `${ColorTheme}` | (string & {});
```
We could simplify this by using template literal types.
I like the object intersection approach to keep the auto-completion.
I think we should keep the name singular, maybe rename it to `ThemeType`?

ui/appui-react/src/appui-react/theme/ThemeManager.tsx
```diff
@@ -13,18 +13,51 @@ import { ThemeProvider } from "@itwin/itwinui-react";
...
   Dark = "dark",
+  System = SYSTEM_PREFERRED_COLOR_THEME,
+  Inherit = "inherit",
+  HighContrastLight = "high-contrast-light",
+  HighContrastDark = "high-contrast-dark",
 }
 
-/** System preferred color theme.
+/**
+ * List of valid theme values
```
GerardasB, line 34:
```suggestion
 * Describes valid theme types.
```

common/api/appui-react.api.md
```diff
@@ -795,7 +795,15 @@ export enum ColorTheme {
     // (undocumented)
     Dark = "dark",
     // (undocumented)
-    Light = "light"
+    HighContrastDark = "high-contrast-dark",
```
GerardasB, line 798: This might not be great if a custom high contrast theme is needed (specifically an iTwinUI `themeOption`).

ui/appui-react/src/appui-react/theme/ThemeManager.tsx
raplemie: Nice, spent a couple hours trying to figure that one and never found this... Definitely what I wanted to do in the first place.

`ThemeType` is the value that iTwinUI-react is using for this, I'd rather not use a clashing name.

common/api/appui-react.api.md
raplemie: iTwinUI is not offering that kind of random naming option, and we no longer have any colors of our own, if someone want to create a custom theme for iTwinUI, they need to override their color vars, and I doubt that people will go through and create "light" "dark" and high contrast variant of their theme. (If they do, they should probably use `"Inherit"` anyway)

At this point, our `ThemeManager` is no longer necessary but to preserve compatibility with someone who might use overrides in our theme, I've left it as is (updating the `<html data-theme={...}>` attribute  . Adding "high contrast" as a separate feature/setting will add to the complexity whereas I'd rather completely remove this theme system altogether.

ui/appui-react/src/appui-react/theme/ThemeManager.tsx
raplemie: Changed (but did not put in "types", as "A type that describe a type" seems too much.)

GerardasB (approved):

raplemie:
@GerardasB  This feature is not present in the release notes (I forgot NextVersion.md, I dont know if you want to update the release notes) it was released in 4.7.0.

GerardasB:
> @GerardasB This feature is not present in the release notes (I forgot NextVersion.md, I dont know if you want to update the release notes) it was released in 4.7.0.

Updated the [4.7.0 release notes](https://github.com/iTwin/appui/releases/tag/release%2F4.7.0) to reference this change.

PR #548, opened 2023-10-24
Do not emit Hidden widget state changed event when dragging a widget

GerardasB:
## Changes

This PR fixes an issue where a Hidden widget state changed event is emitted when dragging a widget tab.

## Testing

Tested via the test app.


GerardasB:

raplemie (approved):
We'll probably want to add some tests so we dont revert this by error at some point.

PR #552, opened 2023-10-24
Fix default `isResizable` and defaultState `Floating`,  Add additional `Widget` stories

### LINKED ISSUE #550: WidgetState.Floating is not causing widget to float initially
Widget with a `defaultState` set to `WidgetState.Floating` is not floating initially, unless `canFloat` property is provided.

GerardasB:
## Changes

Add additional stories for a `Widget` configuration.
Fixes #550 
Fixes an issue to correctly enable `isResizable` flag for every floating widget unless that is disabled explicitly.

## Testing

Tested via storybook and added additional stories.


raplemie (approved):
I think we should put that in 4.6.3 as well.

GerardasB:

PR #534, opened 2023-10-13
Support non base64 data:image/svg+xml icons.

raplemie:
## Changes

Update IconWebComponent to also support svg data uri that are not base64 encoded, this is what the default esbuild dataurl is exporting. We are still doing the same sanitization we used to do for base64 data.

## Testing

- Updated base64 test to validate that the SVG is actually rendered
- Added a similar test for a "raw" svg

## Issue
resolves #421
helps for #521

PR #556, opened 2023-10-24
Fix datauri with comma support in IconWebComponent

raplemie:
## Changes

Additional fix for #534 which was incomplete, according to https://www.rfc-editor.org/rfc/inline-errata/rfc2397.html and https://www.rfc-editor.org/rfc/inline-errata/rfc2396.html, commas are valid characters in a data uri.

However, this will not fix direct import from `esbuild` as they are not generating valid datauri according to specs. (but all valid datauri should work now)

## Testing

Updated tests to use more "reserved" but valid characters in the test.


raplemie:
Didnt add a NextVersion.md as I hope to put that fix in 4.6.3 to complete this issue.

GerardasB:

PR #557, opened 2023-10-24
Resize Popout to fit content on popout

joehenry9498:
## Changes
Originally popouts were set to the default window size of child windows on popout. They will now fit to their content inside on popout. This couldn't be done like how it was with floating widgets because that uses hooks and the popout needs to use class components. The preferredFloatingWidgetSize of the tab will be used to find the inital popout size when bounds are not set.

## Testing
I tested the popout in the popout window frontstage. Pop it out from docking on each side and as a floating widget. I also tested in the exercise widget api frontstage.


ui/appui-react/src/appui-react/childwindow/InternalChildWindowManager.tsx
```diff
@@ -271,12 +271,25 @@ export class InternalChildWindowManager implements FrameworkChildWindows {
...
+
+    if (nineZoneState === undefined) return false;
+
+    const popoutContentContainer = document.getElementById(
+      `content-container:${nineZoneState.widgets[childWindowId].activeTabId}`
+    );
+    if (popoutContentContainer === null) return false;
+
+    const contentWidth = popoutContentContainer.offsetWidth + 20;
+    const contentHeight = popoutContentContainer.offsetHeight + 20;
```
raplemie, line 285: The `open` method is public and a user can put anything in a child window (we just happen to use it ourselves directly for popouts) I think what you are trying to do can work by setting the `location.width` and `location.height` to `contentWidth` and `contentHeight` from wherever we call this method for the popouts though.

joehenry9498: Moved to internal method

ui/appui-react/src/appui-react/childwindow/InternalChildWindowManager.tsx
```diff
@@ -276,7 +276,7 @@ export class InternalChildWindowManager implements FrameworkChildWindows {
     const childWindow = window.open(
       url,
       "",
-      `width=${location.width},height=${location.height},left=${location.left},top=${location.top},menubar=no,resizable=yes,scrollbars=no,status=no,location=no`
+      `width=${location.width}},height=${location.height},left=${location.left},top=${location.top},menubar=no,resizable=yes,scrollbars=no,status=no,location=no`
```
raplemie, line 279: Typo
```suggestion
      `width=${location.width},height=${location.height},left=${location.left},top=${location.top},menubar=no,resizable=yes,scrollbars=no,status=no,location=no`
```

joehenry9498:
Working on adjusting it so that it is only sized like this on initial popout and not after resized

joehenry9498:
Fixed it all up so that behavior is expected now. Will only fit to content when popout bounds are not set

docs/changehistory/NextVersion.md
```diff
@@ -10,6 +10,8 @@ Table of contents:
   - [Fixes](#fixes-1)
 - [@itwin/components-react](#itwincomponents-react)
   - [Fixes](#fixes-2)
+- [@itwin/appui-layout-react](#itwinappui-layout-react)
+  - [Changes](#changes)
```
raplemie, line 14: @GerardasB As this package is internal, should we have NextVersion for it ?

ui/appui-layout-react/src/test/state/NineZoneStateReducer.test.ts
```diff
@@ -1619,6 +1620,22 @@ describe("NineZoneStateReducer", () => {
...
+
+      let state = createNineZoneState();
+      state = addTab(state, "t1");
+      state = addPanelWidget(state, "left", "w1", ["t1"]);
+
+      const newState = NineZoneStateReducer(state, {
+        type: "WIDGET_TAB_POPOUT",
+        id: "t1",
+      });
+      newState.popoutWidgets.allIds.should.length(1);
```
raplemie, line 1637: Shouldn't we check the size ? How is this validating that we are indeed fitting to the content container ?

joehenry9498: It was just checking that it successful created the popout widget using the content container as the bounds. Added in the size check for the blank div + the buffer space

ui/appui-layout-react/src/appui-layout-react/state/NineZoneStateReducer.ts
```diff
@@ -530,9 +530,31 @@ export function NineZoneStateReducer(
...
+
+      let contentHeight = 800;
+      let contentWidth = 600;
+
+      if (tab.preferredFloatingWidgetSize) {
+        contentWidth = tab.preferredFloatingWidgetSize.width;
+        contentHeight = tab.preferredFloatingWidgetSize.height;
+      } else {
+        const popoutContentContainer = document.getElementById(
+          `content-container:${savedTab?.id}`
```
raplemie, line 544: If the `savedTab` wasn't found, this is when we'll actually want to get the container size, and if it's not found, then the id will be `undefined`, as this is the `id` anyway and we are doing a `getElementById`, might as well simply use the provided `id`.
```suggestion
          `content-container:${id}`
```

raplemie (approved):
With the suggested fix, seems to be working quite good !

docs/changehistory/NextVersion.md
```diff
@@ -88,3 +90,9 @@ Table of contents:
 ### Fixes
 
 - Fixed `useDisposable` hook to work in React 18 strict mode.
+
+## @itwin/appui-layout-react
+
+### Changes
+
+- Popouts will now fit to content when bounds are not set.
```
raplemie, line 98: We might want to revise this as the behavior actually changed.

docs/changehistory/NextVersion.md
GerardasB: I'd say no and put the changes under appui-react.

joehenry9498: Moved to appui-react.

PR #560, opened 2023-10-25
Initial keyboard shortcuts story.

GerardasB:
## Changes

Add an initial story about keyboard shortcuts.

## Testing

N/A


raplemie (approved):
Seems good, is the goal to show that someone can use something else than `FrameworkToolAdmin` to handle when the events are triggered by themselves ? (Instead of only handling events when the focus is set to home)

GerardasB:
> Seems good, is the goal to show that someone can use something else than `FrameworkToolAdmin` to handle when the events are triggered by themselves ? (Instead of only handling events when the focus is set to home)

Yes, that's one of the use cases, I need to wrap up the model selector to create a story with `FrameworkToolAdmin`.

PR #574, opened 2023-10-31
Fix `IModelViewportControl` in case `applicationData` is not provided

GerardasB:
## Changes

This PR fixes a reference error in `IModelViewportControl` in case the `applicationData` property is not provided.

## Testing

N/A


ui/appui-react/src/appui-react/content/IModelViewport.tsx
```diff
@@ -102,7 +102,9 @@ export class IModelViewportControl extends ViewportContentControl {
     info: ConfigurableCreateInfo,
     protected _options: IModelViewportControlOptions
   ) {
+    _options = _options ?? {};
     super(info, _options);
+    this._options = _options;
```
raplemie, line 107: This shouldnt be needed, after a quick tests, changing the `_options` before calling `super` makes it save the parameter correctly, on top of that, the base class already saves `_options` as a `protected _appDataOptions` which is accessible as `this.appDataOptions`, so this "protected" redefinition is not needed at all.

(And as the `_options` parameter is not optional, where is this issue comming from ?)

GerardasB: `IModelViewportControl` has two fields `_options` (strongly typed) and `_appDataOptions` (**any** type) pointing to the same object. I agree that it should be enough to have `_appDataOptions` with some helper/cast, however we can't just remove `_options` ATM w/o deprecation and next major version. I did some testing and it appears that not setting the `this._options` field directly only updates _options variable (in that constructor scope). I.e. calling `this._options` from other methods will continue to point to `undefined` instead of empty object.

The control is created in `ContentGroup.tsx` and due to poor typing `_options` (which is required) is `undefined` when `applicationData` (which is optional) is not provided.

raplemie: Ok, I must have screwed up my attempt at testing this with a sandbox then :) !

raplemie (approved):

PR #576, opened 2023-11-02
Update UI for components-react FilterBuilder

morganadrales:
## Changes
Visual updates for the Filter Builder in components-react, based on discussion between the Pineapple and presentation teams.

Before:
image

After:
(small)
image

(medium)
image

(large)
image

API changes include removal of nesting, as we will be able to support "IN" and "BETWEEN" operators instead.

## Testing

Updated tests to reflect the removal of nesting. Maintained the package's 100% test coverage. 

Raphael added a storybook component to test the different sizes of the editors populatable as the filter builder property. The result of those different sizes is below:
image


raplemie (commented):
Before I review more, can I get an explanation as to why we duplicate all these files ? Is that intended ? Even though there is a folder named "deprecated", none of the components in them are marked as such, if the renaming was needed why not use aliases instead of duplicating the code ?

grigasp:
@morganadrales, I thought we discussed deprecating only the prop rather than the whole component. Was there a reason to change the plan?

morganadrales:
> @morganadrales, I thought we discussed deprecating only the prop rather than the whole component. Was there a reason to change the plan?

Hi Grigas. Apologies, I created this PR before having to hastily leave before my honeymoon and I should've marked it as draft to prevent premature reviewal.

Originally, since the logic for supporting nesting has been removed, I duplicated the original component to prevent a breaking change. However, I have now updated the PR so that the ruleDepthLimit prop has been deprecated instead to prevent a breaking change - although note that it's still not being used since all code related to groups/nesting has been removed.

I still have some work to do on this PR to reach 100% code coverage (currently at 99.5%) but would you mind taking another pass at this PR? I can address any additional concerns while I attempt to get the code coverage back up to 100%.

raplemie:
@FlyersPh9 Can you have a look at the CSS, there is a lot of changes and px values, maybe we can/should do it differently.

ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
```diff
@@ -223,7 +220,7 @@ export interface UsePropertyFilterBuilderProps {
  * Options for [[UsePropertyFilterBuilderResult.buildFilter]].
  * @beta
  */
-export interface BuildFilterOptions {
+export interface FilterOptions {
```
raplemie, line 223: Keep the previous name but deprecate it, so we can reduce code breaking change to a minimum. (Even if it is unlikey that someone actually imported this by name...)
```suggestion
 * @deprecated in 4.7.0. Use [[FilterOptions]].
 */
export type BuildFilterOptions = FilterOptions;

/** Options for [[UsePropertyFilterBuilderResult.buildFilter]].
 * @beta
 */
export interface FilterOptions {
```

raplemie (changes requested):

FlyersPh9:
> @FlyersPh9 Can you have a look at the CSS, there is a lot of changes and px values, maybe we can/should do it differently.

I just hopped on a call with @morganadrales and looked through the `SCSS` files together.  Quick summary:
- Where possible, use [an iTwinUI CSS variable](https://itwinui.bentley.com/docs/variables) instead of a hard coded pixel value, this would consist of:
  - Border radius
  - Sizing / spacing
  - Font sizes
  - One exception being border thickness staying at `1px`.  We are considering adding a `--iui-border-thickness` variable in a minor release.
- In other areas where an iTwinUI size CSS variable might not match exactly, converting PX to REM I think will be a worthwhile change.
- I saw `:focus { outline: 0 }` in there.  Wasn't sure where it was being applied, but just make sure everything that is actionable can be focused via pressing the TAB key on the keyboard.
- And bonus points: iTwinUI switched to [logical properties](https://css-tricks.com/css-logical-properties-and-values/). Example: `margin-right` to `margin-inline-end`.  Would be nice to match.

ui/components-react/src/components-react/filter-builder/FilterBuilderRule.tsx
```diff
@@ -31,7 +34,10 @@ export interface PropertyFilterBuilderRuleRendererProps {
   path: string[];
   /** Rule to render. */
   rule: PropertyFilterBuilderRule;
-  isRemovable?: boolean;
+  /** Function to add rule to group */
+  onAddRule: () => void;
```
saskliutas, line 38: Why does it need `onAddRule` callback in props? It has access to `PropertyFilterBuilderActions` and uses them for rule removal.

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
```diff
@@ -61,24 +62,42 @@ export function PropertyFilterBuilderRuleGroupRenderer(
...
+            <PropertyFilterBuilderGroupOrRule
+              path={path}
+              item={item}
+              onAddRule={handleAddRule}
+              size={size}
+            />
+          </Flex>
+        ))}
+      </div>
       <div className="rule-group-remove-action">
```
saskliutas, line 101: Shouldn't this be removed now that nesting is not supported?

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
```diff
@@ -61,24 +62,42 @@ export function PropertyFilterBuilderRuleGroupRenderer(
...
 
   return (
-    <div
+    <Flex
       ref={groupRef}
-      className="rule-group"
+      style={{ alignSelf: "flex-start" }}
+      className="fb-group fb-group-criteria fb-criteria-container"
+      gap="0px"
       data-isactive={
```
saskliutas, line 77: Looks like this is not used anymore

ui/components-react/src/components-react/filter-builder/FilterBuilderLogicalOperator.tsx
```diff
@@ -0,0 +1,61 @@
...
+  const toggle = () => (operator === "And" ? "Or" : "And");
+
+  return (
+    <div
+      className={cx("fb-logical-operator", className)}
+      data-iui-size={size}
+      {...rest}
+    >
+      {undefined === operator ? (
+        <span>{UiComponents.translate("filterBuilder.group")}</span>
```
saskliutas, line 53: What does `Group` means in context of this component? I don't think that whole `PropertyFilterBuilder` component should be kept generic (list of rules/criterias based on provided properties) and groups/grouping concept should be introduced in component that is built on top of `PropertyFilterBuilder`.

ui/components-react/src/components-react/filter-builder/FilterBuilderLogicalOperator.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Props for [[PropertyFilterBuilderLogicalOperator]] component.
+ * @internal
+ */
+interface PropertyFilterBuilderLogicalOperatorProps {
+  /** Size to render the component. If undefined, defaults to iTwinUI "medium" size. */
+  size?: "small" | "large";
+  /** Allows toggling of operator by clicking operator text. */
+  isLinkDisabled?: boolean;
```
saskliutas, line 21: I think better name for this property would be `isDisabled` or `isReadonly`.

ui/components-react/src/components-react/filter-builder/FilterBuilderLogicalOperator.tsx
```diff
@@ -0,0 +1,61 @@
...
+  /** Operator to combine FilterBuilderRules. Must be either "And" or "Or". */
+  operator?: Operator;
+  /** Callback that is invoked when operator changes. */
+  onOperatorChange?: (operator: Operator) => void;
```
saskliutas, line 25: Don't think this needs to be optional.

saskliutas, line 25: Don't think these need to be optional

ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
saskliutas: Was there are reason for renaming this interface?

ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
```diff
@@ -337,20 +332,20 @@ export function defaultPropertyFilterBuilderRuleValidator(
 }
 
 /** @internal */
-export function buildPropertyFilter(
+export function createFilter(
```
saskliutas, line 335: Was there a reason for renaiming these functions?

ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
```diff
@@ -91,14 +91,11 @@ export class PropertyFilterBuilderActions {
   }
 
   /** Adds new rule or group of rules to the group specified by path. */
-  public addItem(path: string[], itemType: "RULE_GROUP" | "RULE") {
+  public addItem(path: string[]) {
```
saskliutas, line 94: I don't think there is any reason to remove nesting support from PropertyFilterBuilder state and actions. It should be enough to change UI components to not use nesting. That way same state and actions implementation could be used for different UI components that do support nesting.

ui/components-react/src/components-react/filter-builder/FilterBuilderToolbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * Props for [[PropertyFilterBuilderToolbar]] component.
+ * @internal
+ */
+interface PropertyFilterBuilderToolbarProps {
+  /** Size to render the component. If undefined, defaults to iTwinUI "medium" size. */
+  size?: "small" | "large";
+  /** Function to add child of current group. */
+  onAddChild?: () => void;
+  /** Function to remove child from current group. */
+  onDelete?: () => void;
```
saskliutas, line 21: I think these should be made required as buttons for these actions are always rendered.

ui/components-react/src/components-react/filter-builder/FilterBuilderLogicalOperator.tsx
```diff
@@ -0,0 +1,61 @@
...
+import "./FilterBuilderLogicalOperator.scss";
+import { UiComponents } from "../UiComponents";
+
+type Operator = "And" | "Or";
+
+/**
+ * Props for [[PropertyFilterBuilderLogicalOperator]] component.
+ * @internal
+ */
+interface PropertyFilterBuilderLogicalOperatorProps {
```
saskliutas, line 17: I think this needs to be exported and tagged as `beta`.

ui/components-react/src/components-react/filter-builder/FilterBuilderToolbar.tsx
```diff
@@ -0,0 +1,59 @@
...
+}
+
+/** Toolbar displaying the "add" and "delete" row buttons in the filter builder
+ * @beta
+ */
+export const PropertyFilterBuilderToolbar = (
+  props: FlexProps & PropertyFilterBuilderToolbarProps
```
saskliutas, line 28: @GerardasB @raplemie Doesn't exposing `itwinui-react` types require to make `itwinui-react` as peer dependency?

raplemie: That's already a direct dependency, not sure why it would need to become a peer, not thrilled to expose these dependencies like that though, is this necessary ? iTwinUI appears to be hoping to move much faster that we actually want to ourselves with Major (`FlexProps` type is probably no longer exposed anyway in iTwinUI 3.0 as they removed most of their props types...)

I'm not sure what all the FlexProps are, but I doubt that it make much sense at this point to let user change these for the toolbar, these should be uniformized, not free for all I think

ui/components-react/src/components-react/editors/EditorContainer.tsx
```diff
@@ -47,6 +47,8 @@ export interface PropertyEditorProps extends CommonProps {
    * @internal
    */
   shouldCommitOnChange?: boolean;
+  /** Size to render the component. If undefined, defaults to iTwinUI "medium" size. */
+  size?: "small" | "large";
```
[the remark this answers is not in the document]
morganadrales: iTwinUI uses the pattern of "small" | "large" for sizing and the default "medium" if unspecified so I've kept that, but set the component size to small in the PropertyFilterBuilderRenderer. I've updated the screenshots to reflect the different sizes as well.

ui/components-react/src/components-react/filter-builder/FilterBuilderLogicalOperator.tsx
morganadrales: so the idea is that a group is 2 or more rules such that the "+" button on each rule will add an additional rule to the group. The operator here is on the group level and would be "And" in the below screenshot.
image
if we only want to support a singular rule in this component and have the higher order component handle the grouping, then we may need to rethink how rule addition will work since all of the addition exists within the rule/rule group components.

morganadrales: Ah ok, sorry for the misunderstanding. Yes, the group string should be able to be removed.

ui/components-react/src/components-react/filter-builder/FilterBuilder.scss
```diff
@@ -0,0 +1,132 @@
...
+    }
+  }
+
+  .fb-component-row:not(.fb-component-indent) {
+    ~ .fb-component-indent {
+      > div {
+        border-radius: 0;
+      }
+
+      &:nth-last-child(1 of .fb-component-indent) > div {
```
raplemie, line 53: @FlyersPh9 are we fine using css that is not 1 year old yet for most browsers ? Are these expected to be of different classes than `.fb-component-indent` ?

ui/components-react/src/components-react/filter-builder/FilterBuilderLogicalOperator.tsx
```diff
@@ -0,0 +1,61 @@
...
+    >
+      {undefined === operator ? (
+        <span>{UiComponents.translate("filterBuilder.group")}</span>
+      ) : isDisabled ? (
+        <span>
+          {operator === PropertyFilterRuleGroupOperator.And ? "And" : "Or"}
+        </span>
+      ) : (
+        <Anchor onClick={() => onOperatorChange(toggle())}>
+          {operator === PropertyFilterRuleGroupOperator.And ? "And" : "Or"}
```
raplemie, line 56: These "And" : "Or" should be translated as well here.

ui/components-react/src/components-react/editors/EditorContainer.tsx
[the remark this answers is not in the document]
raplemie: This change is affecting the components even when they are not in the property filter, like in tool settings, or anyone directly using these component will suddenly see size change. At best, we should add "medium" as a choice and default to small when not provided.

Master: 
image

PR:
image

ui/components-react/src/components-react/filter-builder/FilterBuilderRule.tsx
morganadrales: it at least needs the onNewRuleAdded callback from FilterBuilderRuleGroup since that is called after actions.addItem to focus the newly added row

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
morganadrales: We need both the PropertyFilterBuilderRuleGroupOperator and the PropertyFGilterBuilderGroupOrRule to be rendered here, otherwise nothing is shown. We are still supporting grouping (i.e. 2 or more rules linked by the same oeprator), we're just not supporting nesting i.e. indenting rule(s) under an additional rule group.

ui/components-react/src/components-react/filter-builder/FilterBuilder.tsx
```diff
@@ -95,7 +90,9 @@ export interface PropertyFilterBuilderRendererProps {
   ) => React.ReactNode;
   /** Custom renderer for property selector in rule. */
   propertyRenderer?: (name: string) => React.ReactNode;
-  /** Specifies how deep rule groups can be nested. */
+  /** Specifies how deep rule groups can be nested.
+   * @deprecated in 4.7.2. Nesting is no longer supported moving forward.
```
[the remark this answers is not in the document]
raplemie: Will be `4.8.0` now.

ui/components-react/src/components-react/filter-builder/FilterBuilderRule.tsx
saskliutas: In that case I would prefer renaming this to `onRuleAdded` and pass `onNewRuleAdded` from `FilterBuilderRuleGroup` to this component.

ui/components-react/src/components-react/filter-builder/FilterBuilder.tsx
```diff
@@ -150,12 +154,13 @@ export function PropertyFilterBuilderRenderer(
         <ActiveRuleGroupContext.Provider
```
saskliutas, line 154: I think this `ActiveRuleGroupContext` is not user anywhere. Could you remove it and all related code?

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
```diff
@@ -155,59 +104,57 @@ export interface PropertyFilterBuilderRuleGroupOperatorProps {
...
-      },
-      {
-        value: PropertyFilterRuleGroupOperator.Or,
-        label: UiComponents.translate("filterBuilder.operators.or"),
-      },
-    ],
-    []
-  );
+  const { operator, size } = props;
+  const [op, setOp] = React.useState(operator);
```
saskliutas, line 108: Why is this state needed?

ui/components-react/src/components-react/filter-builder/FilterBuilderToolbar.tsx
```diff
@@ -0,0 +1,58 @@
...
+  props: PropertyFilterBuilderToolbarProps
+) => {
+  const { size, onAddChild, onDelete, ...rest } = props;
+
+  return (
+    <Flex
+      className={"fb-toolbar fb-row-toolbar"}
+      gap="0"
+      justifyContent="flex-end"
+      {...rest}
```
saskliutas, line 36: Looks like this can be removed.

ui/components-react/src/components-react/filter-builder/FilterBuilderToolbar.tsx
```diff
@@ -0,0 +1,58 @@
...
+    >
+      <IconButton
+        size={!size ? "small" : size === "medium" ? undefined : "large"}
+        data-testid="fb-add-rule-button"
+        label={UiComponents.translate("filterBuilder.add")}
+        styleType="borderless"
+        onClick={() => onAddChild?.()}
```
saskliutas, line 43:
```suggestion
        onClick={onAddChild}
```

ui/components-react/src/components-react/filter-builder/FilterBuilderToolbar.tsx
```diff
@@ -0,0 +1,58 @@
...
+      >
+        <SvgAdd />
+      </IconButton>
+      <IconButton
+        size={!size ? "small" : size === "medium" ? undefined : "large"}
+        data-testid="fb-remove-rule-button"
+        label={UiComponents.translate("filterBuilder.delete")}
+        styleType="borderless"
+        onClick={() => onDelete?.()}
```
saskliutas, line 52:
```suggestion
        onClick={onDelete}
```

ui/components-react/src/components-react/filter-builder/FilterBuilderRule.tsx
```diff
@@ -113,50 +123,48 @@ export function PropertyFilterBuilderRuleRenderer(
...
-                <SvgStatusError className="iui-input-icon" />
-                <div className="iui-message">{rule.errorMessage}</div>
-              </>
-            ) : null}
-          </div>
-        ) : null}
-      </div>
+          ) : null}
+          <PropertyFilterBuilderToolbar
+            onAddChild={() => handleRuleAdded()}
```
saskliutas, line 162:
```suggestion
            onAddChild={handleRuleAdded}
```

ui/components-react/src/test/filter-builder/FilterBuilderState.test.ts
```diff
@@ -71,67 +67,10 @@ describe("usePropertyFilterBuilder", () => {
     ).to.be.true;
   });
 
-  it("adds rule to nested group", async () => {
```
saskliutas, line 74: I think these tests should be added back as state still supports nesting.

ui/components-react/src/test/filter-builder/FilterBuilderRuleValue.test.tsx
```diff
@@ -12,7 +12,7 @@ import { render, screen, waitFor } from "@testing-library/react";
 import { PropertyFilterBuilderRuleValue } from "../../components-react/filter-builder/FilterBuilderRuleValue";
 import TestUtils, { userEvent } from "../TestUtils";
 
-describe("PropertyFilterBuilderRuleValue", () => {
+describe("FilterBuilderRuleValue", () => {
```
saskliutas, line 15:
```suggestion
describe("PropertyFilterBuilderRuleValue", () => {
```

ui/components-react/src/test/filter-builder/FilterBuilder.test.tsx
```diff
@@ -66,8 +66,8 @@ describe("PropertyFilterBuilder", () => {
     });
   });
 
-  it("renders propertyFilterBuilder with single rule correctly", async () => {
-    const propertyFilter: PropertyFilter = {
+  it("renders PropertyFilterBuilder with single rule correctly", async () => {
+    const PropertyFilter: PropertyFilter = {
```
saskliutas, line 70: Variable names should be in camelCase.

ui/components-react/src/test/filter-builder/FilterBuilderRuleGroup.test.tsx
```diff
@@ -129,123 +95,4 @@ describe("PropertyFilterBuilderRuleGroupRenderer", () => {
...
-    );
-    const removeItemSpy = sinon.stub(actions, "removeItem");
-
-    const removeButton = getByTestId("rule-group-remove");
-    removeButton.click();
-
-    expect(removeItemSpy).to.be.calledOnceWith(defaultProps.path);
-  });
-
-  it("dispatches operator change event when operator is selected", async () => {
```
saskliutas, line 223: I think this test should not be removed. We still need to check if operator changes for root group that is render.

saskliutas (changes requested):

ui/components-react/public/locales/en/UiComponents.json
```diff
@@ -123,6 +123,9 @@
   "filterBuilder": {
     "rule": "Rule",
     "ruleGroup": "Rule Group",
+    "group": "Group",
```
saskliutas, line 126: I think this is unused.

ui/components-react/src/components-react/filter-builder/FilterBuilder.tsx
```diff
@@ -95,10 +94,16 @@ export interface PropertyFilterBuilderRendererProps {
...
+  /** Specifies how deep rule groups can be nested.
+   * @deprecated in 4.8.0. Nesting is no longer supported moving forward.
+   */
   ruleGroupDepthLimit?: number;
   /** Specifies whether component is disabled or not. */
   isDisabled?: boolean;
+  /** Size to render the component. If undefined, defaults to iTwinUI "small" size. */
+  size?: "medium" | "large";
+  /** Controls whether the group operator is toggle-able. */
+  isOperatorToggleDisabled?: boolean;
```
saskliutas, line 106: I would suggest renaming to `isGroupOperatorDisabled` or `isLogicalOperatorDisabled`.

ui/components-react/src/components-react/filter-builder/FilterBuilderLogicalOperator.tsx
```diff
@@ -0,0 +1,63 @@
...
+    <div
+      className={cx("fb-logical-operator", className)}
+      data-iui-size={size}
+      {...rest}
+    >
+      {isDisabled ? (
+        <span>
+          {operator === PropertyFilterRuleGroupOperator.And
+            ? UiComponents.translate("filterBuilder.operators.and")
+            : UiComponents.translate("filterBuilder.operators.or")}
```
saskliutas, line 52: Move this into local variable to avoid duplicating code.

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.tsx
```diff
@@ -155,59 +109,56 @@ export interface PropertyFilterBuilderRuleGroupOperatorProps {
...
-        value={operator}
-        onChange={onChange}
-        size="small"
+    <Flex.Item flex="0" alignSelf="stretch">
+      <PropertyFilterBuilderLogicalOperator
+        className="fb-group-operator"
+        operator={operator}
+        onOperatorChange={(value: PropertyFilterRuleGroupOperator) => {
+          onChange(value);
+        }}
```
saskliutas, line 121:
```suggestion
        onOperatorChange={onChange(value)}
```

saskliutas (changes requested):

ui/components-react/src/components-react/filter-builder/FilterBuilderLogicalOperator.tsx
```diff
@@ -0,0 +1,62 @@
...
+  /** Callback that is invoked when operator changes. */
+  onOperatorChange: (operator: PropertyFilterRuleGroupOperator) => void;
+}
+
+/** Component to render the operator inside of the filter builder
+ * @beta
+ */
+export const PropertyFilterBuilderLogicalOperator = (
+  props: Omit<React.HTMLProps<HTMLDivElement>, "size"> &
+    PropertyFilterBuilderLogicalOperatorProps
```
saskliutas, line 32:
```suggestion
  props: PropertyFilterBuilderLogicalOperatorProps
```

Doesn't look like any of `HTMLDivElement` props are passed when rendering this component. Also other FilterBuilder components do not extend props of native HTML elements.

saskliutas (changes requested):

ui/components-react/src/components-react.ts
```diff
@@ -230,3 +232,11 @@ export * from "./components-react/tree/controlled/internal/SparseTree";
...
  * Classes and components for working with PropertyFilterBuilder.
  */
+/**
+ * @docs-group-description FilterBuilderLogicalOperator
+ * Operator used in the PropertyFilterBuilder. Supports 'And' or 'Or'.
+ */
+/**
+ * @docs-group-description FilterBuilderToolbar
+ * Functions and components that provide a toolbar for the PropertyFilterBuilder.
+ */
```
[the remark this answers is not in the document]
morganadrales: Could you clarify that? Does that mean add them to the PropertyFilterBuilder docs-group-description? Sorry, I'm not familiar with your documentation system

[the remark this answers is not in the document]
saskliutas: You can add comment like this https://github.com/iTwin/appui/blob/857067cb4a3ecd8d17bc96d43dc4a2adabd88eb9/ui/components-react/src/components-react/filter-builder/FilterBuilderRule.tsx#L5 to those files to specify docs group.

ui/components-react/src/components-react/filter-builder/FilterBuilderLogicalOperator.tsx
```diff
@@ -0,0 +1,61 @@
...
+  onOperatorChange: (operator: PropertyFilterRuleGroupOperator) => void;
+}
+
+/** Component to render the operator inside of the filter builder
+ * @beta
+ */
+export const PropertyFilterBuilderLogicalOperator = (
+  props: PropertyFilterBuilderLogicalOperatorProps
+) => {
+  const { className, size, isDisabled, operator, onOperatorChange, ...rest } =
```
saskliutas, line 33: Add `className` to props interface and remove `...rest`.

saskliutas (changes requested):

ui/components-react/src/components-react/filter-builder/FilterBuilder.scss
```diff
@@ -0,0 +1,29 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+.fb-component {
+  --fb-operator-width: 4.5rem;
+  --divider-height: var(--iui-size-s);
+  --divider-width: var(--iui-size-2xs);
+  --divider-background-color: var(--iui-color-border-subtle);
+
+  // set the CSS var for the width of the logical operator (if size=small)
+  [data-iui-size="small"] {
```
GerardasB, line 12: Could we avoid referencing iTwinUI attributes?

ui/components-react/src/components-react/filter-builder/FilterBuilder.scss
```diff
@@ -0,0 +1,29 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+.fb-component {
```
GerardasB, line 5: We are not using CSS modules, could we make root selectors more unique? I.e. 
`.components-filterBuilder-filterBuilder { .fb-wrapper { /* ... */ } }`

GerardasB (changes requested):

ui/components-react/src/components-react/filter-builder/FilterBuilderRule.scss
```diff
@@ -2,44 +2,70 @@
...
-    padding-top: 3px;
+  .fb-row-toolbar {
+    width: var(--toolbar-width);
+    margin-left: auto;
   }
 }
+
+.fb-row:not(:last-child) {
+  //only show buttons on last row to prevent visual clutter
+  .iui-button:first-child {
```
GerardasB, line 63: Avoid referencing iTwinUI internal class names. The preferred way is to add your class name to the component and select based on that.

GerardasB (changes requested):

ui/components-react/src/components-react/filter-builder/FilterBuilderRule.scss
```diff
@@ -2,44 +2,70 @@
...
   }
 
-  &:focus-within .rule-remove-action,
-  &:hover .rule-remove-action {
-    visibility: visible;
+  .fb-row-container {
+    flex: 1;
+  }
+
+  .fb-row-operator {
```
raplemie, line 37: This class is not used anywere

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleGroup.scss
```diff
@@ -2,67 +2,83 @@
...
+  padding: var(--iui-size-xs);
+  align-self: flex-start;
+}
 
-  &:last-child {
-    padding-bottom: 11px;
-  }
+.fb-rule-group {
+  flex-direction: column;
 }
```
raplemie, line 84: These 2 classes do not appear to be used anywhere.

ui/components-react/src/components-react/filter-builder/FilterBuilderLogicalOperator.tsx
```diff
@@ -0,0 +1,61 @@
...
+      : PropertyFilterRuleGroupOperator.And;
+
+  const operatorDisplayText =
+    operator === PropertyFilterRuleGroupOperator.And
+      ? UiComponents.translate("filterBuilder.operators.and")
+      : UiComponents.translate("filterBuilder.operators.or");
+
+  return (
+    <div className={cx("fb-logical-operator", className)} data-iui-size={size}>
+      {isDisabled ? (
```
raplemie, line 52: Dont reuse `data-iui`, as these should be reserved to iTwinUI in our ecosystem.

raplemie (changes requested):

ui/components-react/src/components-react/filter-builder/FilterBuilderRuleValue.tsx
```diff
@@ -66,6 +68,7 @@ export function PropertyFilterBuilderRuleValue(
       onCommit={onValueChange}
       setFocus={false}
       shouldCommitOnChange={false}
+      size={size}
```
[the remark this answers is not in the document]
raplemie: image
Example in Storybook, the button group are not of the correct size.

raplemie (changes requested):
I'll try to provide a story or an example showcasing all the visual so we can more easily detect the issues.
image

ui/components-react/src/components-react/iconpicker/IconPickerButton.tsx
```diff
@@ -165,13 +173,14 @@ export class IconPickerButton extends React.PureComponent<
 
     return (
       <>
-        <button
+        <Button
```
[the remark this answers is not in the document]
raplemie: After some more testing, this is already broken in master, we'll file an issue for it. (But apparently nobody used that for a while...)

ui/core-react/src/core-react/inputs/numberinput/NumberInput.tsx
```diff
@@ -294,7 +295,7 @@ const ForwardRefNumberInput = React.forwardRef<
         onKeyDown={handleKeyDown}
         onFocus={handleFocus}
         onBlur={handleBlur}
-        size="small"
+        size={size}
```
raplemie, line 298: This one is confusing, as we were actually defaulting to `"small"` if nothing was provided, and `"medium"` was not an option (as `"small"` and `undefined` would both be `"medium"`, and `"large"` would be the only thing doing something.)

For this one in particular, add the full options list: `"small" | "medium" | "large"` and keep `undefined` as `small` so the behavior do not change.

ui/components-react/src/components-react/editors/EnumButtonGroupEditor.tsx
```diff
@@ -214,18 +217,31 @@ export class EnumButtonGroupEditor
...
   }
 
+  private getSizeClassname(size: string | undefined) {
+    switch (size) {
+      case "large":
+        return "components-enumbuttongroup-large";
+      case undefined:
+        return "components-enumbuttongroup-small";
+      default:
+        return "components-enumbuttongroup-medium     ";
```
raplemie, line 234: Typo:
```suggestion
        return "components-enumbuttongroup-medium";
```

common/changes/@itwin/components-react/morgan-filter-builder_2023-11-02-18-38.json
raplemie, line 1: You can remove this file and keep only one.

ui/core-react/src/core-react/inputs/numberinput/NumberInput.tsx
morganadrales: The InputProps from iTwinUI that the NumberInput is extending only allows for "small", "large", or undefined - undefined implying the default "medium" size. Having undefined also be "small" in this instance I think would prevent the "medium" size from being usable

raplemie: I know, that's why I ask to override the ItwinUI props that we use with a full set `"small" "medium" "large"` and change the default (`undefined` to mean small like it used to be.)

ui/components-react/src/test/filter-builder/FilterBuilderRuleGroup.test.tsx
```diff
@@ -130,122 +102,121 @@ describe("PropertyFilterBuilderRuleGroupRenderer", () => {
...
-        group={{
-          id: "id",
-          groupId: "parentGroupId",
-          operator: PropertyFilterRuleGroupOperator.And,
-          items: [],
-        }}
-      />,
+    rootGroup.operator = PropertyFilterRuleGroupOperator.Or;
+    const props: PropertyFilterBuilderRuleGroupRendererProps = {
+      group: rootGroup,
```
saskliutas, line 173: Do not mutate `rootGroup` in test as it is shared between tests.
```suggestion
    const props: PropertyFilterBuilderRuleGroupRendererProps = {
      group: { ...rootGroup, operator: PropertyFilterRuleGroupOperator.Or },
```

ui/components-react/src/test/filter-builder/FilterBuilderRuleGroup.test.tsx
```diff
@@ -130,122 +102,121 @@ describe("PropertyFilterBuilderRuleGroupRenderer", () => {
...
+    );
+    expect(selector).to.be.null;
   });
 
-  it("dispatches operator change event when operator is selected", async () => {
+  it("'And' Operator should not be clickable if toggled disabled", async () => {
     const actions = new PropertyFilterBuilderActions(sinon.spy());
+    rootGroup.operator = PropertyFilterRuleGroupOperator.And;
+    const props: PropertyFilterBuilderRuleGroupRendererProps = {
+      group: rootGroup,
```
saskliutas, line 200:
```suggestion
    const props: PropertyFilterBuilderRuleGroupRendererProps = {
      group: rootGroup,
```

ui/components-react/src/test/filter-builder/FilterBuilderRuleGroup.test.tsx
```diff
@@ -130,122 +102,121 @@ describe("PropertyFilterBuilderRuleGroupRenderer", () => {
...
+    const { container, findByText } = renderWithContext(
+      <PropertyFilterBuilderRuleGroupRenderer {...defaultProps} size="large" />,
+      { actions }
     );
+    const setRuleGroupOperatorSpy = sinon.stub(actions, "setRuleGroupOperator");
 
-    getByText(
-      TestUtils.i18n.getLocalizedString("Components:filterBuilder.operators.or")
+    const selector = container.querySelector<HTMLAnchorElement>(
+      ".fb-group-operator .iui-anchor"
```
saskliutas, line 114: Do not use `ITwinUI` class names to select element (at least in new tests, older tests that uses `iui-*` class names should be fixed with https://github.com/iTwin/appui/pull/666). Select by text or use custom class name.

ui/components-react/src/components-react/filter-builder/FilterBuilderToolbar.tsx
```diff
@@ -0,0 +1,62 @@
...
+  return (
+    <Flex
+      className={"fb-toolbar fb-row-toolbar"}
+      gap="0"
+      justifyContent="flex-end"
+    >
+      <IconButton
+        size={getiTwinUISize(size)}
+        className="fb-add-rule-button"
+        data-testid="fb-add-rule-button"
```
saskliutas, line 43: I think `data-testid` can be removed because buttons can be selected by class name.

saskliutas (changes requested):

GerardasB:
Duplicate of https://github.com/iTwin/appui/pull/686

PR #581, opened 2023-11-09
New `enableMaximizedFloatingWidget` preview feature

raplemie:
### Preview feature

* Added new preview feature to allow floating widgets to have a maximize button, once clicked, the widget will take all the visible space of the ConfigurableUI, SendBack is disabled in this state, however "Popout" is still available.
* Added a `previewState` and `previewDispatch` to the appui-layout-react `usePreviewFeatures` so it is now possible to save state for a preview feature without having to change the existing code (NinezoneState actually get saved to localstorage, so I did not want the preview feature to be saved there, that's the main reason, the second is to centralize so we dont have to create a new context for each new feature)

### Actions
* Removed "PreviewFeatures" type from the redux actions and replaced with a more general one because of the way the type was being computed, the *beta* tag was lost and cause extract-api to complain. Added a note to the action to directly use the properly typed public method instead if possible.
* Removed "ThemeId" type from the redux actions and replaced with "string". The type computation of the string union was causing extract-api to change the order from time to time, causing more pain than anything else. I removed that and also mention in the action to use `UiFramework.setColorTheme` which is properly typed. 

## Testing

Updated tests, ran multiple time and attempted to break, there are still some issue with focus (which I think are edge cases for this experiment)

### DIFF
--- a/common/api/appui-layout-react.api.md
+++ b/common/api/appui-layout-react.api.md
@@ -1568,7 +1568,10 @@ export interface UsePanelTargetArgs {
 export const usePointerCaptor: <T extends HTMLElement>(onPointerDown?: ((args: PointerCaptorArgs, e: PointerCaptorEvent) => void) | undefined, onPointerMove?: ((args: PointerCaptorArgs, e: PointerCaptorEvent) => void) | undefined, onPointerUp?: ((e: PointerCaptorEvent) => void) | undefined) => (instance: T | null) => void;
 
 // @internal
-export const usePreviewFeatures: () => PreviewFeaturesState;
+export const usePreviewFeatures: () => KnownPreviewFeatures & {
+    previewState: PreviewFeatureState;
+    previewDispatch: React_2.Dispatch<PreviewActions>;
+};
 
 // @internal (undocumented)
 export const useResizeGrip: <T extends HTMLElement>() => [(instance: T | null) => void, boolean, boolean];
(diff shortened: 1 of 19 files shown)


ben-polinsky (approved):
LGTM

joehenry9498:
All the code looks good. The only thing I'm noticing is a slight graphical inconsistency. Now that there are three buttons in the widget toolbar you can tell that the send back button has different padding from the popout and maxSize button. This causes the toolbar to look uneven. I think all widget toolbar tools should be the same size so that this doesn't happen.
Screenshot 2023-11-14 101609
Screenshot 2023-11-14 101634
Screenshot 2023-11-14 101649

raplemie:
> All the code looks good. The only thing I'm noticing is a slight graphical inconsistency. Now that there are three buttons in the widget toolbar you can tell that the send back button has different padding from the popout and maxSize button. This causes the toolbar to look uneven. I think all widget toolbar tools should be the same size so that this doesn't happen. Screenshot 2023-11-14 101609 Screenshot 2023-11-14 101634 Screenshot 2023-11-14 101649

The `sendBack` icons are not designed to be square (I dont know why though), so I'll leave them like they are and we can evaluate these when we get into the proper implementation. (Also, if we do add more options to this, we will need to move to a dropdown button, we can redesign then too.)

joehenry9498 (approved):

PR #582, opened 2023-11-09
Hidden Widget does not ignore DefaultPosition and DefaultSize when set to floating

joehenry9498:
## Changes
- Whenever a widget has its default state as hidden and then later set to floating, it would float into the top left corner and ignore the defaultFloatingPosition and defaultFloatingSize of the widget. This will no longer be the case, it will float to the size and position of the default parameters, until the size is changed by the user. The problems were that the position was just not being passed in with this case and the userSized variable of the floating widget was not being set.

## Testing
I tested by creating a new widget in one of the frontstages, the widgets default state is hidden. I then later changed its state to floating with the click of a button. The recommended code to test this is in the linked issue.


ui/appui-layout-react/src/appui-layout-react/state/NineZoneStateReducer.ts
```diff
@@ -668,6 +668,11 @@ export function NineZoneStateReducer(
       const nzBounds = Rectangle.createFromSize(state.size);
       const containedBounds = preferredBounds.containIn(nzBounds);
 
+      const userSized =
+        tab.userSized ||
+        (tab.isFloatingWidgetResizable &&
+          /* istanbul ignore next */ !!tab.preferredFloatingWidgetSize);
```
raplemie, line 674: This gives the impression that if the widget is *not* resizable we will not take into account the `canFloat.defaultSize`, is that correct ? (It is even more important if the widget cannot be resized that the size given by the dev should be enforced)

joehenry9498: I don't believe so, it should take into account the canFloat.defaultSize. It is setting the userSized to true if the tab has it set as true or if the tab is resizable and is no longer its preferred size. The userSized is always set in our code like this as well. 
image

raplemie: Only issue I see is that if you "SendBack", then the position is lost and it resets without reusing the default position (or the last position), however it keeps the default size. But that is doing the same in master and do not appear to be related to this PR, should we file a different issue or you think this can be related to the area you already explored with this ?

https://github.com/iTwin/appui/assets/1904889/d6838930-ae6e-46cc-b417-ec318e442829

joehenry9498: That shouldn't be too bad to fix, I think I know what needs to be done. I'll just tack it onto this PR, it's in the same area I'm working on.

raplemie (approved):
LGTM!
Thanks!

GintV:
Can this be backported to 4.6.x ?

joehenry9498:
This fix will be included when we release 4.7 early next week @GintV

GintV:
@joehenry9498 If I remember correctly, this was working on 4.5.x, so it would be good to get 4.6.x patch for this in case 4.7 will bring more regressions since we plan to release in upcoming weeks.

joehenry9498:
@mergifyio backport release/4.6.x

PR #597, opened 2023-11-22
Add double-click support in TreeEvents

ViliusBa:
## Changes

This PR adds the ability to add custom functionality for double-clicking tree nodes and edits the logic of `onNodeClicked` to maintain the node in selection when it is double-clicked. 

This change is required for this [item]( https://github.com/orgs/iTwin/projects/20/views/1?pane=issue&itemId=38880551).

## Testing

Tests for all possible workflows for the new implementations written.


ui/components-react/src/components-react/tree/controlled/TreeEventDispatcher.ts
```diff
@@ -209,8 +209,19 @@ export class TreeEventDispatcher implements TreeActions {
...
+    const isNodeSelected = node && node.isSelected;
+
+    if (
+      event.detail === 2 &&
+      this._treeEvents.onNodeDoubleClick !== undefined
+    ) {
+      this._treeEvents.onNodeDoubleClick({ nodeId });
+
+      // trigger onNodeClicked as to not lose highlight of the node when double clicking.
+      !isNodeSelected && this._selectionManager.onNodeClicked(nodeId, event);
```
saskliutas, line 221: Why is this needed? Do we want to select nodes when they are double clicked? What about deselecting?

ViliusBa: The issue this solves is that double-clicking won't deselect an already selected node.

saskliutas: > _What about deselecting?_ What do you mean by this?

I was thinking if we should deselect when selected node is double clicked.

Now that I think, I guess it makes sense to select node after it is double clicked. Have you tested if it works correctly in all SelectionModes?

ViliusBa: > I was thinking if we should deselect when selected node is double clicked.

I would be voting for keeping the node selected. Because it keeps the element highlighted when a certain action is executed on double-click (for example an element which represents the tree node is zoomed). Also, I think it would be quite annoying to need to select the node again once a custom double-click action is executed.

Tested it in all SelectionModes, seems to be working as expected.

saskliutas (approved):

GerardasB (changes requested):

GerardasB (approved):

PR #598, opened 2023-11-22
Add demo iModel selector to storybook

GerardasB:
## Changes

This PR adds a demo iModel selector to the storybook.

## Testing

1. Run a storybook 
2. Open `useIsBackageOpen` story
3. Select Demo iModel


docs/storybook/src/AppUiStory.tsx
```diff
@@ -14,62 +15,138 @@ import {
...
-  }, [props]);
+
+    const cleanup = appInitializer.initialize(startup, shutdown);
+    return () => {
+      ignore = true;
+      setInitialized(false);
+      cleanup();
+    };
+  }, [props, demoIModel]);
   if (!initialized) return null;
```
raplemie, line 122: With that change, switching component/iModel can take a certain time, I suggest adding something so we know something is going on.
```suggestion
  if (!initialized) return <ProgressLinear indeterminate labels={["Getting things ready!"]} />;
```

(Will require the import though...)

raplemie (approved):
LGTM, suggestions only.

PR #602, opened 2023-11-24
Add a Toolbar based on iTwinUI button group

GerardasB:
## Changes

...

## Testing

...


GerardasB:
@mayank99 we are trying to add a surface below a button group to style the toolbar. However, the surface is either filling the parent (`width: 100%`)

Screenshot 2024-03-06 at 10 40 29

OR not growing after the overflow (`width: fit-content`)

Screenshot 2024-03-06 at 10 42 01

Would it make sense to enhance the `ButtonGroup` to provide a different container based on which the overflow would be calculated (or have it nested by default)?
Alternatively, could we avoid using `ButtonGroup.overflowButton` and implement custom overflow logic?

GerardasB:
@mayank99  we are trying to build this component to navigate the group toolbar items (nested menus):
Screenshot 2024-03-06 at 10 47 09

After clicking on `Group 2_1`:
Screenshot 2024-03-06 at 10 50 59

It would need to support multiple columns as well:

Screenshot 2024-03-06 at 10 53 20

Would it make sense to use `DropdownMenu` or we'd have to build it from basic iTwinUI components?
The only issues I've noticed so far are related to keyboard navigation:
- After changing the group (click on a group item OR back button) the auto-focus is not kicking in
- It would probably make sense to support `ArrowLeft`/`ArrowRight` keyboard navigation in multi-column mode

mayank99:
> we are trying to add a surface below a button group to style the toolbar. However, the surface is either filling the parent (`width: 100%`)
> 
> OR not growing after the overflow (`width: fit-content`)

This makes sense because the ButtonGroup itself is normally "invisible", i.e. it does not have any background or border (unlike Surface). With `100%`, it always fills the parent container, and then the buttons inside it will leave some space left over for overflow. With `fit-content` it never has the chance to grow, because it only looks at the "current" items. (In iTwinUI, we only use 100% for overflow)

image

> Would it make sense to enhance the `ButtonGroup` to provide a different container based on which the overflow would be calculated (or have it nested by default)?

Depending on what you have in mind, you could contribute it to iTwinUI. The overflow logic was done a long time ago so probably needs improvement anyway.

> Alternatively, could we avoid using `ButtonGroup.overflowButton` and implement custom overflow logic?

From your screenshot it looks like you don't even need ButtonGroup because you have completely different visuals. So you could just put buttons inside a Surface if you want.

---

> @mayank99 we are trying to build this component to navigate the group toolbar items (nested menus)

This looks like a separate task that you probably shouldn't tackle in the same PR as the base toolbar?

We have had an open feature request for nested menus for a while. There was an attempt to work on it, but it's quite difficult to get right, so it was abandoned. https://github.com/iTwin/iTwinUI/issues/653

> It would need to support multiple columns as well:

Obviously, this isn't supported in our DropdownMenu currently. 

> Would it make sense to use `DropdownMenu` or we'd have to build it from basic iTwinUI components? The only issues I've noticed so far are related to keyboard navigation:
> 
> * After changing the group (click on a group item OR back button) the auto-focus is not kicking in
> * It would probably make sense to support `ArrowLeft`/`ArrowRight` keyboard navigation in multi-column mode

I don't think it makes sense to shoehorn DropdownMenu into this scenario. Is your component even a menu? Generally speaking, a menu should consist of clear actionable items.

image

Even then, you might want to start with a simpler implementation where you don't implement arrow key navigation. Make it a list of regular buttons.

```jsx
<div role="list"> {/* or <ul> */}
  <div role="listitem"> {/* or <li> */}
    <button>Item 1</button>
  </div>
  <div role="listitem">
    <button>Item 2</button>
  </div>
  <div role="listitem">
    <button>Group 3</button>
    <div role="list"> {/* submenu */}
      <div role="listitem">
        <button>Item 3.1</button>
      </div>
    </div>
  </div>
</div>
```

The only thing you'll need to do after this is to move focus when moving back and forth between different menu levels. The regular <kbd>Tab</kbd> key navigation will allow the user to focus buttons within a list.

Also you can use our [Popover](https://itwinui.bentley.com/docs/popover), which will take care of <kbd>Esc</kbd> key handling and outside clicks.

GerardasB:
> > we are trying to add a surface below a button group to style the toolbar. However, the surface is either filling the parent (`width: 100%`)
> > OR not growing after the overflow (`width: fit-content`)
> 
> This makes sense because the ButtonGroup itself is normally "invisible", i.e. it does not have any background or border (unlike Surface). With `100%`, it always fills the parent container, and then the buttons inside it will leave some space left over for overflow. With `fit-content` it never has the chance to grow, because it only looks at the "current" items. (In iTwinUI, we only use 100% for overflow)
> 
> image
> 
> > Would it make sense to enhance the `ButtonGroup` to provide a different container based on which the overflow would be calculated (or have it nested by default)?
> 
> Depending on what you have in mind, you could contribute it to iTwinUI. The overflow logic was done a long time ago so probably needs improvement anyway.
> 
> > Alternatively, could we avoid using `ButtonGroup.overflowButton` and implement custom overflow logic?
> 
> From your screenshot it looks like you don't even need ButtonGroup because you have completely different visuals. So you could just put buttons inside a Surface if you want.
> 
> > @mayank99 we are trying to build this component to navigate the group toolbar items (nested menus)
> 
> This looks like a separate task that you probably shouldn't tackle in the same PR as the base toolbar?
> 
> We have had an open feature request for nested menus for a while. There was an attempt to work on it, but it's quite difficult to get right, so it was abandoned. [iTwin/iTwinUI#653](https://github.com/iTwin/iTwinUI/issues/653)
> 
> > It would need to support multiple columns as well:
> 
> Obviously, this isn't supported in our DropdownMenu currently.
> 
> > Would it make sense to use `DropdownMenu` or we'd have to build it from basic iTwinUI components? The only issues I've noticed so far are related to keyboard navigation:
> > 
> > * After changing the group (click on a group item OR back button) the auto-focus is not kicking in
> > * It would probably make sense to support `ArrowLeft`/`ArrowRight` keyboard navigation in multi-column mode
> 
> I don't think it makes sense to shoehorn DropdownMenu into this scenario. Is your component even a menu? Generally speaking, a menu should consist of clear actionable items.
> 
> image
> 
> Even then, you might want to start with a simpler implementation where you don't implement arrow key navigation. Make it a list of regular buttons.
> 
> ```js
> <div role="list"> {/* or <ul> */}
>   <div role="listitem"> {/* or <li> */}
>     <button>Item 1</button>
>   </div>
>   <div role="listitem">
>     <button>Item 2</button>
>   </div>
>   <div role="listitem">
>     <button>Group 3</button>
>     <div role="list"> {/* submenu */}
>       <div role="listitem">
>         <button>Item 3.1</button>
>       </div>
>     </div>
>   </div>
> </div>
> ```
> 
> The only thing you'll need to do after this is to move focus when moving back and forth between different menu levels. The regular Tab key navigation will allow the user to focus buttons within a list.
> 
> Also you can use our [Popover](https://itwinui.bentley.com/docs/popover), which will take care of Esc key handling and outside clicks.

Thanks, that's very helpful. I've noticed some quirks in the `iTwinUI` overflow logic, because even if the overflow is disabled, the layout hooks still run and the button group is rendered w/ fewer items in between frames (that caused some issues in my custom overflow logic). I'll try to remove the button group, as suggested and replace the drop-down menu with a custom component.

ignas-k (changes requested):

ui/appui-react/src/appui-react/preview/new-toolbars/useOverflow.tsx
```diff
@@ -0,0 +1,124 @@
...
+  TComponent extends HTMLElement
+>(
+  allItems: ReadonlyArray<string>,
+  orientation: Orientation,
+  getItemSize: (item: string) => SizeProps,
+  getOverflowSize: () => SizeProps
+) => {
+  const [visible, setVisible] = React.useState(allItems.length);
+  const [calculate, setCalculate] = React.useState(false);
+  const componentRef = React.useRef<TComponent>();
```
saskliutas, line 27: Looks like this ref is unused. Only used in `useRefs` call.

GerardasB: Good catch, removed.

docs/storybook/.storybook/preview.ts
```diff
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 import type { Preview } from "@storybook/react";
 import { withDemoIModel, demoIModelGlobalType } from "./addons/DemoIModel";
+import { resizerGlobalType, withResizer } from "./addons/Resizer";
```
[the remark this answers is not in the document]
GerardasB: Good catch, removed.

ignas-k (approved):

PR #605, opened 2023-11-27
Fix incorrect multiline value renderer

ViliusBa:
## Changes

There was an issue with `MultilineTextPropertyRenderer` failing to render anything other than `PrimitiveStringProperties`. See attached item for more info.

The fix was to use the renderer that is appropriate for the record instead of predetermining the type without checking the record itself.

This PR also includes a little refactoring by moving a method that is used by many into a common file to reduce redundancy.

## Testing

Extended tests to check `MultilineTextPropertyRenderer` behavior with more `PropertyRecord` types.


raplemie (approved):
LGTM, can you add a comment with a link to the issue, I dont see any

saskliutas (approved):

PR #614, opened 2023-12-02
Dialog codemod

AntanasBuk:
## Changes

Codemod to accommodate changes required for `Dialog` deprecation and moving to iTwinUI `Dialog`. This does not include codemods for similar `Dialog` JSXElements like `ModelessDialog` or `GlobalDialog`.
Also includes structural changes to codemod tool as well as extended functionality.
 
## Testing

Extensive tests to confirm codemod behavior


ui/core-react/src/core-react/dialog/Dialog.tsx
```diff
@@ -391,3 +391,76 @@ export class Dialog extends React.Component<DialogProps, DialogState> {
     }
   };
 }
+
+/**
+ * A temporary function for converting DialogButtonDef items to Button elements
+ * @public
```
raplemie, line 397: Not sure we want to make this public

common/config/rush/pnpm-lock.yaml
raplemie, line 1: We'll need to revert the changes on this file

tools/codemod/src/utils/Extensions.ts
```diff
@@ -2,33 +2,134 @@
...
   toExpressionName,
   useCallExpression,
 } from "./CallExpression";
-import { useImportDeclaration } from "./ImportDeclaration";
 import {
-  findRootIdentifiers,
-  sortSpecifiers,
-  useImportSpecifier,
-} from "./ImportSpecifier";
+  ImportDeclarationCollection,
```
raplemie, line 14: Unused

tools/codemod/src/utils/Extensions.ts
```diff
@@ -2,33 +2,134 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-import type { ASTNode, Collection, JSCodeshift } from "jscodeshift";
+import type { ASTNode, ASTPath, Collection, JSCodeshift } from "jscodeshift";
+import type { ASTNode as ASTNodeType, Type } from "ast-types/lib/types";
+import type { NodePath } from "ast-types/lib/node-path";
```
raplemie, line 7: Implementation commented out, not used.

tools/codemod/src/utils/Extensions.ts
```diff
@@ -2,33 +2,134 @@
...
+              const childNodePathName =
+                typeof childPath.name === "number"
+                  ? childPath.parentPath.name
+                  : childPath.name;
+              return nodePathName === childNodePathName;
+            }
+            return true;
+          })
+          .paths()
+      );*/
```
raplemie, line 113: Is it better ? Is it a reference ? Should this be used or the comment removed ?

AntanasBuk: The commented out code is theoretically slower since it finds all children of children of children... and then filters only the first child layer. The code above is a tweaked implementation of jscodeshift find method so it would stop after finding only the first layer of children (immediate children). So it is up to you to decide which one to keep (above one is faster, commented one is more readable)

GerardasB:
Will not be merged until we work on Dialog deprecation. Changes are available in a branch https://github.com/iTwin/appui/tree/dialog-codemod

PR #616, opened 2023-12-04
Delay viewport creation to non 0x0 div.

### LINKED ISSUE #608: Properly manage zero width/height for viewport components
For civil, we use ViewportComponent and FloatingViewportContent react components. We have our homebrew solution to make sure those components are not rendered unless they have a valid width or height (>0) because it throws an error in core's `ScreenViewport.create`.

 It would be nice if appui could enforce positive viewport width/height, or replace by a simple <div> when it's null.

raplemie:
## Changes

Include a `ResizeObserver` to validate that the `div` targeted for iTwin.js `viewport` actually have non 0 height and width (otherwise it crashes anyway and the fix was needed by user devs)
`viewportRef` will not be set until the `div` is large enough, and resize observer is completely removed after initial render, to have same result as before.

## Testing

Updated tests to have a size, and added a tests for 0x0 sizes.

## Issue
Resolves #608

### DIFF
--- a/common/changes/@itwin/appui-react/raplemie-0x0ViewportHandling_2023-12-04-21-46.json
+++ b/common/changes/@itwin/appui-react/raplemie-0x0ViewportHandling_2023-12-04-21-46.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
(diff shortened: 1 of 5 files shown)


GerardasB (approved):

raplemie:

PR #619, opened 2023-12-06
Change behavior of property setting in Filter State

ViliusBa:
## Changes

Change the behavior of property setting for rule item to reset operator on change.

## Testing

Added a test case where property is changed for the same rule item.


saskliutas (approved):

raplemie:

raplemie:
We dont have an issue in AppUI for this, but the fix was included in 4.7.1 today.

PR #620, opened 2023-12-07
Fix constructors of data filterers

ViliusBa:
## Changes

Change the constructor implementation of data filterers to automatically convert filterText to lowercase, as it prevents bugs from happening if user would forget to manually set the filterer text after constructing.

## Testing

Tested all filterers with a case where filterText isn't set with the setter, but rather with the constructor.


saskliutas (approved):

jjbeckman13:
We just need this in a patch release of 4.7.x. We will pick this up in our next release.

raplemie:

PR #621, opened 2023-12-07
Correctly dock closed popout widget to panel section

### LINKED ISSUE #202: canPopout: if frontstage is switched while popout is open, it does not return to original position
**Describe the bug**
If a widget is popped out into a separate browser window. That window must be closed before the frontstage is closed in order for it to retain it's original position.

**To Reproduce**

1. Set the "canPopout" property of a widget in a frontstage.
2. Select the frontstage, Set the widget in a docked position at the bottom of the window
3. Pop out the widget into a separate window.
4. **Without closing the separate window** switch to a different frontstage
5. Switch back to frontstage with popout widget.
6. Widget appears floating in a tiny window in the main view.

**Expected behavior**
Widget should return to the docked position in this instance rather than defaulting to a floating position.

**Screenshots**
Widget is docked, popout window is pressed
image

Frontstage is switched, then switched back, widget is not docked:
image

- OS: Windows
- Browser: Chrome
- Version: 110.0.5481.180 (Official Build) (64-bit)
- iTwin.js Version: 3.6.0

GerardasB:
## Changes

Fixes #202. Correctly docks a closed popout widget to the panel section from which the widget was popped out.

## Testing

Added additional e2e test.


ui/appui-layout-react/src/appui-layout-react/state/NineZoneStateReducer.ts
```diff
@@ -366,37 +371,60 @@ export function NineZoneStateReducer(
...
           tabs,
         });
-      } else {
-        // if widget panel section was removed because it was empty insert it
-        const sectionIndex = widgetPanelSectionId.endsWith("End") ? 1 : 0;
-        state = insertPanelWidget(
+      }
+
+      // Insert a new panel section.
+      if (isPanelWidgetRestoreState(home)) {
```
raplemie, line 407: There is a bug here where we try to use the `widgetId` that we already validated did not exist instead of the correct `getWidgetPanelSectionId` that was saved and should be created, and this cause the poped out widget to always go back to `start` section if the `end` section do not exist. (It works well if there is already a widget in the end section though.)

raplemie (changes requested):
Excellent improvement !

ui/appui-layout-react/src/appui-layout-react/state/NineZoneStateReducer.ts
GerardasB: Fixed by removing the substring check that was copied from appui-react.

raplemie (approved):
LGTM!

PR #622, opened 2023-12-07
Fix popouts opening and saving as incorrect size

joehenry9498:
## Changes
There was a bug where each time you opened and closed a popout widget it would get smaller or larger than you had left it. This was due to how window.open() works with browser zoom. We were expecting the child window to open to certain specified dimensions but the browsers would open them to varying dimensions depending on zoom and browser type. This snowballed in our code to cause the graphical issues shown in the bug. Being that this is a long time issue with window.open, we had to find a way to work around it on our end. So the expected size of the popout is saved when opened so we now how much to offset the final save of the dimensions by. More changes may be needed later if users are using different browsers then mentioned below and issues are noticed.

## Testing
Test popout widget stays same size as expected when opened and closed on Chrome, Edge, Firefox, and Electron.

### DIFF
--- a/ui/appui-react/src/appui-react/childwindow/InternalChildWindowManager.tsx
+++ b/ui/appui-react/src/appui-react/childwindow/InternalChildWindowManager.tsx
@@ -25,6 +25,7 @@ import type {
   OpenChildWindowInfo,
 } from "../framework/FrameworkChildWindows";
 import { createLayoutStore, NineZone } from "@itwin/appui-layout-react";
+import type { ChildWindow } from "./ChildWindowConfig";
 
 const childHtml = `<!DOCTYPE html>
 <html>
@@ -141,12 +142,25 @@ export class InternalChildWindowManager implements FrameworkChildWindows {
 
   // istanbul ignore next: Used in `open` which is not tested.
   private renderChildWindowContents(
-    childWindow: Window,
+    childWindow: ChildWindow,
     childWindowId: string,
     content: React.ReactNode,
     title: string
   ) {
     childWindow.document.title = title;
+    if (childWindow.expectedHeight && childWindow.expectedWidth) {
+      childWindow.deltaWidth =
+        childWindow.expectedWidth -
+        (childWindow.currentBrowser === "chromium based edge"
+          ? childWindow.outerWidth
+          : childWindow.innerWidth);
+      childWindow.deltaHeight =
+        childWindow.expectedHeight -
+        (childWindow.currentBrowser === "chromium based edge"
+          ? childWindow.outerHeight
+          : childWindow.innerHeight);
+    }
+
     const reactConnectionDiv = childWindow.document.getElementById("root");
     if (reactConnectionDiv) {
       // set openChildWindows now so components can use it when they mount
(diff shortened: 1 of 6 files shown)


ui/appui-react/src/appui-react/childwindow/InternalChildWindowManager.tsx
```diff
@@ -257,6 +271,25 @@ export class InternalChildWindowManager implements FrameworkChildWindows {
     return outLocation;
   }
 
+  // This is needed due to the differences in how browsers calculate the
+  // inner vs outer width of windows. Calculations are different between
+  // browsers. We are starting with some of the mainly used browsers but can
+  // add more if issues arise with others.
```
raplemie, line 277: Can you validate how safari handles this? I suppose Mac users default browser would be safari?

ui/appui-react/src/appui-react/childwindow/InternalChildWindowManager.tsx
```diff
@@ -257,6 +271,25 @@ export class InternalChildWindowManager implements FrameworkChildWindows {
...
+  private findBrowser(agent: string): string {
+    switch (true) {
+      case agent.indexOf("edge") > -1:
+        return "edge";
+      case agent.indexOf("firefox") > -1:
+        return "firefox";
+      case agent.indexOf("edg/") > -1:
+        return "chromium based edge";
```
raplemie, line 285: Since we use this for only 1 particular purpose and the only value that we appear to care is this one, should we reduce this to `const shouldUseOuterSize = agent.indexOf("edg/") > -1;` ? Seems that we are reducing the agent string, but with no clear reason at this point. Or do you foresee that we migth need completely different solutions depending on these browsers ?

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
```diff
@@ -913,11 +914,21 @@ export class FrontstageDef {
...
-    const adjustmentHeight = ProcessDetector.isElectronAppFrontend ? 39 : 0;
+    let width =
+      childWindow.currentBrowser === "chromium based edge"
+        ? childWindow.outerWidth
+        : childWindow.innerWidth;
+    let height =
+      childWindow.currentBrowser === "chromium based edge"
+        ? childWindow.outerHeight
+        : childWindow.innerHeight;
+    if (childWindow.deltaHeight && childWindow.deltaWidth) {
```
raplemie, line 925: Could their be only a delta in either ? (deltaHeight === 2, deltaWidth === 0)

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
```diff
@@ -913,11 +914,21 @@ export class FrontstageDef {
     const widgetDef = this.findWidgetDef(tabId);
     if (!widgetDef) return;
 
-    const adjustmentWidth = ProcessDetector.isElectronAppFrontend ? 16 : 0;
```
raplemie, line 916: Is that no longer needed, or lost while testing ?

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
joehenry9498: There were always be a delta width and delta height, even if it's just 0. All `ChildWindow` props are optional parameters because they aren't set when the window is initially created and there were errors about that. So I don't think changing it is necessary

ui/appui-react/src/appui-react/childwindow/InternalChildWindowManager.tsx
joehenry9498: Getting the repository set up on my Mac and trying it out

ui/appui-react/src/appui-react/frontstage/FrontstageDef.tsx
raplemie: With the current `if` condition, if either is `0` none will be applied. (`0` is falsy, and the `&&` makes it so both need to be truthy) Should be a `||`

ui/appui-react/src/appui-react/childwindow/InternalChildWindowManager.tsx
raplemie: I think for now we can remove this part, we'll still have the PR history to get back with this solution when we need it, if it happens that this is the solution we need then.

joehenry9498: Removed

raplemie (approved):
LGTM!

PR #642, opened 2023-12-20
Revert Viewport 0x0 handling

raplemie:
## Changes

Revert #616 as this is only half a solution and a full solution would require behavior change that should go in a Minor once we figure the correct details.

## Testing

Removed additional test, validated that everthing was as before.

## Issues
Reopens #608


raplemie:

joehenry9498 (approved):
Through both testing and code review, everything is working as expected

PR #643, opened 2023-12-20
Add eslint deprecation exceptions for appui-abstract@4.3 compatibility

ben-polinsky:
Add eslint-ignore, or use deprecation replacement where possible, for lint issues related to deprecation of recent appui-abstract functionality. Most of these result from the caution we're taking around `UiError` and `UiEvent` to ensure any type assertions are not disrupted.


GerardasB (dismissed):
Instead of ignoring could we throw a non-deprecated object and use suggested replacement APIs in other places?
EDIT: Throwing another object might break some consumer exception handling, however, we will have to do that at one point eventually (at least when `UiError` is eventually removed). Should we wait for a major to make the change? Is there a better strategy?

ben-polinsky:
> Instead of ignoring could we throw a non-deprecated object and use suggested replacement APIs in other places? EDIT: Throwing another object might break some consumer exception handling, however, we will have to do that at one point eventually (at least when `UiError` is eventually removed). Should we wait for a major to make the change? Is there a better strategy?

Missed this comment. Yeah, I thought about that. I think we have to wait for a major to make the change to be on the safe side. Do you think we can do it in the next major rather than `n+1` since it's not an "explicit" public API change?

GerardasB:
> > Instead of ignoring could we throw a non-deprecated object and use suggested replacement APIs in other places? EDIT: Throwing another object might break some consumer exception handling, however, we will have to do that at one point eventually (at least when `UiError` is eventually removed). Should we wait for a major to make the change? Is there a better strategy?
> 
> Missed this comment. Yeah, I thought about that. I think we have to wait for a major to make the change to be on the safe side. Do you think we can do it in the next major rather than `n+1` since it's not an "explicit" public API change?

Yes, I think we'll have to do that with the next major version.

common/changes/@itwin/appui-react/add-eslint-deprecation-exceptions_2023-12-20-18-38.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "ignore appui-abstract eslint deprecation m",
```
raplemie, line 5: The message is incomplete, but I dont think we need to put that in the changelogs anyway.
```suggestion
      "comment": "",
```

raplemie (approved):
I think that's fine for now, we'll handle the rest later. Thanks

PR #655, opened 2023-12-21
useThisWidget hook creation

joehenry9498:
## Changes
- Created a new hook called `useWidget` which you can use to get information about the widget such as its location and state. You can also set state with the hook. Added a new widget in the WidgetApi frontstage to test this hook

## Testing
- Test with new widget in WidgetApi frontstage. Ensure state and location is as expected and that set state buttons work as expected.


ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+ * the ability to set the state of the widget.
+ * @alpha
+ */
+export function useThisWidget(): ThisWidget {
+  const tabId = React.useContext(TabIdContext);
+  assert(!!tabId);
+
+  const frontstage = useActiveFrontstageDef();
+
+  const [widgetDef, setWidgetDef] = React.useState(() =>
```
GerardasB, line 35: Could we re-use `useWidgetDef`, we'll want to add `onUiProviderRegisteredEvent` event handling there?

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+  setState(state: Exclude<WidgetState, WidgetState.Floating>): void;
+  state: WidgetState;
+}
+
+/** Hook that returns information about the Widget in the current context.
+ * @returns ThisWidget interface that contains the WidgetLocation, WidgetState, and
+ * the ability to set the state of the widget.
+ * @alpha
+ */
+export function useThisWidget(): ThisWidget {
```
GerardasB, line 29: Will we need a hook to get information about other widgets? Should we instead consider something like:
`useWidget(widgetId? string): Widget;`, for now at least we could just go with `useWidget()`, but would probably need to come up with a better name for `ThisWidget`.

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  assert(!!widgetDef);
+
+  const [widgetLocation, setWidgetLocation] = React.useState<
+    "docked" | "floating" | "popout"
+  >(findLocation(tabId, true));
+
+  const [state, setState] = React.useState(widgetDef.state);
+
+  function findLocation(locationTabId: string, initial: boolean): any {
```
GerardasB, line 47: Let's refactor this: get rid of `initial` argument and explicit `any` return type.

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }, [frontstage, widgetDef]);
+
+  React.useEffect(() => {
+    return InternalFrontstageManager.onFrontstageNineZoneStateChangedEvent.addListener(
+      () => {
+        findLocation(tabId, false);
+      }
+    );
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [frontstage, tabId]);
```
GerardasB, line 87: We should move `findLocation` outside of component body (might require additional arguments). Should be enough to only keep `tabId` in the dependency list (+frontstage or +nineZoneState). Remove eslint disable.

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const [state, setState] = React.useState(widgetDef.state);
+
+  function findLocation(locationTabId: string, initial: boolean): any {
+    let tabLocation;
+    if (frontstage?.nineZoneState)
+      tabLocation = getTabLocation(frontstage.nineZoneState, locationTabId);
+
+    if (tabLocation) {
+      if ("floatingWidgetId" in tabLocation) {
+        if (initial) return "floating";
```
GerardasB, line 54: Just return value w/o `setState()`. In initializer just call this fn, in other places just do `setWidgetLocation(findLocation())`.

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  const [state, setState] = React.useState(widgetDef.state);
+
+  function findLocation(locationTabId: string, initial: boolean): any {
+    let tabLocation;
+    if (frontstage?.nineZoneState)
+      tabLocation = getTabLocation(frontstage.nineZoneState, locationTabId);
+
+    if (tabLocation) {
+      if ("floatingWidgetId" in tabLocation) {
```
GerardasB, line 53: Reuse type guards: `isFloatingTabLocation()`, `isPopoutTabLocation()`, `isPanelTabLocation()`.

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  const [widgetLocation, setWidgetLocation] = React.useState<
+    "docked" | "floating" | "popout"
+  >(findLocation(tabId, true));
+
+  const [state, setState] = React.useState(widgetDef.state);
+
+  function findLocation(locationTabId: string, initial: boolean): any {
+    let tabLocation;
+    if (frontstage?.nineZoneState)
```
GerardasB, line 49: Simplify nesting, by returning early instead.

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+import { useActiveFrontstageDef } from "../frontstage/FrontstageDef";
+import { InternalFrontstageManager } from "../frontstage/InternalFrontstageManager";
+import { UiItemsManager } from "../ui-items-provider/UiItemsManager";
+import type { WidgetState } from "./WidgetState";
+
+/** Interface to be used with the useThisWidget hook. Holds information
+ * about the Widget such as the state and location. Also contains a function to set WidgetState.
+ * @alpha
+ */
+export interface ThisWidget {
```
GerardasB, line 18: Will we need additional information? Do we plan to add that additional information as optional properties only? Or we could keep this in `alpha` for a while and still do that?

I was thinking about other strategies, i.e. could we consider returning an implicit type from the hook, add additional `@note` that the return type should never be used outside of this hook. But https://www.itwinjs.org/learning/api-support-policies/ doesn't seem to allow such cases.

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
```diff
@@ -0,0 +1,102 @@
...
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [frontstage, tabId]);
+
+  React.useEffect(() => {
+    return UiItemsManager.onUiProviderRegisteredEvent.addListener(() => {
+      setWidgetDef(frontstage?.findWidgetDef(tabId));
+    });
+  }, [frontstage, tabId]);
+
+  function setWidgetState(
```
GerardasB, line 95: Should we make this stable with `useCallback`?

common/api/appui-react.api.md
```diff
@@ -1865,7 +1865,7 @@ export interface FrameworkChildWindows {
     closeAll(): void;
     find(childWindowId: string | undefined): OpenChildWindowInfo | undefined;
     findId(contentWindow: Window | undefined | null): string | undefined;
-    open(childWindowId: string, title: string, content: React.ReactNode, location: ChildWindowLocationProps, useDefaultPopoutUrl?: boolean): boolean;
+    open(childWindowId: string, title: string, content: React.ReactNode, location: ChildWindowLocationProps, tabId: string, useDefaultPopoutUrl?: boolean): boolean;
```
GerardasB, line 1868: Not allowed, this is a breaking change.

GerardasB (changes requested):

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
raplemie: I'd rather not suggest giving access to other widget with this one, the goal being that this will be a tight link to yourself, if we add access to other widget, then the potential state solution we were thinking will be really awkward as any widget will be able to have an impact on any other widgets, hence the suggestion for "ThisWidget" (which we can change, but that was the idea). If needed, we can have a different hook which may give a different set of tools to other widget.

common/api/appui-react.api.md
joehenry9498: Removed

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: Raphael thought we should start with the included props. I don't think the optional properties will be an issue with this? Since the type is really only used to store information returned from this hook. No information is needed from the user. So if we add additional information the hook should not have a problem returning it, but then the user wouldn't need to use it if they didn't want to. I think we should keep it in alpha for now though

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: Ill rename it to useWidget()

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: implemented

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: Fixed

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: Moved it

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: Done

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: implemented. It already has that event

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: refactored

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: implemented

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: removed interface, renamed it to useWidget()

ui/appui-react/src/appui-react/widgets/useThisWidget.tsx
joehenry9498: Removed the interface and returning object now returns just data. Changed it to the recommendations of this comment in the iTwin team https://teams.microsoft.com/l/message/19:6582ace6c5754414bdab0a95d7420bc6@thread.skype/1703264604580?tenantId=067e9632-ea4c-4ed9-9e6d-e294956e284b&groupId=2fc3ba7e-a05a-4cc0-9411-e9d2d2652ec1&parentMessageId=1703260701203&teamName=iTwin.js&channelName=General&createdTime=1703264604580

ui/appui-react/src/appui-react/widgets/useWidget.tsx
```diff
@@ -0,0 +1,103 @@
...
+     * Set widget to different state
+     * @param WidgetState that you want to set the widget to.
+     */
+    setState: setWidgetState,
+  };
+}
+
+function findLocation(locationTabId: string, frontstage: FrontstageDef) {
+  let tabLocation;
+  if (frontstage?.nineZoneState)
```
GerardasB, line 85: Frontstage is always defined. I think it should be safe to assert on both `nineZoneState` and returned `tabLocation`. Otherwise, we incorrectly say that the widget is `docked`.

GerardasB (changes requested):

ui/appui-react/src/appui-react/widgets/useWidget.tsx
```diff
@@ -0,0 +1,103 @@
...
+     */ state,
+    /**
+     * Where the widget is located ("docked" | "popout" | "floating").
+     */
+    widgetLocation,
+    /**
+     * Set widget to different state
+     * @param WidgetState that you want to set the widget to.
+     */
+    setState: setWidgetState,
```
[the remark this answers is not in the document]
joehenry9498: It did not, but it also didn't complain about it. Changed it to your formatting, prettier also says that's fine.

test-apps/appui-test-app/appui-test-providers/src/ui/widgets/useWidgetHookWidget.tsx
```diff
@@ -0,0 +1,46 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
+import { useWidget, WidgetState } from "@itwin/appui-react";
+import { Button } from "@itwin/itwinui-react";
+
+export function UseThisWidgetHookWidget() {
+  const thisWidget = useWidget();
```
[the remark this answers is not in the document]
joehenry9498: Replaced all instances

ui/appui-react/src/appui-react/widgets/useWidget.tsx
joehenry9498: You can't assert on tabLocation, if you are currently dragging the widget the tabLocation is undefined. It will default to floating in this case. Asserted on the nineZoneState

ui/appui-react/src/appui-react/childwindow/InternalChildWindowManager.tsx
```diff
@@ -161,32 +162,35 @@ export class InternalChildWindowManager implements FrameworkChildWindows {
...
     const reactConnectionDiv = childWindow.document.getElementById("root");
-    if (reactConnectionDiv) {
+    if (reactConnectionDiv && content && React.isValidElement(content)) {
       // set openChildWindows now so components can use it when they mount
       this._openChildWindows.push({
         childWindowId,
         window: childWindow,
         parentWindow: window,
       });
+      const tabId = content.props.widgetDef.id as string;
```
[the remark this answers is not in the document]
raplemie: `open` is a public API on `UiFramework.childWindows`, `widgetDef` is always defined when it is a widget that is being popped out, but you could open anything in a child window. (Search for the "OpenPopupdialogTool.tsx" file in the appui-test-providers, and look in the Popout frontstage in the test app, you will have a couple of buttons that will break this logic as they are passed a PopoutDialog or a PopoutTestView component )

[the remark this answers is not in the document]
raplemie: However, in that case, the content is NOT a widget and the `TabIdContext` is not expected to work, so if `tabId` is undefined, you could probably branch and simply not render it.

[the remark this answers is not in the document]
joehenry9498: I understand. I added that conditional render of the tab provider so it should be good now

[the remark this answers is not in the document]
joehenry9498: Through testing it was breaking that dialog stuff before, it is good now though

GerardasB (approved):

PR #658, opened 2023-12-22
Tool settings changes

FlyersPh9:
## Changes

- Replace tool settings overflow button to use iTwinUI-React `<IconButton>` and iTwinUI-Icons `more.svg`.

## Testing

None yet.

### DIFF
--- a/common/config/rush/pnpm-lock.yaml
+++ b/common/config/rush/pnpm-lock.yaml
@@ -499,6 +499,7 @@ importers:
       '@itwin/core-react': workspace:*
       '@itwin/eslint-plugin': ^4.0.0-dev.45
       '@itwin/itwinui-icons-react': ^2.1.0
+      '@itwin/itwinui-react': ^2.11.11
       '@itwin/itwinui-variables': ^2.0.0
       '@testing-library/dom': ^8.11.2
       '@testing-library/react': ^12.0.0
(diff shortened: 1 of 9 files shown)


ui/appui-react/src/appui-react/layout/tool-settings/Handle.scss
```diff
@@ -4,20 +4,14 @@
  *--------------------------------------------------------------------------------------------*/
 
 .nz-toolSettings-handle {
```
GerardasB, line 6: Should we replace this Handle component with iTwinUI borderless button?

GerardasB, line 6: The drag handle icon is super tiny is that expected?

GerardasB (approved):
Would be great to start adding some before/after screenshots to the PR description for visual changes.

ui/appui-react/src/appui-react/layout/tool-settings/Handle.scss
FlyersPh9: I think if we used a button it would be expected that a click OR drag & drop would undock the tool settings. I'm thinking maybe we revisit this decision in a future PR where we try to address #727 .  What do you think?

FlyersPh9: @GerardasB I'll wait for you to check this change before merging.

GerardasB: Sounds good, let's do that.

PR #660, opened 2024-01-02
Virtualized Property Grid: display loader on all loads

ViliusBa:
## Changes

This PR changes the logic of `VirtualizedPropertyGridWithDataProvider` to display the loading spinner on all selections if delay threshold is met (250 ms currently). Which basically means that if a `getData` call takes longer than the 250 ms to complete a loading spinner will be shown instead of the grid until the call is completed.

## Testing

Added a test case which replicates the aforementioned situation.


ui/components-react/src/components-react/propertygrid/component/VirtualizedPropertyGridWithDataProvider.tsx
```diff
@@ -47,25 +47,35 @@ export interface VirtualizedPropertyGridWithDataProviderProps
...
+  const [showSpinner, setShowSpinner] = useState(false);
+
+  const { modelSource, inProgress } = useTrackedPropertyGridModelSource({
     dataProvider: props.dataProvider,
   });
+
   const model = usePropertyGridModel({ modelSource });
   const eventHandler = usePropertyGridEventHandler({ modelSource });
+  useEffect(() => {
+    const timeout = setTimeout(() => {
```
saskliutas, line 59: I think if `inProgress` is `false` we don't need a delay before updating `showSpinner` flag.

saskliutas (changes requested):

ui/components-react/src/components-react/propertygrid/component/VirtualizedPropertyGridWithDataProvider.tsx
```diff
@@ -47,25 +47,41 @@ export interface VirtualizedPropertyGridWithDataProviderProps
...
+  }, [inProgress, model]);
 
   return (
-    <VirtualizedPropertyGrid
-      {...props}
-      model={model}
-      eventHandler={eventHandler}
-    />
+    <>
+      {showSpinner || !model ? (
```
raplemie, line 74: Why dont we simply use `inProgress` here instead of adding a new variable? `DelaySpinner` already have a builtin delay which could be adjusted with the `delay` prop if it is not of the right time.

ViliusBa: Because we wanted to have a delay on deciding when to switch between `VirtualizedPropertyGrid` and `DelaySpinner` to avoid flickering between the components. Instead of always switching to the spinner component when `inProgress===true`, we only switch when inProgress is true for more than a certain delay. This makes our component less _flickery_ if the requests are fast to complete.

ui/components-react/src/test/common/DelayedSpinner.test.tsx
```diff
@@ -17,17 +16,13 @@ describe("<DelayedSpinner />", () => {
   });
 
   it("renders spinner with delay", async () => {
-    const clock = sinon.useFakeTimers({ now: Date.now() });
-    const delay = 100;
-    const { container } = render(<DelayedSpinner delay={delay} />);
+    const { container } = render(<DelayedSpinner delay={100} />);
     expect(container.children.length).to.be.eq(0);
```
ViliusBa, line 20: Using sinon.useFakeTimers changes behavior of other tests unpredictably and I couldn't find a way to properly dispose of it in a way that it wouldn't impact other tests. Therefore, I decided to get rid of it altogether. 

That being said, removing it in this test might make it flaky, so I am open to suggestions on how to remedy this

grigasp: I assume you tried this example: https://github.com/sinonjs/fake-timers?tab=readme-ov-file#faking-the-native-timers? Seems unlikely it's a problem with the library, based on the fact that it has 9M weekly downloads and no related issue in its repo.

ViliusBa: Changed the tests that use `sinon.useFakeTimers` to only initialize once per test suite and it seems like it fixed the issue.

ViliusBa: Should probably change it for other tests also, so that it wouldn't occur again...

ui/components-react/src/components-react/propertygrid/internal/PropertyGridHooks.ts
```diff
@@ -46,6 +46,7 @@ export function usePropertyData(props: {
  * Custom hook that creates a [[PropertyGridModelSource]] and subscribes it to data updates from the data provider.
  * @throws if/when `IPropertyDataProvider.getData()` promise is rejected. The error is thrown in the React's render loop, so it can be caught using an error boundary.
  * @public
+ * @deprecated in 4.9.0. Use useTrackedPropertyGridModelSource instead.
```
grigasp, line 49:
```suggestion
 * @deprecated in 4.9.0. Use `useTrackedPropertyGridModelSource` instead.
```

ui/components-react/src/components-react/propertygrid/component/VirtualizedPropertyGridWithDataProvider.tsx
raplemie: What is displayed by the Grid when the model is defined but the call are still `inProgress`, why is `model` even defined if we havent got the info we need, does it have any use ? It feels that this will cause much more flickering when the requests are not fast to complete. (but, maybe these are edgecases ?)

So, before this PR:
 - If there is a `model`, show the `VPGrid`. (what is actually displayed here if not ready ?)
 - If there is no `model`, show the `DelayedSpinner`, so an empty div.
   - After `500ms`, show `Spinner` in that div.

Now:
 - If there is no `model`, show the `DelayedSpinner`, so an empty div.
   - After `500ms`, show `Spinner` in that div.
 - If there is a `model`, show the `VPGrid`.
   - If `inProgress`, after `150ms`, remove `VPGrid` (what was the grid displaying for those 150ms ?) show the `DelayedSpinner`, so an empty div.
     - After `500ms`, show `Spinner` in that div.

ViliusBa: The implementation has been updated to use a custom renderer for a spinner so that we wouldn't have timeouts in different places. 

The workflow after this PR:
- If there is no model, show `ProgressRadial` immediately, no empty div. (Check for model being defined is only relevant for the first `getData` call as it is defined afterwards)
- If there is a defined model and `inProgress` is false, show the `VPGrid`
  - If `inProgress`, after `250ms` render 'ProgressRadial', so no empty div.

Hope this clears things up, let me know if I missed something.

raplemie: Yes, sorry I missed the change from `DelayedSpinner` to `ProgressRadial` within the `DelayedLoaderRenderer` component. 

> Could you expand on why you think so? (If you still think that is the case)

What confuses me is that while we are waiting for the loader to display, we display the grid, which, according to your answer above, may have content already as `model` may be defined. So the user will see data as we have a `model` but are still in progress, for a split second while `inProgress` is `true` at the same time that `model` is defined, then loose everything and that will be replaced by a spinner until `inProgress` turns `false`.

ViliusBa: Thanks for clarifying.

If I understood correctly you are talking about situations where the window between delay ending and the `getData` requests completing is rather small. Lets say delay of waiting before showing the loader is `250ms` and requests are completing within `251 to 350ms`, then yes, the component will look a bit flickery from switching between loader and `VPGrid` rapidly. 

However, I believe these are edge cases and I don't think there is a simple way to remedy it without removing this feature altogether.

raplemie (approved):
LGTM!

saskliutas (approved):

PR #664, opened 2024-01-05
Add visual tests and story for Editors

raplemie:
## Changes
In preparation for #576 which ended changing a lot of visuals, I added a tests for each base variant of editors (not all the interactions yet though) so we can more easily compare the changes.

We can add more elaborate interaction tests later.

## Testing

Yes... That's the basic point of this PR  :)


GerardasB (approved):
Great to see these in storybook as well!

PR #667, opened 2024-01-09
Allow `WidgetDef.show()` to work with popouts

joehenry9498:
## Changes
Changed widgets show() function to work with popouts as well.
This behavior ended up greatly depending on the browser used and is not guaranteed.
Just calling window.focus() worked for all browsers except Safari (but might differ depending on the user settings).
Safari has security measures that prevent you from changing focus to a different tab programmatically. Apple support docs to issue: https://discussions.apple.com/thread/251676767?sortBy=best . 

## Testing

- Test Chrome, Edge, Safari, Electron, and Firefox and evaluate the behavior. There is a widget in the WidgetApi frontstage that enables you to call show on any widget, which makes for easy testing.

I.e.:
- Start standalone test-app
- Open <http://localhost:3000/?frontstage=appui-test-providers:WidgetApi>
- Popout `WL-A` widget
- Open `Layout Controls` widget
- In the `Widget id` input select `WL-A`
- Click `Show` button
- Observe how the window of the popout widget is brought to front (if supported by the browser)


ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -541,10 +542,43 @@ export class WidgetDef {
...
+      const testWindow = UiFramework.childWindows.find(
+        tabLocation.popoutWidgetId
+      );
+      if (testWindow) {
+        // We have to check if Safari is the browser being used. Safari has security
+        // measures that do not allow you to change focus to another tab programmatically.
+        // Therefore the only and recommended way around this at the moment, is to close
+        // and then open the tab again. This remounts the popout so it is not ideal. Thats
+        // why this is not being done for all browsers.
+        // Apple support docs: https://discussions.apple.com/thread/251676767?sortBy=best
```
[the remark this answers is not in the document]
joehenry9498: It has been changed to a toast message with a do not show toggle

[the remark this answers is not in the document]
joehenry9498: Screen Shot 2024-01-16 at 1 46 33 PM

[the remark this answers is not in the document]
joehenry9498: The do not show again aspect is being finicky since it's in a toast, working on it now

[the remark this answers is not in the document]
joehenry9498: Should be no problems with it now

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -543,10 +547,82 @@ export class WidgetDef {
...
+                    );
+                  } else {
+                    window.localStorage.setItem(
+                      "hideSafariPopoutFocusMessage",
+                      "false"
+                    );
+                  }
+                }}
+              />
+              <label htmlFor="doNotShowAgain">Do not show again</label>
```
raplemie, line 598: Localize this string

ui/appui-react/public/locales/en/UiFramework.json
```diff
@@ -280,7 +280,8 @@
     },
     "errorMessage": {
       "unknownError": "Something went wrong...",
-      "widgetPopoutFail": "Widget failed to popout, ensure you allow popout for this site."
+      "widgetPopoutFail": "Widget failed to popout, ensure you allow popout for this site.",
+      "popoutSafariFocusFail": "Popout focus restricted by Safari"
```
raplemie, line 284: We should add the widget label in the message, so the user know what is happening.
```suggestion
      "popoutSafariFocusFail": "Popout widget '{{widgetLabel}}' attempted to get focus, but this action was restricted by the browser."
```

You can then fill this `{{widgetLabel}}` using a call similar to this:
```ts
UiFramework.localization.getLocalizedString(
  "widget.errorMessage.popoutSafariFocusFail",
  { widgetLabel: this.label, ns: UiFramework.localizationNamespace }
)
```

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -543,10 +547,82 @@ export class WidgetDef {
...
+                  ) {
+                    window.localStorage.setItem(
+                      "hideSafariPopoutFocusMessage",
+                      "true"
+                    );
+                  } else {
+                    window.localStorage.setItem(
+                      "hideSafariPopoutFocusMessage",
+                      "false"
+                    );
```
raplemie, line 594: Since setting this to `true` once will make the notification disappear forever, use `UiFramework.getUiStateStorage()` instead of `window.localStorage`, so if we need to change this later to a SettingsProvider and allow the user to change this preference, it will be easier.

raplemie (changes requested):

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
joehenry9498: Changed

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
joehenry9498: Changed to use UiStateStorage

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -537,16 +541,97 @@ export class WidgetDef {
    * I.e. opens the stage panel or brings the floating widget to front of the screen.
    * @public
    */
-  public show() {
+  public async show() {
```
raplemie, line 544: Having this change to a Promise<void> will cause users to have linter warnings that they should listen to the result, which they will not ever really care about. You should be able to refactor this to get the storage setting and then using the result directly with a `.then` or even `.finally` method, which would remove the requirement of the `async`.

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -537,16 +541,97 @@ export class WidgetDef {
...
+          navigator.userAgent.toLowerCase().indexOf("safari/") > -1;
+
+        const isChrome =
+          navigator.userAgent.toLowerCase().indexOf("chrome") > -1;
+
+        const hideFocusFail = await UiFramework.getUiStateStorage().getSetting(
+          "popoutFocus",
+          "hidePopoutFocusFail"
+        );
+        if (!isChrome && isSafari && hideFocusFail.setting !== true) {
```
raplemie, line 573: Getting the setting is only needed if the 2 first conditions are true, and if the setting is true we should probably simply do nothing, instead of trying the focus anyway, in case this starts to throw an error at some point.

raplemie (changes requested):

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
joehenry9498: Removed async and utilized .then

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
joehenry9498: Reworked to this logic

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -543,10 +551,93 @@ export class WidgetDef {
...
+        const isSafari =
+          navigator.userAgent.toLowerCase().indexOf("safari/") > -1;
+
+        const isChrome =
+          navigator.userAgent.toLowerCase().indexOf("chrome") > -1;
+
+        if (!isChrome && isSafari) {
+          void UiFramework.getUiStateStorage()
+            .getSetting("popoutFocus", "hidePopoutFocusFail")
+            .then((hideFocusFail: UiStateStorageResult) => {
```
GerardasB, line 576: Prefer using the `await` operator instead.

GerardasB (dismissed):

docs/changehistory/NextVersion.md
```diff
@@ -25,6 +25,8 @@ Table of contents:
 
   `@itwin/appui-layout-react` will no longer be published as a separate package.
 
+- Update widget `show()` function to work with popout widgets. #564
```
GerardasB, line 28: Sort by PR number.

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
raplemie: I asked for this change because await forces this public method to be async, which then forces every consumer to react on the method now returing a promise and probably getting lint error because of an await that will not return anything.

GerardasB: I see, I think this might get addressed while refactoring.

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -543,10 +551,93 @@ export class WidgetDef {
...
+        // because we don't want the state of the widget to be lost. Instead a toast will
+        // be displayed that informs the user that safari cannot focus the popout.
+        // Apple support docs: https://discussions.apple.com/thread/251676767?sortBy=best
+        const isSafari =
+          navigator.userAgent.toLowerCase().indexOf("safari/") > -1;
+
+        const isChrome =
+          navigator.userAgent.toLowerCase().indexOf("chrome") > -1;
+
+        if (!isChrome && isSafari) {
```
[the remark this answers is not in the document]
GerardasB: @FlyersPh9 I have decided to remove the toast message for now, since we are only checking one potential case of failure, but it seems like there could be multiple other cases and there's no good way to check if the `focus` had any effect.

saskliutas (approved):

ignas-k (approved):

PR #668, opened 2024-01-10
Fix panel resize issue

GerardasB:
## Changes

This PR fixes an issue that caused the panel to become unresizable due to a [transient update](https://github.com/pmndrs/zustand?tab=readme-ov-file#transient-updates-for-often-occurring-state-changes) handler added after the [PANEL_INITIALIZE](https://github.com/iTwin/appui/blob/e3e44fd0c8f276fefecc30baa0f454438eb6429b/ui/appui-layout-react/src/appui-layout-react/widget-panels/useAnimatePanel.tsx#L108) action was dispatched.

## Testing

Tested via standalone test-app with a single right panel in a frontstage.

### DIFF
--- a/common/changes/@itwin/appui-layout-react/fix-panel-resize_2024-01-10-17-56.json
+++ b/common/changes/@itwin/appui-layout-react/fix-panel-resize_2024-01-10-17-56.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-layout-react",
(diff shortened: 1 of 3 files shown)


GerardasB:
@Mergifyio backport release/4.8.x

joehenry9498 (approved):
Code changes look good. Single panel can be resized when no other action has been dispatched. Looks like there are a couple playwright tests that need fixed for the build to complete. Will approve

PR #670, opened 2024-01-11
Fix panel initialization

GerardasB:
## Changes

Add additional e2e tests and fix panel initialization quirks introduced with https://github.com/iTwin/appui/pull/636

## Testing

Added additional e2e tests


joehenry9498 (approved):
Not seeing anything out of the ordinary here, looks good

ui/appui-react/src/appui-react/widget-panels/Frontstage.tsx
```diff
@@ -553,10 +553,11 @@ export function initializePanel(
       maxSize,
     });
 
+  const collapsed = panelDef.defaultState !== StagePanelState.Open;
```
raplemie, line 556: What is a stagepanel in a "Popup" state (last enum value)? I dont see that used anywhere in the code ? (And, it would mean that when in this stage, the panel is collapsed, which would sound unexpected...) Should the enum value be marked as _deprecated_ ?

PR #676, opened 2024-01-15
Deprecate useDisposable

saskliutas:
## Changes

Deprecated `useDisposable` hook. It is not compatible with React 18 StrictMode and there is no way to make it work correctly without changing API. Updated `useOptionalDisposable` to work correctly. Made some changes to `useDisposable` hook to make it work in React 18 StrictMode as best as it can. Current implementation will leak some disposable resources in StrictMode but should work correctly when StrictMode is disabled.

Deprecated `useTreeEventsHandler` because it was based on `useDisposable` hook. Added new `useControlledTreeEventsHandler` hooks to replace it.

## Testing

Ran modified tests in Presentation repo that uses `useDisposable` hook https://github.com/iTwin/presentation/blob/master/packages/components/src/test/hooks/UseRulesetRegistration.test.ts


ui/components-react/src/components-react/tree/controlled/TreeHooks.ts
```diff
@@ -86,6 +86,7 @@ export function useTreeModelSource(dataProvider: TreeDataProvider) {
  * be used for that purpose based on whether the input is a factory function or params object.
  *
  * @public
+ * @deprecated in 4.9.0. This hooks does not work correctly in React 18. Use [[useControlledTreeEventsHandler]] instead.
```
grigasp, line 89:
```suggestion
 * @deprecated in 4.9.0. This hook does not work correctly in React 18 Strict mode. Use [[useControlledTreeEventsHandler]] instead.
```

ui/components-react/src/components-react/tree/controlled/TreeHooks.ts
```diff
@@ -94,5 +95,26 @@ export function useTreeEventsHandler<TEventsHandler extends TreeEventHandler>(
...
+/**
+ * Custom hook which creates and takes care of disposing a TreeEventsHandler. The input is either a factory method
+ * for a custom `TreeEventHandler` implementation or parameters for the default implementation.
+ *
+ * @note Caller must ensure `factoryOrParams` changes only when a new handler needs to be created. `useCallback` or `useMemo` can
+ * be used for that purpose based on whether the input is a factory function or params object.
+ *
+ * @returns `undefined` on first render and valid TreeEventsHandler on all subsequent renders.
+ * @public
+ */
```
grigasp, line 111:
```suggestion
/**
 * Custom hook which creates and takes care of disposing a [[TreeEventHandler]]. The input is either a factory method
 * for a custom [[TreeEventHandler]] implementation or parameters for the default implementation.
 *
 * @note Caller must ensure `factoryOrParams` changes only when a new handler needs to be created. `useCallback` or `useMemo` can
 * be used for that purpose based on whether the input is a factory function or params object.
 *
 * @returns `undefined` on first render and a valid [[TreeEventHandler]] on all subsequent renders.
 * @public
 */
```

ui/core-react/src/core-react/utils/hooks/useDisposable.ts
```diff
@@ -13,11 +13,31 @@ import type { IDisposable } from "@itwin/core-bentley";
  * A custom hook which creates a disposable object and manages its disposal on unmount
  * or factory method change.
  * @public
+ * @deprecated in 4.9.0. This hook is not compatible with React 18. Use [[useOptionalDisposable]] or `useState` + `useEffect` for
+ * creating and disposing disposable resources.
```
grigasp, line 17:
```suggestion
 * @deprecated in 4.9.0. This hook is not compatible with React 18 Strict mode. Use [[useOptionalDisposable]] or `useState` + `useEffect` for
 * creating and disposing disposable resources.
```

ui/core-react/src/core-react/utils/hooks/useDisposable.ts
```diff
@@ -13,11 +13,31 @@ import type { IDisposable } from "@itwin/core-bentley";
...
  * or factory method change.
  * @public
+ * @deprecated in 4.9.0. This hook is not compatible with React 18 Strict mode. Use [[useOptionalDisposable]] or `useState` + `useEffect` for
+ * creating and disposing disposable resources.
  */
 export function useDisposable<TDisposable extends IDisposable>(
   createDisposable: () => TDisposable
 ): TDisposable {
-  return useOptionalDisposable(createDisposable)!;
+  const [value, setValue] = useState(createDisposable());
```
raplemie, line 22: Why cant this keep using the new/fixed `useOptionalDisposable` ? If strict mode is running twice, why in this case are we trying to keep only one of the run instead of simply disposing of the first run and using the second, as we do in the optional one ?

saskliutas: It cannot use new `useOptionalDisposable` implementation because it would break API. `useDisposable` function definition states that it never returns `undefined` but `useOptionalDisposable` is always returning `undefined` on first render.

The problem with `useDisposable` hook is that it has side effect, it creates disposable when initializing state and if `useEffect` if never run that disposable is never disposed. Changes made to StrictMode in React 18 helps to find this side effect because components are rendered twice. As a result `useDisposable` in StrictMode always leaks one disposable resource:
1. Component is rendered and `useDisposable` is called, disposable resource is created.
1. That component is thrown away. `useEffect` was never run, resource is not disposed.
1. Component is rendered again and `useDisposable` is called, disposabled resource is created.
1.  `useEffect` is run. New resource is not created due to `useInitialValue` flag.
1.  `useEffect` is run again. Cleanup function disposes `prevValue` (value that was created initializing state). Effect callback is run, new resource is created.

New `useDisposable` implementation does not try to keep only one of the runs. It tries to make sure that disposable resource created during render loop initializing state is disposed. I think it could be refactored a little bit to make it easier to read:
```tsx
const [value, setValue] = useState(createDisposable());
const initialValue = useRef(value);
useEffect(() => {
  if (initialValue.current) {
    initialValue.current.dispos();
    initialValue.current = undefined;
  }
  const newValue = createDisposable();
  setValue(newValue);
  return () => {
   newValue.dispose();
  };
}, [createDisposable]);
```

raplemie (changes requested):

ui/core-react/src/core-react/utils/hooks/useDisposable.ts
saskliutas: Yes this part was missing, nice catch. 😞

raplemie (dismissed):
Can you add a note in NextVersion.md

saskliutas:
> Can you add a note in NextVersion.md

@raplemie Added note to NextVersion.md

GerardasB (approved):
I'd appreciate it if you'd add a PR number to changes in `NextVersion.md` while resolving the merge conflicts (i.e. https://github.com/iTwin/appui/pull/700)

PR #696, opened 2024-01-26
Fix popup positioning when an ascendant node is a positioned element

GerardasB:
## Changes

Fixes an issue where a popup is rendered with additional offset when an ascendant node is positioned and AppUI root node bounds don't match the viewport bounds by correctly positioning the popup relative to its container and making sure that the root container is a positioned element.

image

I've considered using `position: fixed` instead, but there are additional issues i.e. maintaining the offset when scrolling.

## Testing

Added `mode` URL parameter to the standalone test app.


GerardasB:
@Mergifyio backport release/4.8.x

PR #697, opened 2024-01-26
Remove Appui-layout-react package

### LINKED ISSUE #677: Fold `appui-layout-react` into `appui-react`, discontinue the former.
AppUI-react package currently use AppUi-layout-react package which only exposes _internal_ elements, and they have to be kept in sync so the proper values are available.

Fold the code in AppUI-React and discontinue publishing the separate package, removing a peer dependency management requirement.

raplemie:
## Changes

Completely removes everything related to appui-layout-react package and moves its content to appui-react/layout, everything in there is internal, so it has not been added to the barrel.

## Testing

Moved all tests from appui-layout-react to appui-react/test/layout as well, ensured that all tests passed (whch required `useFakeTimers` removal, and refactored an svg import which was causing problem since now the svg-loader component is "active" by default and caused issues with the snapshot system (svg code is now embeded instead of a separate file, akin the PinToggle)

All full-stack-ui tests works as well.

## Issue
resolves #677


GerardasB:
Prereq: https://github.com/iTwin/itwinjs-core/pull/6392

GerardasB (approved):

PR #702, opened 2024-01-30
FilterBuilder: Add Between operator

saskliutas:
## Changes

Added `Between` and `Not between` operators support to `PropertyFilterBuilder` components. This requires to make some breaking changes to internal components of `PropertyFilterBuilder (mainly renames) that are exposed for customization. These breaking changed should not have impact if `PropertyFilterBuilder` is used without heavy customization.

To avoid breaking `PropertyFilter` type range values entered for rule using `Between` operator are serialized into `PropertyValue`. Later when converting `PropertyFilterBuilderRule` into `PropertyFilter` rule using `Between` operator is split into two rules with `greater-or-equal` and `less-or-equal` operator. This approach allows to keep `PropertyFilter` type unchanged and does not introduce additional changes to consumers in order to handle `Between` operator.

## Testing

Added new unit tests and visually checked component in Storybook.


ui/components-react/src/test/filter-builder/FilterBuilderState.test.ts
```diff
@@ -684,14 +692,160 @@ describe("usePropertyFilterBuilder", () => {
...
+          );
+          const { result } = renderHook(() =>
+            usePropertyFilterBuilder({
+              initialFilter,
+            })
+          );
+          const { buildFilter } = result.current;
+
+          const buildFilterResult = buildFilter();
+          expect(buildFilterResult).to.not.be.eq(initialFilter);
```
grigasp, line 814: why are we expecting them to **not** match? maybe the check should be more specific about what the result should be?

ui/components-react/src/test/filter-builder/FilterBuilderState.test.ts
```diff
@@ -726,6 +880,212 @@ describe("usePropertyFilterBuilder", () => {
     });
   });
 
+  describe("range rule", () => {
+    it("converts `>=` and `<=` rules into `Between` rule", () => {
+      const { result } = renderHook(() =>
+        usePropertyFilterBuilder({
+          initialFilter: {
+            operator: PropertyFilterRuleGroupOperator.And,
```
grigasp, line 888: maybe `PropertyFilterBuilderRuleGroup.operator` should also be changed to accept strings like `PropertyFilterBuilderRule.operator`?

docs/changehistory/NextVersion.md
```diff
@@ -35,10 +36,24 @@ Table of contents:
 - `PropertyFilterBuilderToolbar` component to display the action buttons in the filter builder. #686
 - `isGroupOperatorDisabled` property of `PropertyFilterBuilderRendererProps` to control whether the group operator is toggle-able. #686
 
+### FilterBuilder changes
```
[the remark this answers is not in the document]
GerardasB: Make sure the PR number of your change is referenced.
Follow the `NextVersion.template.md`: instead of a new section, just indent your list if needed (I'd rather not mix 686 in there):
```
- ... #686
- Added support ... #702
  - Changed ...
  - Updated ...
```

ui/components-react/src/components-react/filter-builder/Operators.ts
```diff
@@ -20,49 +19,53 @@ export enum PropertyFilterRuleGroupOperator {
 }
 
 /**
- * Operators for comparing property value in rule.
+ * Operators for comparing property value in [[PropertyFilterRule]].
  * @beta
  */
 export enum PropertyFilterRuleOperator {
```
GerardasB, line 25: Any reason why this is an enum while `PropertyFilterBuilderRuleOperator` is a string literal union type? Should we stay consistent and use either enums + enum union OR literal union types?

docs/changehistory/NextVersion.md
```diff
@@ -35,10 +36,24 @@ Table of contents:
...
+
+- Updated UI of the `FilterBuilder` component. #686
+- Updated `PropertyFilterRule.operator` property type to support setting it using `PropertyFilterRuleOperator` enum or string value.
+- Changed `PropertyFilterBuilderRuleValue` component props to `PropertyFilterBuilderRuleValueRendererProps` in order to have access to rule operator when rendering value input.
+- Changed `PropertyFilterBuilderRule.operator` type from `PropertyFilterRuleOperator` to `PropertyFilterBuilderRuleOperator`.
+- Renamed `getPropertyFilterOperators` to `getPropertyFilterBuilderOperators`.
+- Renamed `getPropertyOperatorLabel` to `getPropertyFilterBuilderOperatorLabel`.
+- Renamed `PropertyFilterBuilderRuleOperatorProps` to `PropertyFilterBuilderRuleOperatorRendererProps`.
+- Added `PropertyFilterBuilderRuleOperator` type that has `between` and `not-between` value in addition to `PropertyFilterRuleOperator` values.
```
GerardasB, line 50: Seems like these should go under the `Additions` section.

ui/components-react/src/components-react/filter-builder/Operators.ts
saskliutas: I want to use literal union but that breaks `presentation-components` as it uses `PropertyFilterRuleOperator` enum. I was hoping to switch `PropertyFilterRuleOperator` into literal union in next major AppUI release.

docs/changehistory/NextVersion.md
saskliutas: I wanted to keep all changes related to addition of `Between` operator in one group. But I can move them under different sections if you want.

ui/components-react/src/components-react/filter-builder/Operators.ts
GerardasB: Since you own the `presentation-components` could you just react to the change? Since we are making all these other breaking changes, i.e. some of the renames seem not to be necessary.

docs/changehistory/NextVersion.md
GerardasB: Please do, otherwise, we'll end up with all the changes under the `Changes` section.

ui/components-react/src/components-react/filter-builder/Operators.ts
saskliutas: I could change enum to literal union and react to these changes but `presentation-components` have peer dependency on `@itwin/components-react` specified as `^4.0.0` so removing enum in minor bump would mean that all previous `presentation-components` versions would break when used with new versions of `components-react`. Current approach keeps older versions working and allows to start using string literals instead of enum in new versions.

GerardasB: We can keep it as is if you think it'll be too destructive, just a slight inconsistency :)

Just for argument's sake: this could be a good argument to NOT make these breaking changes to these beta APIs.
I.e. `PropertyFilterRuleGroupOperator` enum values were changed. What if your consumers used a literal number instead of an enum (even though it is unlikely)
```
function presentationFn(x: PropertyFilterRuleGroupOperator);
presentationFn(x.And); // continues to work
presentationFx(0); // OUCH after update
```

saskliutas: Yes, that example would break but I have hope that raw numbers weren't used . It was my mistake to use enum from the start, so I am trying to fix that before promoting API to public. Removing enum in minor version proved to be too destructive so I hoped to move gradually: allow using enum or string literal and in major release leave only literal union.

docs/changehistory/NextVersion.md
saskliutas: Moved in https://github.com/iTwin/appui/pull/702/commits/f56b07d9bf70e668418ffe577c9bda2524bfd61d

grigasp (approved):

GerardasB (approved):

PR #704, opened 2024-01-31
Add story for split viewport

joehenry9498:
## Changes
- Added a split viewport story to the storybook. The tools in toolbars change based on which view is active

## Testing
- Check out the storybook

### DIFF
--- a/docs/storybook/src/frontstage/SplitViewport.stories.tsx
+++ b/docs/storybook/src/frontstage/SplitViewport.stories.tsx
@@ -4,6 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 import type { Meta, StoryObj } from "@storybook/react";
 import {
+  ConditionalBooleanValue,
   ConditionalStringValue,
   StandardContentLayouts,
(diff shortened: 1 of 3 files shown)


docs/storybook/src/frontstage/SplitViewport.stories.tsx
```diff
@@ -0,0 +1,114 @@
...
+            new ConditionalIconItem(testIcon2, [
+              SyncUiEventId.ActiveContentChanged,
+            ]),
+            new ConditionalStringValue(
+              () => (leftViewportActive ? "Test 3" : "Test 4"),
+              [SyncUiEventId.ActiveContentChanged]
+            ),
+            () => undefined
+          ),
+        ],
```
raplemie, line 110: Can you add a 3rd items that show the `isHidden` (which you can set with the `overrides` param of `createActionItem`) and add a `isDisabled` to the second one, so the example showcase even more info ?

GerardasB (approved):
LGTM!

PR #705, opened 2024-01-31
`horizontalPanelAlignment` preview feature

raplemie:
## Changes

Adding the `horizontalPanelAlignment` preview feature, which allow both top and bottom panel to be either justified (default behavior), centered(between left and right), left or right.

Also changed the different button implementation to iTwinUI small borderless buttons (pin, popout, dock, sendback...) The button implementation is NOT a preview itself. It does NOT use button group yet.

## Testing

Updated snaphots for tests, validated.


ui/appui-react/src/appui-react/layout/widget-panels/Panel.tsx
```diff
@@ -59,6 +60,19 @@ export function WidgetPanel() {
...
+  const previewHorizontalPanelAlignAttributes = React.useMemo(
+    () =>
+      previewHorizontalPanelAlignmentEnabled
+        ? {
+            "data-preview-horizontal-panel-align-top": alignments.top,
+            "data-preview-horizontal-panel-align-bottom": alignments.bottom,
+          }
+        : {},
+    [alignments.bottom, alignments.top, previewHorizontalPanelAlignmentEnabled]
+  );
```
raplemie, line 75: This feature is introduced as a parallel implementation to the existing "span" feature for the top and bottom panel because "span" is saved to the local storage state, and if the preview feature is disabled, it will cause the last state to stick and not be changeable. (And the current implementation only supports `true`/`false`, for `justify`/`center`, so I did not want to change this part, it was simpler/safer to build it in a parallel system)

ui/appui-react/src/appui-react/layout/widget/SendBack.scss
GerardasB, line 1: Wow, that's a good riddance :)

ui/appui-react/src/appui-react/layout/widget/PreviewHorizontalPanelAlign.tsx
```diff
@@ -0,0 +1,192 @@
...
+      }}
+    >
+      {children}
+    </HorizontalPanelAlignContext.Provider>
+  );
+}
+
+function SvgSide(props: React.SVGProps<SVGSVGElement>) {
+  return (
+    <svg
```
FlyersPh9, line 91: @VincentChabot1 is working on new icons for these.

VincentChabot1: Here is the PR
https://github.com/iTwin/iTwinUI-icons/pull/77

mayank99: Do these really need to be in the icons package or can they be part of the appui codebase?

GerardasB: For now we could keep this in AppUI repo, until there are other consumers for these icons. My main concern for initial icons was styling and licensing.

VincentChabot1: [DockIcons.zip](https://github.com/iTwin/appui/files/14130173/DockIcons.zip)

FlyersPh9: I replaced the SVGs with the ones in the ZIP.  Raphael is using transforms to get the other 3 icons, if we want to use SVGs instead of transforms, that still needs to be done.

FlyersPh9: center-bottom
center-top
justify-bottom
justify-top
left-bottom
left-top
right-bottom
right-top

raplemie: Is it by design that they are not squares ?
image

GerardasB: I think we should be fine using the transforms.

GerardasB (approved):
LGTM!

PR #721, opened 2024-02-05
Add `widgetActionDropdown` preview feature

GerardasB:
## Changes

This PR adds the `widgetActionDropdown` preview feature to display widget title bar buttons in a drop-down menu when a specified threshold value is exceeded.

image

## Testing

Added additional story and tested via the standalone test app.

### DIFF
--- a/ui/appui-react/src/test/layout/widget/Dock.test.tsx
+++ b/ui/appui-react/src/test/layout/widget/Dock.test.tsx
@@ -6,22 +6,29 @@ import { fireEvent, render } from "@testing-library/react";
 import * as React from "react";
 import * as sinon from "sinon";
 import type { NineZoneDispatch } from "../../../appui-react/layout/base/NineZone";
-import { NineZoneDispatchContext } from "../../../appui-react/layout/base/NineZone";
+import {
+  NineZoneDispatchContext,
+  NineZoneLabelsContext,
+} from "../../../appui-react/layout/base/NineZone";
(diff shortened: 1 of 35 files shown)


ui/appui-react/src/test/layout/widget/Dock.test.tsx
```diff
@@ -6,22 +6,29 @@ import { fireEvent, render } from "@testing-library/react";
...
+  NineZoneDispatchContext,
+  NineZoneLabelsContext,
+} from "../../../appui-react/layout/base/NineZone";
 import { Dock } from "../../../appui-react/layout/widget/Dock";
 
 describe("Dock", () => {
   it("should dispatch TOOL_SETTINGS_DOCK", () => {
     const dispatch = sinon.stub<NineZoneDispatch>();
-    const { container } = render(
+    const component = render(
```
saskliutas, line 18: Is there a reason to avoid destrucutring `render` result? E.g. `const { getByTitle } = render(...)`. In my opinion desctructuring `render` result makes test cleaner as elements lookup is less verbose.

saskliutas (approved):

ui/appui-react/src/test/layout/widget/Dock.test.tsx
GerardasB: Just a personal preference, I find it easier to debug the tests when/if needed (all the helpers are available on the returned object)

PR #729, opened 2024-02-12
Deprecate much of FrameworkUiAdmin

ben-polinsky:
Deprecate much of `FrameworkUiAdmin` in anticipation of `UiAdmin` being deprecated in core. In most cases, functions are moved to `UiFramework`. New functions which still use appui-abstract types to be deprecated have been marked as internal until a plan is formalized to fully deprecate those types. 

I have set up ref sharing via the react context api for `ConfigurableUiContent` so we can get the wrapper element in a React way.

A few `FrameworkUiAdmin` methods will not have 1:1 functionality after the move to `UiFramework`:
- `showHTMLElement` has been replaced with `showComponent` in the spirit of embracing React.
-  It is user's responsibility to gather and filter the `ToolList` from `iModelApp.tools.getToolList()`. 
- No featureFlags or ability to disable/endable toggling of the `keyInPalette`.
- The `showHeight...` and `showWidthInputEditor`s have been replaced with a single method: `showDimensionEditor` with an option to choose height or width. 
- `Keyin` types have been moved to their own module.


ui/appui-react/src/appui-react/uiadmin/FrameworkUiAdmin.tsx
```diff
@@ -180,20 +173,16 @@ export class FrameworkUiAdmin extends UiAdmin {
...
-      hidePopup,
-      hidePopup
-    );
+    return PopupManager.showKeyinPalette(keyIns, el);
   }
 
-  /** Hides the Key-in Palette. */
+  /**
+   * Hides the Key-in Palette.
+   * @deprecated in 4.10.x. Use {@link UiFramework.hideComponent}
```
calebmshafer, line 184: May want to add the same additional information on what to pass into it. Similar to `showKeyinPaletter` message.

common/api/appui-react.api.md
```diff
@@ -4840,8 +4911,11 @@ export class UiFramework {
     // @internal (undocumented)
     static loggerCategory(obj: any): string;
     static readonly onUiVisibilityChanged: UiVisibilityChangedEvent;
+    static openContextMenu(items: AbstractMenuItemProps[], location: XAndY, anchorElement?: HTMLElement): boolean;
```
GerardasB, line 4914: Same concern as before, will we deprecate these abstract types? If yes, we need a better strategy for these new APIs/types...

common/api/imodel-components-react.api.md
```diff
@@ -440,6 +439,9 @@ export interface DrawingViewportChangeEventArgs {
     rotation: Matrix3d;
 }
 
+// @public
+export function emitTimelineControlEvent({ uiComponentId, timelineAction, }: TimelinePausePlayArgs): void;
```
GerardasB, line 443: I understand that the point is to support the previous API, however, it makes little sense to have this sort of declarative API.
Could we simply add a `isPlaying` prop to a component for now and see what issues/complaints we get?

ui/appui-react/src/appui-react/UiFramework.ts
```diff
@@ -854,4 +906,354 @@ export class UiFramework {
...
+    offset: XAndY,
+    onItemExecuted: OnItemExecutedFunc,
+    onCancel: OnCancelFunc,
+    relativePosition?: RelativePosition,
+    anchorElement?: HTMLElement
+  ): boolean {
+    if (relativePosition === undefined)
+      relativePosition = RelativePosition.TopRight;
+
+    const el = UiFramework.resolveHtmlElement(anchorElement)
```
GerardasB, line 1020:
```suggestion
    relativePosition: RelativePosition = RelativePosition.TopRight,
    anchorElement?: HTMLElement
  ): boolean {
    const el = UiFramework.resolveHtmlElement(anchorElement)
```
Can be replaced with an initializer.

ui/appui-react/src/appui-react/UiFramework.ts
```diff
@@ -854,4 +906,354 @@ export class UiFramework {
...
+  public static showToolbar(
+    toolbarProps: AbstractToolbarProps,
+    location: XAndY,
+    offset: XAndY,
+    onItemExecuted: OnItemExecutedFunc,
+    onCancel: OnCancelFunc,
+    relativePosition?: RelativePosition,
+    anchorElement?: HTMLElement
+  ): boolean {
+    if (relativePosition === undefined)
```
GerardasB, line 1017: This is not ideal: we have 2x of the same code.
Can we simply call `UiFramework.showToolbar` in deprecated `FrameworkUiAdmin`? There are other places like this, not commenting everywhere.

ui/appui-react/src/appui-react/popup/ComponentPopup.tsx
```diff
@@ -0,0 +1,80 @@
...
+import { PopupManager } from "./PopupManager";
+import { PositionPopup } from "./PositionPopup";
+import type { OnCancelFunc, RelativePosition } from "@itwin/appui-abstract";
+import { useEffect, useState } from "react";
+import { WrapperContext } from "../configurableui/ConfigurableUiContent";
+
+/**
+ *
+ */
+export interface ComponentPopupProps extends Omit<PopupPropsBase, "el"> {
```
GerardasB, line 19: Let's not export Props types, since `React.ComponentProps<typeof ComponentPopup>` type helper is available and it makes it easier to stay backwards compatible.

ui/appui-react/src/appui-react/popup/ComponentPopup.tsx
```diff
@@ -0,0 +1,80 @@
...
+  relativePosition: RelativePosition;
+  orientation: Orientation;
+  onCancel: OnCancelFunc;
+  anchorRef?: React.RefObject<HTMLElement>;
+}
+
+/**
+ * @alpha
+ */
+export const ComponentPopup: React.FC<ComponentPopupProps> = ({
```
GerardasB, line 30: Do we want/need this exported? What is the use-case for this component (missing docs)? Can we think of a better name?

common/api/summary/appui-react.exports.csv
```diff
@@ -619,4 +627,5 @@ internal;WidgetsChangedEventArgs
 public;WidgetState
 public;WidgetStateChangedEvent 
 public;WidgetStateChangedEventArgs
-public;WidgetType
\ No newline at end of file
+public;WidgetType
+public;WrapperContext: React_2.Context
```
GerardasB, line 631: Let's keep `WrapperContext` internal.

common/api/summary/appui-react.exports.csv
```diff
@@ -400,6 +407,7 @@ alpha;SheetNavigationAid
 alpha;SheetNavigationAidControl 
 alpha;SheetNavigationProps 
 alpha;SheetsModalFrontstage 
+public;ShowComponentOptions
```
GerardasB, line 410: Similar to component props, can we not export Args/Options type for easier backwards compatibility?

GerardasB (changes requested):

ui/appui-react/src/appui-react/popup/PopupManager.tsx
```diff
@@ -49,6 +51,17 @@ export interface PopupInfo {
   parentDocument: Document;
 }
 
+/**
+ * @public
```
GerardasB, line 55: nit:
I'm not sure if we could have a rule for linter, but please keep this in one line: `/** @public */`
In other places prefer to add content in first line.
```
/** This component...
 * @internal
 */
```

common/api/appui-react.api.md
```diff
@@ -2075,40 +2079,64 @@ export interface FrameworkToolSettings {
...
+    // @deprecated
     get isFocusOnHome(): boolean;
-    // (undocumented)
-    get localizedKeyinPreference(): KeyinFieldLocalization;
-    set localizedKeyinPreference(preference: KeyinFieldLocalization);
-    // @internal (undocumented)
-    onInitialized(): void;
-    openDialog(uiDataProvider: DialogLayoutDataProvider, title: string, isModal: boolean, id: string, optionalProps?: DialogProps_2): boolean;
+    // @deprecated (undocumented)
+    get localizedKeyinPreference(): Keyin.KeyinFieldLocalization;
```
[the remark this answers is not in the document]
ben-polinsky: Hm... it also doesn't complain with non-enum types and if the name matches and the [type is in another namespace instead of global](https://www.typescriptlang.org/play?#code/KYOwrgtgBAKsDOAXKBvAUFTUAaUC8UADADQZYCa+UAjGgL5pqICeADsLG8NVellAEMAXFCQAnAJYgA5mUwAjEeAjzgY+o1CRYCRDz5ZcBEnKiUCtBmmAAPVgHsxyEAIgJWAgMYcAaqlO2Dk5QLOyc7DwEBvyYwqKIkjKm-IpQyqpimKYMVoGOzq7uXhww-vx5wVrQcEj6yYZUJjEUVNSk-FbldvlQVTpIZc04je1D5jSjzQD0U1AAWlQATMRQM1AA8gCqAMIAElAA7gAWoFBgIJ72EG4giMAAJtmMXUHIoSVckYPNceJSskMoKl0mosh0NIw0AAzc6eRASewgEK6AAUNhENUQAEpvlBLiB4PYADbAAB0RPs0jRWIA3BCYRd4YjkUhFmiRD5STBPljonjEYSSeTKdTTGJgIgwGIkTZ6HTGHckCjCLSmKjMaTsKrFYgUZjqJraatZls9mqlTAubpDTTjVAiWBPABrZgAfnNust+ptdtNu0Y+IGiCO-wxn14pjiACIQ1HJkCRIsIYGCm54B5vDAQzIMVzw1FIyIo8x7HHTKkAMwQknIcXwXPciJ0mtQOuLMNNtB1qg6tkuNMZ4BZ-5YrsIRY93R9wrp4rDmSjoA): 

Willing to be strict, but perhaps this is okay? To answer the below question of why, I wanted to move the type outside of `FrameworkUiAdmin`, keeping the same name, but allowing users to migrate to the non-deprecated version. Without putting it in a namespace, we'd have conflicts exporting from the barrel file. 

I suppose I could export the new version but not in the barrel (and remove the namespace). Then, revert to using the old deprecated type for any public apis.

common/api/appui-react.api.md
```diff
@@ -2818,14 +2849,29 @@ export interface KeyboardShortcutProps extends ItemProps {
     shortcuts?: KeyboardShortcutProps[];
 }
 
-// @public
+// @public (undocumented)
+export namespace Keyin {
```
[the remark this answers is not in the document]
ben-polinsky: See above answer

ui/appui-react/src/appui-react/popup/PopupManager.tsx
ben-polinsky: Will do thanks. I'm sure we could add a rule if we think it is worth it.

common/api/imodel-components-react.api.md
ben-polinsky: I can look into that, sure.

ui/appui-react/src/appui-react/UiFramework.ts
ben-polinsky: Yep will look into it.

ben-polinsky:
I was thinking about the move from `showKeyinPalette` to `showComponent` last night. I think it's cruel :) I get trying to clean up some of these apis, but we're not gaining too much (still have all these other show/hide methods) and the replacement is a bit clunky to use, especially if we're making the user keep track of the id for the keyinpalette (there's little chance they'd want two keyinpalettes). 

I may re-expose that method on UiFramework.

@calebmshafer

calebmshafer:
> I was thinking about the move from `showKeyinPalette` to `showComponent` last night. I think it's cruel :) I get trying to clean up some of these apis, but we're not gaining too much (still have all these other show/hide methods) and the replacement is a bit clunky to use, especially if we're making the user keep track of the id for the keyinpalette (there's little chance they'd want two keyinpalettes).
> 
> I may re-expose that method on UiFramework.
> 
> @calebmshafer

I'm fine with that for now to avoid a jarring UX. Maybe something to consider as we look at the API in the future. Wondering @GerardasB thoughts here as well.

common/api/appui-react.api.md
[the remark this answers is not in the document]
GerardasB: Resolving in favor of other comment/

common/api/appui-react.api.md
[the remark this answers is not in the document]
ben-polinsky: Ah ha, right! Thanks

ben-polinsky:
> > > I was thinking about the move from `showKeyinPalette` to `showComponent` last night. I think it's cruel :) I get trying to clean up some of these apis, but we're not gaining too much (still have all these other show/hide methods) and the replacement is a bit clunky to use, especially if we're making the user keep track of the id for the keyinpalette (there's little chance they'd want two keyinpalettes).
> > > I may re-expose that method on UiFramework.
> > > @calebmshafer
> > 
> > 
> > I'm fine with that for now to avoid a jarring UX. Maybe something to consider as we look at the API in the future. Wondering @GerardasB thoughts here as well.
> 
> I agree with Ben, let's keep the API as is for now. If we want to refactor this we need to revisit the whole PopupManager, PopupRenderer approach.

As an aside, using a static from outside a component to initialize state inside of a component is a... choice :) (See PopupRenderer)

common/api/appui-react.api.md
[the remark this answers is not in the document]
ben-polinsky: Done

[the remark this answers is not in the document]
ben-polinsky: Done

common/api/appui-react.api.md
```diff
@@ -2841,11 +2887,11 @@ export function KeyinPalettePopup({ el, id, keyins, onCancel, onItemExecuted, }:
 // @public
 export interface KeyinPalettePopupProps {
     // (undocumented)
-    el: HTMLElement;
+    el?: HTMLElement;
     // (undocumented)
     id: string;
     // (undocumented)
-    keyins: KeyinEntry[];
+    keyins: Keyin.KeyinEntry[];
```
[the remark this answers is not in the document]
ben-polinsky: Fixed.

common/api/imodel-components-react.api.md
ben-polinsky: That makes sense to me. I left the old deprecated way to avoid breaking and added a note that the two ways are mutually exclusive and should not be used together.

Haven't used class components in a while :)

ui/appui-react/src/appui-react/popup/ComponentPopup.tsx
ben-polinsky: Fixed.

ui/appui-react/src/appui-react/popup/ComponentPopup.tsx
ben-polinsky: We do need it exported so it can be used by the `PopupManager`, unless we want to inline it over there. Name-wise, it's a bit tough. What the component does is wrap and position a component as a "popup." But there's already a component `PositionPopup` which could get confusing. 

As discussed, all of these and the PopupManager/Renderer could be revisited - there's a fair amount of shared functionality, but not sure this is the best PR to do it in.  

Doc added. Set access level to @internal.

ui/appui-react/src/appui-react/UiFramework.ts
ben-polinsky: These should be fixed.

common/api/summary/appui-react.exports.csv
ben-polinsky: Done

common/api/summary/appui-react.exports.csv
ben-polinsky: Yep, fixed.

ui/appui-react/src/appui-react/popup/PopupManager.tsx
ben-polinsky: Fixed.

common/api/imodel-components-react.api.md
ben-polinsky: Done

common/api/appui-react.api.md
```diff
@@ -3895,6 +3944,20 @@ export class SheetsModalFrontstage implements ModalFrontstageInfo {
...
+    // (undocumented)
+    anchorRef?: React_2.RefObject<HTMLElement>;
+    // (undocumented)
+    location: XAndY;
+    // (undocumented)
+    offset: XAndY;
+    // (undocumented)
+    onCancel: OnCancelFunc;
+    // (undocumented)
+    relativePosition: RelativePosition;
```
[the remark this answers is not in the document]
ben-polinsky: Taking a look now. Inferring that:
image
`RightTop` and `LeftTop` are "actually" alignments? Meaning your "start" and  "end" literals?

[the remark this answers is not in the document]
ben-polinsky: First attempt pushed. It's not quite what you had in mind, so let me know if you're set on the `dashed-syntax` and why exactly you were interested in the string template literals. 

I'm also confused by RelativePosition in general - TopLeft vs LeftTop for instance. I see how the difference is calculated between them (in [Popup.tsx](https://github.com/iTwin/appui/blob/6bcc75cec129d5c08f86edff91ce736c34cf8796/ui/core-react/src/core-react/popup/Popup.tsx#L376) for example) but I don't understand semantically what is going on. Why, for instance, is there no `LeftBottom` or `RightBottom`?  Why is the `offset` only considered sometimes? 

Any clarification would be helpful :)

ui/appui-react/src/appui-react/popup/ComponentPopup.tsx
GerardasB: Thanks, I meant access level (release tag).

common/api/imodel-components-react.api.md
```diff
@@ -933,9 +933,11 @@ export enum TimelinePausePlayAction {
 }
 
 // @public
-export interface TimelinePausePlayArgs extends GenericUiEventArgs {
```
GerardasB, line 936: Breaking?

common/api/appui-react.api.md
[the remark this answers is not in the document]
GerardasB: I wasn't advocating to try and match the existing `RelativePosition` types to new literal values for new APIs (since I agree that they do not make much sense semantically..).
Seems like side (left/right/top/bottom) + placement (start/end) should be enough i.e. https://github.com/floating-ui/floating-ui/blob/495954da10d2abe8a5e6a0f0247582f071d6f031/packages/utils/src/index.ts#L4

ui/appui-react/src/appui-react/utils/Placement.ts
```diff
@@ -0,0 +1,62 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+type Side = "Left" | "Top" | "Right" | "Bottom";
+type Corner = "TopLeft" | "TopRight" | "BottomLeft" | "BottomRight" | "RightTop" | "LeftTop"
+
+/**
+ * Identifies the placement of a given element.
+ * @public
+ */
+export type Placement = Side | Corner;
```
GerardasB, line 12: Let's define the `Alignment = "start" | "end";` and placement as ``` Placement = Side | `${Side}-${Alignment}` ```.

ui/appui-react/src/appui-react/utils/Placement.ts
```diff
@@ -0,0 +1,62 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+type Side = "Left" | "Top" | "Right" | "Bottom";
```
GerardasB, line 5: Let's keep this lower case to match floating UI (and iTwinUI types).

ui/appui-react/src/appui-react/utils/Placement.ts
```diff
@@ -0,0 +1,62 @@
...
+ * Identifies the placement of a given element.
+ * @public
+ */
+export type Placement = Side | Corner;
+
+/**
+ * Copy of appui-abstract's RelativePosition for use in below utilities.
+ * Should be removed N+1 major versions after the deprecation.
+ */
+enum RelativePosition {
```
GerardasB, line 18: Why not simply import it instead?

ui/appui-react/src/appui-react/popup/ComponentPopup.tsx
```diff
@@ -0,0 +1,74 @@
...
+import type { PopupPropsBase } from "./PopupManager";
+import { PopupManager } from "./PopupManager";
+import { PositionPopup } from "./PositionPopup";
+import { useEffect, useState } from "react";
+import { WrapperContext } from "../configurableui/ConfigurableUiContent";
+import type { Placement } from "../utils/Placement";
+
+// Props used for the ComponentPopup.
+interface ComponentPopupProps extends Omit<PopupPropsBase, "el"> {
+  Component: React.ReactElement;
```
GerardasB, line 18:
```suggestion
  children?: React.ReactNode;
```

ui/appui-react/src/appui-react/popup/ComponentPopup.tsx
```diff
@@ -0,0 +1,74 @@
...
+  placement: Placement;
+  orientation: Orientation;
+  onCancel: () => void;
+  anchorRef?: React.RefObject<HTMLElement>;
```
GerardasB, line 22:
```suggestion
  anchor?: HTMLElement;
```

ui/appui-react/src/appui-react/utils/Placement.ts
ben-polinsky: Eh, figured it would be one less abstract import to clean up, but if we have to clean this up anyway... kind of useless.

ben-polinsky: fixed

ui/appui-react/src/appui-react/utils/Placement.ts
ben-polinsky: fixed

ui/appui-react/src/appui-react/utils/Placement.ts
ben-polinsky: fixed

ben-polinsky:
Take a look at the `Placement` changes I made, I've added the math for `getPopupRect` in `CursorPopup.tsx`. As `showComponent`/`ComponentPopup` was the only new popup, I decided to work with that first. Checkout the tests in CursorPopup.test.tsx for an idea of the logic. 

Unsure if `offset` should change with the placement - currently I'm always subtracting the offset, but I could see the offset being "in the direction of the placement" meaning if the placement is "top" then the offset would adjust on the positive-y axis. I'll take a look at what other libraries do tomorrow.

ui/appui-react/src/appui-react/utils/Placement.ts
```diff
@@ -0,0 +1,79 @@
+import { RelativePosition } from "@itwin/appui-abstract";
+
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
```
GerardasB, line 6:
```suggestion
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 import { RelativePosition } from "@itwin/appui-abstract";

```

common/api/imodel-components-react.api.md
GerardasB: Should be fine w/ added `uiComponentId` property.

ui/appui-react/src/appui-react/cursor/cursorpopup/CursorPopup.tsx
```diff
@@ -104,61 +105,112 @@ export class CursorPopup extends React.Component<
...
-        popupRect.top = pt.y + offset.y;
-        popupRect.left = pt.x + offset.x;
-        popupRect.bottom = popupRect.top + popupSize.height;
-        popupRect.right = popupRect.left + popupSize.width;
-        break;
+    if (typeof relativePosition === "string") {
+      const placement = relativePosition.split("-");
+      const [mainPlacement, subPlacement] = placement;
+
+      switch (mainPlacement) {
```
GerardasB, line 117: nit: error prone duplication, instead we could allways map: `relativePosition` to `placement | X` (where X is a relative position that can't be mapped to placement). And just handle `placement` values and a few of RelativePosition values.

ui/appui-react/src/appui-react/cursor/cursorpopup/CursorPopup.tsx
```diff
@@ -30,7 +31,7 @@ export interface CursorPopupProps extends CommonProps {
   content: React.ReactNode;
   pt: XAndY;
   offset: XAndY;
-  relativePosition: RelativePosition;
+  relativePosition: RelativePosition | Placement;
```
GerardasB, line 34: Should we just deprecate `relativePosition` instead for better semantics? Placement is usually optional anyways.
```suggestion
  /** @deprecated in 4.10.0. Use `placement` instead. */
  relativePosition: RelativePosition;
  placement?: Placement;
```

GerardasB:
> Take a look at the `Placement` changes I made, I've added the math for `getPopupRect` in `CursorPopup.tsx`. As `showComponent`/`ComponentPopup` was the only new popup, I decided to work with that first. Checkout the tests in CursorPopup.test.tsx for an idea of the logic.
> 
> Unsure if `offset` should change with the placement - currently I'm always subtracting the offset, but I could see the offset being "in the direction of the placement" meaning if the placement is "top" then the offset would adjust on the positive-y axis. I'll take a look at what other libraries do tomorrow.

I think it mostly makes sense for `offset` to change bounds based on the side of the placement. I.e. 
```
const offset = 10;
switch (side) {
  case "top": // bounds-10 to move up by 10px
  case "bottom": // bounds+10 to move bottom by 10px
}
```

ui/appui-react/src/appui-react/cursor/cursorpopup/CursorPopup.tsx
ben-polinsky: Done

ui/appui-react/src/appui-react/cursor/cursorpopup/CursorPopup.tsx
GerardasB: Oh, I assumed there is an overlap, if that's not the case - feel free to ignore :)

ui/appui-react/src/appui-react/UiFramework.ts
```diff
@@ -854,4 +910,364 @@ export class UiFramework {
...
+   * @param relativePosition Position relative to the given location. Defaults to TopRight.
+   * @param anchorElement The HTMLElement that anchors the tool settings. If undefined, the location is relative to the overall window.
+   * @return true if the tool settings were displayed, false if the tool settings could not be displayed.
+   */
+  public static openToolSettingsPopup(
+    dataProvider: DialogLayoutDataProvider,
+    location: XAndY,
+    offset: XAndY,
+    onCancel: () => void,
+    placement: Placement | RelativePosition = "top-end",
```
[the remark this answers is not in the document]
ben-polinsky: Ah makes more sense, no worries :)

common/api/appui-react.api.md
```diff
@@ -3895,6 +3944,20 @@ export class SheetsModalFrontstage implements ModalFrontstageInfo {
     title: string;
 }
 
+// @public (undocumented)
+export interface ShowComponentOptions {
+    // (undocumented)
+    anchorRef?: React_2.RefObject<HTMLElement>;
```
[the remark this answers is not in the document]
ben-polinsky: Done

ui/appui-react/src/appui-react/UiFramework.ts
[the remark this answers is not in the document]
ben-polinsky: Done

common/api/appui-react.api.md
GerardasB: That's not a good option, since we are already introducing a `@public` type which we can't change after? Maybe we could mark these as `@internal` for now and have no replacement for now (not ideal either) :?

GerardasB (approved):

GerardasB (approved):
I'd appreciate it if you could fill out `NextVersion.md` as well :)

common/api/appui-react.api.md
ben-polinsky: Right, I was thinking it was core and could sit on master while devs worked off release branches... 
Hm maybe I'll suck it up and try to fit this in here :) Hopefully won't be too messy.

ben-polinsky:
> I'd appreciate it if you could fill out `NextVersion.md` as well :)

Absolutely.

ben-polinsky:
Most recent changes take advantage of the fact that most PopUp Components are in alpha or beta and modifies their prop types without regard for breaking changes.

ui/appui-react/src/appui-react/shared/ConditionalValue.ts
ben-polinsky, line 1: @GerardasB - A note: I initially added a generic version of this class, replacing the static methods with instance methods (the static methods make no sense, because they must narrow the type from `string | ConditionalStringValue` to `ConditionalStringValue` to even use ([see BackstageItemsManager.ts](https://github.com/iTwin/appui/blob/e5e8bdba7602e1cdfc9ac426dd371f875cf7c173/ui/appui-react/src/appui-react/backstage/BackstageItemsManager.ts#L146) for example).

However, this made the conversion from any types which used the old Conditional{string|boolean}Value to the new generic version very verbose and annoying. I still think it's something which should be done, but I am trying my best to limit the already large scope of this PR/task.

GerardasB: Yes, the common interface would help a lot. I've come up with this https://github.com/iTwin/appui/blob/7834abf9488f660af465a267751110e6c2ca0d22/ui/appui-react/src/appui-react/hooks/useConditionalValue.tsx but that is still not merged.
I agree, we should stop expanding the scope of this PR, let's aim for multiple smaller ones..

common/api/core-react.api.md
```diff
@@ -122,6 +121,13 @@ export interface BadgeProps extends CommonProps {
     svg: any;
 }
 
+// @public
+export enum BadgeType {
```
GerardasB, line 125: Ideally I'd inline this in the component props (so we can i.e. add new union values).
However, since `BadgeType` is used in multiple APIs it will get quite repetitive.
Either way, let's avoid adding enums and use string literal types instead.
Ideally with the capability to add support for additional/custom badge types in the future (in non breaking way w/o introducing additional types).
I.e. does this make sense? `type BadgeType = "new" | "technical-preview" | (string & {});`
https://www.typescriptlang.org/play?#code/PTAEDkFNIE10BmB7ATqAlgOwC6QDZ7oDOkmJAUNgJ4AOkoAgplQMrYpYDmoAvKABRF2XUADJQAbwC+ASgDc5SAA8aqbKGp1QAIQCGMTpAAqtenwBEmSAHdzoAD6hzuAMYALTOhe68AWhookABu6DZ2jkyswpicCuTkCACumC7Y6EiYGpBC-EoAXDr6hiZ0MpLkUvG4OZZh8pTZ2PyWSNi+nIG6bQBGiW3WkL7emL76ML5W1r7dRdmjRL4DBOYy8UA

ben-polinsky: Today I learned that is the recommended way to not widen string literals inside of a union with other strings. Hoped there would be an `as const` or something. Agree that sounds good.

common/api/appui-react.api.md
```diff
@@ -2995,10 +3062,10 @@ export class MenuItemHelpers {
     // (undocumented)
     static createMenuItemNodes(itemList: MenuItem[]): React_2.ReactNode[];
     // (undocumented)
-    static createMenuItems(itemPropsList: MenuItemProps[], onSelection?: () => void): MenuItem[];
+    static createMenuItems(itemPropsList: CursorMenuItemProps[], onSelection?: () => void): MenuItem[];
 }
 
-// @public
+// @public @deprecated
```
GerardasB, line 3068: Can we get away w/o deprecation and introduction of `CursorMenuItemProps`? Are `CursorMenuItemProps` and `AbstractMenuItemProps` identical?

ui/appui-react/src/appui-react/shared/ConditionalValue.ts
GerardasB: I am a little cautious about breaking alpha and beta APIs, since some of them were marked like that for quite long and should probably have already been moved to public. Please, let's try to break those only if there's no other option..

common/api/appui-react.api.md
ben-polinsky: No, they're not the same. The new `CursorMenuItemProps` extends `ItemProps` from `appui-react` rather than `CommonItemProps` from `appui-abstract`. Even without that, we'd run into issues with {String/Boolean}ConditionalValue classes not being interchangeable (private member class comparison).

ui/appui-react/src/appui-react/shared/ConditionalValue.ts
ben-polinsky: Alright I will take a look

common/api/appui-react.api.md
```diff
@@ -540,6 +545,13 @@ export interface BackstageToggledArgs {
     readonly isOpen: boolean;
 }
 
+// @public
+export enum BadgeType {
```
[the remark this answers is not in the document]
ben-polinsky: done

common/api/core-react.api.md
ben-polinsky: done - interesting trick

common/api/appui-react.api.md
GerardasB: Btw, can/should we do the same trick we did with enums for ConditionalValue classes?

ben-polinsky: Good idea, let me see.

ben-polinsky: Looks like we can. Changed.

ui/appui-react/src/appui-react/shared/ConditionalValue.ts
ben-polinsky: Was able to make the changes without breaking.

ben-polinsky:
Ah nextVersion...

ben-polinsky:
> Ah nextVersion...

@GerardasB NextVersion updated. Added unit tests for new UIFramework methods which surfaced one bug I fixed. This existed before this PR. `findIndex` returns `-1` not `false` if nothing found.

https://github.com/iTwin/appui/pull/729/files#diff-1cb11f2cf9c88eef2fa2ef735236136d8d05733e98e817c190caf04ac16f380fL631

docs/changehistory/NextVersion.md
```diff
@@ -3,15 +3,77 @@
...
   - [Changes](#changes)
   - [Fixes](#fixes)
+- [@itwin/imodel-components-react](#itwinimodel-components-react)
+  - [Deprecations](#deprecations-1)
 
 ## @itwin/appui-react
 
+### Removals
+
+- Description. #PR
```
GerardasB, line 18: Remove a leftover

docs/changehistory/NextVersion.md
```diff
@@ -3,15 +3,77 @@
 Table of contents:
 
 - [@itwin/appui-react](#itwinappui-react)
+  - [Removals](#removals)
+  - [Deprecations](#deprecations)
+  - [Additions](#additions)
   - [Changes](#changes)
   - [Fixes](#fixes)
+- [@itwin/imodel-components-react](#itwinimodel-components-react)
```
GerardasB, line 11: Add a note about addition of BadgeType to core-react.

common/changes/@itwin/core-react/adminui-removal-adjustments_2024-02-21-19-45.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-react",
+      "comment": "add BadgeType",
```
GerardasB, line 5:
```suggestion
      "comment": "Add BadgeType",
```

common/changes/@itwin/appui-react/adminui-removal-adjustments_2024-02-21-19-45.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "eprecate FrameworkUiAdmin; introduce Placement, a replacement for ",
```
GerardasB, line 5:
```suggestion
      "comment": "Deprecate `FrameworkUiAdmin`. Introduce `Placement`, a replacement for `RelativePosition`",
```

GerardasB (approved):

common/api/appui-react.api.md
```diff
@@ -3382,7 +3447,7 @@ export interface PopupInfo {
     // (undocumented)
     id: string;
     // (undocumented)
-    parentDocument: Document;
+    parentDocument?: Document;
```
GerardasB, line 3450: Breaking.

common/api/appui-react.api.md
```diff
@@ -2841,7 +2894,7 @@ export function KeyinPalettePopup({ el, id, keyins, onCancel, onItemExecuted, }:
 // @public
 export interface KeyinPalettePopupProps {
     // (undocumented)
-    el: HTMLElement;
+    el?: HTMLElement;
```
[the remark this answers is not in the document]
ben-polinsky: fixed

common/api/appui-react.api.md
ben-polinsky: fixed

docs/changehistory/NextVersion.md
ben-polinsky: done - I left the removal heading, wasn't sure if you wanted it as a placeholder

common/api/appui-react.api.md
ben-polinsky: Taking a look at the remaining appui-abstract types I've used in new APIs in UIFramework... some optimism I can find a way forward.

ben-polinsky: Alright, as discussed, any `UiFramework` methods added which utilize appui-abstract functionality that doesn't have a deprecation plan are `@internal` and corresponding `FrameworkUiAdmin` methods have not deprecated. Similarly, `FrameworkUiAdmin` itself will not yet be deprecated. Let me know if you see anything I missed.

Issue #126, opened 2023-02-01
Use iTwinUI keyboard components


GerardasB:
Investigate using iTwinUI keyboard components in favor of custom implementation in `ToolAssistanceField`.

![Screenshot 2024-03-29 at 12 09 34](https://github.com/iTwin/appui/assets/10091419/d3c52fe7-3861-4d1a-9912-a89ce759d17d)

Update: current `KeyboardShortcutMenu` implementation displays shortcuts as underlined symbols (example below). It can be changed to iTwinUI `Kbd` component.

![keyboard-shortcut-menu-test-1-chromium-win32](https://github.com/iTwin/appui/assets/138560995/c147e4ab-9b49-4362-ae45-32ef099c66f0)

mayank99:
I don't think this should use `<Kbd>`. That component is intended to be used alongside other static text, e.g. for documentation purposes or to indicate keyboard shortcuts in a **Help** page.

A better approach would be to show these as static, muted, monospace text. Ideally right aligned like in macOS menubar:

![image](https://github.com/iTwin/appui/assets/9084735/e08c4463-51fc-4ec2-b83d-71563849c99a)

I guess right aligned becomes difficult if you also have to show an arrow/caret, so you could show it as sublabel underneath the main label, though it wouldn't look as clean.

GerardasB:
> I don't think this should use `<Kbd>`. That component is intended to be used alongside other static text, e.g. for documentation purposes or to indicate keyboard shortcuts in a **Help** page.

Good point, I agree about the menu. I've updated the issue description w/ image of `ToolAssistanceField`, which I think is a good candidate for  `<Kbd />`.

PR #745, opened 2024-02-27
Fix padding variables overrides

ignas-k:
## Changes
This PR removes uicore-bordered-padding and uicore-unbordered-padding variable overrides in inputs/variables.scss. The overriden values were used only by ValidationTextBox which is an alpha component and it is not used by anyone (and it should be removed in favor of iTwinUI Input component).

## Testing
No practical changes, tested via standalone test-app and storybook. Updated snapshots because before playwright tests rendered content differently (and incorrectly) than storybook and test-app.


full-stack-tests/ui/tests/keyboardshortcut/keyboard-shortcut-menu.test.ts-snapshots/keyboard-shortcut-menu-test-1-chromium-win32.png
FlyersPh9, line 1: I assume the `M`, `A`, `R` are keyboard shortcuts?  If yes, is it possible to use the [iTwinUI-React Kbd component](https://itwinui.bentley.com/docs/kbd)?

GerardasB (approved):

full-stack-tests/ui/tests/keyboardshortcut/keyboard-shortcut-menu.test.ts-snapshots/keyboard-shortcut-menu-test-1-chromium-win32.png
ignas-k: I believe it is possible to use the `Kbd` component from iTwinUI. Found a similar issue and modified it: #126.

Issue #709, opened 2024-01-31
`MessageCenterField` visual changes


GerardasB:
**Is your feature request related to a problem? Please describe.**
`MessageCenterField` visuals are outdated. There is no indication of new messages, only a total message count, which the end-user can not clear.

**Describe the solution you'd like**
- Remove the chat bubble from the message center field
- Add a visual indication in the form of an active button when a popup is opened instead (see information button in https://itwin.github.io/iTwinUI/css/information-panel.html). Make sure that [iTwinUI borderless button is used #708](https://github.com/iTwin/appui/issues/708)
- Use the notification marker https://itwin.github.io/iTwinUI/react/?story=notificationmarker--basic to indicate new messages in this form (with https://itwin.github.io/iTwinUI-icons/#icons-chat icon)
![image](https://github.com/iTwin/appui/assets/10091419/b039c0a5-3481-4499-bac3-36698d812324)
- Re-use https://itwinui.bentley.com/docs/tabs#pill tabs instead of custom tabs in the popup
- Refactor `MessageCenterField` from a class to a function component

**When closed:**
![image](https://github.com/iTwin/appui/assets/849817/69fbd700-3fa5-4cb7-8847-af82e4fe2487)

**When opened:**
![image](https://github.com/iTwin/appui/assets/849817/515566ec-253e-43d1-844e-660793a5c268)

PR #746, opened 2024-02-27
Refactor of MessageCenterField

AdamMeza-Bentley:
### Complete changes requested in #709 

Most of the changes listed in the issue were solved by relying on iTwinUI components. Some of the major changes include:

- Replaced custom tabs with pill `Tab`'s from iTwinUI.
- Migrated `MessageCenterFIeld` from class to functional React.
- **Responsive notification marker:** After replacing the unwanted icon with `SvgChat` from iTwinUI, I added the `NotifcationMarker`. It rerenders based on the `updateMessageCenter` event from `MessageManager`, and adjusts to the  `msgType`. It also removes the notification after opened as per request from @FlyersPh9. 

One request that I wasn't able to fulfil was adding the active state to the `Message` button. This is due to the `isActive` prop being available only to the [`IconButton`](https://itwinui.bentley.com/docs/button#props-2). This seems like a gap in functionality on the component that I plan to fill out a request for.

### Visual Changes:

Old Version
Screenshot 2024-03-28 083258

New Version:

image

### Delete unnecessary components

There were a number of files deleted in this PR. Among them were several small React Components and all the corresponding SCSS files. 

- `Tab.tsx` which exported `MessageCenterTab` has been replaced it with iTwinUI component `PopOver`.
- `Indicator.tsx` which exported `MessageCenter`. This component felt to be a good example of the bloated nature of AppUI. It's functionality wasn't much more complex than `FooterIndiactor` which it utilized. It was only used in one place. Yet, it was tested several times, had several prop declarations, and mentioned in other files. 
- `Dialog.tsx` which exported `MessageCenterDialog` which relied almost entirely on `Dialog`. This was replaced with a simple div with the right `className`. I've noticed several similar files/components, `StatusBarPopUp` for example, which do almost nothing but add a class to another component. 

Deleting extraneous files seems like a good way to clean up AppUI. Doing so led to `MessageCenterFIeld` having more elaborate JSX than before. While best practice is usually to use more atomic React components, it feels as though one of the issues with AppUI is the "bloated" number of files. In the interest of solving this problem, I believe a "Marie Kando"/ minimalist approach would serve us best. How better to organize something than to simply have less of it? 

### Change file names

A common occurrence in AppUI is disparity in filename to exported function name, (see several examples listed above). This leads to several files with overlapping names, confusion about component relationships, and other navigation issues. I would love to talk with everyone else on the team to see if there's a reason for this. If not, it seems like something to add to a style guide. All that said, I've taken the liberty of changing file names where I've noticed this. 

## Testing

Several testing files were altered or deleted as a result of this refactor. Aside from the aforementioned deleted files, testing for `MessageCenterField` was updated to reflect the new html structure. 

Additionally, all functionality has been tested locally on a Windows machine.


ui/appui-react/src/appui-react/statusfields/MessageCenter.tsx
```diff
@@ -5,227 +5,165 @@
...
+      );
+    };
+  });
+
+  const getMessages = (tab: string): React.JSX.Element[] => {
+    return messages
+      .slice(0)
+      .reverse()
+      .map((details: NotifyMessageDetailsType, index: number) => {
+        /* istanbul ignore else */
```
GerardasB, line 85: Avoid adding these ignores. (not adding a comment for other places)

ui/appui-react/src/appui-react/statusfields/MessageCenter.tsx
```diff
@@ -5,227 +5,165 @@
...
-        const iconSpec = MessageManager.getIconSpecFromDetails(details);
-        const message = details.briefMessage;
+    return () => {
+      MessageManager.onMessagesUpdatedEvent.removeListener(
+        handleMessagesUpdatedEvent
+      );
+    };
+  });
+
+  const getMessages = (tab: string): React.JSX.Element[] => {
```
GerardasB, line 80: Keep implicit return type.
```suggestion
  const getMessages = (tab: string) => {
```

ui/appui-react/src/appui-react/statusfields/MessageCenter.tsx
```diff
@@ -5,227 +5,165 @@
...
-    }
-  }
-
-  private _handleMessagesUpdatedEvent = () => {
-    // istanbul ignore else
-    if (this._unloadMessagesUpdatedHandler)
-      this.setState({ messageCount: MessageManager.messages.length });
+export const MessageCenterField: React.FC = () => {
+  const [messages, setMessages] = React.useState(MessageManager.messages);
+  const [notify, setNotify] = React.useState("");
```
GerardasB, line 33: Seems like this can be removed? [Use computed](https://react.dev/learn/choosing-the-state-structure#avoid-redundant-state)

ui/appui-react/src/appui-react/statusfields/MessageCenter.tsx
```diff
@@ -5,227 +5,165 @@
...
+  const [isOpen, setIsOpen] = React.useState(false);
+  const [active, setActive] = React.useState(false);
+
+  const indicator = React.createRef<HTMLDivElement>();
+  const title = UiFramework.translate("messageCenter.messages");
+  const divStyle = { height: "100%" };
```
GerardasB, line 39: Can we inline this w/o a variable declaration?

ui/appui-react/src/appui-react/statusfields/MessageCenter.tsx
```diff
@@ -5,227 +5,165 @@
...
+        handleMessagesUpdatedEvent
+      );
+    };
+  });
+
+  const getMessages = (tab: string): React.JSX.Element[] => {
+    return messages
+      .slice(0)
+      .reverse()
+      .map((details: NotifyMessageDetailsType, index: number) => {
```
GerardasB, line 84: Use implicit types.
```suggestion
      .map((details, index) => {
```

ui/appui-react/src/appui-react/statusfields/MessageCenter.tsx
```diff
@@ -5,227 +5,165 @@
...
+        /* istanbul ignore else */
+        if (messages.length > 0) {
```
GerardasB, line 86: Can this be false? :)

ui/appui-react/src/appui-react/statusfields/MessageCenter.tsx
```diff
@@ -5,227 +5,165 @@
...
+    return () => {
+      MessageManager.onMessagesUpdatedEvent.removeListener(
+        handleMessagesUpdatedEvent
+      );
+    };
+  });
+
+  const getMessages = (tab: string): React.JSX.Element[] => {
+    return messages
+      .slice(0)
```
GerardasB, line 82: nit (personal pref): use spread `return [...messages].reverse`

ui/appui-react/src/appui-react/statusfields/MessageCenter.tsx
```diff
@@ -5,227 +5,165 @@
...
+            !notify ? (
+              <SvgChat />
+            ) : (
+              <NotificationMarker status={notifyStatus()}>
+                <SvgChat />
+              </NotificationMarker>
+            )
+          }
+        >
+          <div // eslint-disable-line jsx-a11y/click-events-have-key-events
```
GerardasB, line 156: Why is this div needed? Seems like we can re-factor to use buttons props, i.e. `Button["startIcon"]`

ui/appui-react/src/test/layout/widget-panels/Content.test.snap
```diff
@@ -0,0 +1,31 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`WidgetPanelsContent should render 1`] = `
```
GerardasB, line 3: Avoid adding new snapshot tests. We are trying to get rid of these in all our components, use react testing library queries instead.

ui/core-react/src/core-react/tabs/Tabs.tsx
```diff
@@ -43,7 +43,7 @@ export interface TabsProps
     CommonProps {
   /** Text shown for each tab
    * @public */
-  labels: Array<string | TabLabel>;
+  labels: Array<string | TabLabel | any>;
```
GerardasB, line 46: Remove public tag. Adding `any` to a public type is a breaking change, which we can't do.
```
/** Text shown for each tab */
labels: Array<string | TabLabel>;
```

GerardasB (changes requested):
Looks like a good progress, we need to add some offset to the popup as well.

ui/appui-react/src/appui-react/layout/footer/message-center/Message.tsx
```diff
@@ -7,37 +7,49 @@
...
+/** Message entry in [[MessageCenter]] component.
  * @internal
  */
-export class MessageCenterMessage extends React.PureComponent<MessageCenterMessageProps> {
-  public override render() {
-    const className = classnames(
-      "nz-footer-messageCenter-message",
-      this.props.className
-    );
+export const MessageCenterMessage: React.FC<MessageCenterMessageProps> = (
```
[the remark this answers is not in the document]
AdamMeza-Bentley: ok yeah! I wasn't sure if that was an intentional choice or not.

ui/appui-react/src/appui-react/layout/footer/message-center/Message.tsx
```diff
@@ -7,37 +7,49 @@
  */
 
 import "./Message.scss";
-import classnames from "classnames";
 import * as React from "react";
 import type { CommonProps } from "@itwin/core-react";
+import type { MessageType } from "@itwin/core-react";
```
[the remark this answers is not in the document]
GerardasB: This applies to `@itwin/itwinui-react` declaration as well.

ui/appui-react/src/appui-react/statusfields/MessageCenterField.tsx
```diff
@@ -0,0 +1,165 @@
...
+      );
+    };
+  });
+
+  const getMessages = (tab: string): React.JSX.Element[] => {
+    console.log(messages);
+    return [...messages]
+      .reverse()
+      .map((details: NotifyMessageDetailsType, index: number) => {
+        /* istanbul ignore else */
```
GerardasB, line 78: remove istanbul ignores

ui/appui-react/src/appui-react/statusfields/MessageCenterField.tsx
```diff
@@ -0,0 +1,165 @@
...
+        if (messages.length > 0) {
+          const iconClassName = MessageManager.getIconClassName(details);
+          const iconSpec = MessageManager.getIconSpecFromDetails(details);
+          const message: MessageType = details.briefMessage;
+          /* istanbul ignore else */
```
GerardasB, line 83: remove istanbul ignores

ui/appui-react/src/appui-react/statusfields/MessageCenterField.tsx
```diff
@@ -0,0 +1,165 @@
...
+      message.priority === OutputMessagePriority.Error ||
+      message.priority === OutputMessagePriority.Fatal
+    );
+  };
+
+  const notifyStatus = () =>
+    messages.some((msg) => isProblemStatus(msg)) ? "negative" : "primary";
+
+  const handleMessagesUpdatedEvent = () => {
+    // istanbul ignore else
```
GerardasB, line 55: remove istanbul ignores

ui/appui-react/src/appui-react/statusfields/MessageCenterField.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { SvgChat } from "@itwin/itwinui-icons-react";
+import type { MessageType } from "@itwin/core-react";
+import { Icon } from "@itwin/core-react";
+import { UiFramework } from "../UiFramework";
+import { OutputMessagePriority } from "@itwin/core-frontend";
+import { MessageCenterMessage } from "../layout/footer/message-center/Message";
+import { MessageManager } from "../messages/MessageManager";
+import { TitleBar } from "../layout/footer/dialog/TitleBar";
+import type { NotifyMessageDetailsType } from "../messages/ReactNotifyMessageDetails";
+import "../layout/footer/message-center/Dialog.scss";
```
GerardasB, line 25: Imported scss file name should match tsx file name (should only contain CSS for that component and not be imported by other components).

ui/appui-react/src/appui-react/statusfields/MessageCenterField.tsx
```diff
@@ -0,0 +1,165 @@
...
+
+    return () => {
+      MessageManager.onMessagesUpdatedEvent.removeListener(
+        handleMessagesUpdatedEvent
+      );
+    };
+  });
+
+  const getMessages = (tab: string): React.JSX.Element[] => {
+    console.log(messages);
```
GerardasB, line 74: Leftover console.log

ui/appui-react/src/appui-react/statusfields/MessageCenter.tsx
GerardasB: Much better now, I think we can still remove the `div` and probably even `span` for the label?

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.scss
```diff
@@ -0,0 +1,22 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+@import "~@itwin/core-react/lib/cjs/core-react/z-index";
+@import "../variables";
+
+.nz-footer-messageCenter-dialog {
+  width: 16vw;
```
GerardasB, line 9: Still using viewport units for width.

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,155 @@
...
+    MessageManager.registerAnimateOutToElement(indicator.current);
+
+    return () => {
+      MessageManager.onMessagesUpdatedEvent.removeListener(
+        handleMessagesUpdatedEvent
+      );
+    };
+  });
+
+  const getMessages = (tab: string): React.JSX.Element[] => {
```
GerardasB, line 72: Keep implicit return type.
```suggestion
  const getMessages = (tab: string) => {
```

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const isProblemStatus = (message: NotifyMessageDetailsType): boolean => {
+    // See priority values in DgnPlatform defined in NotificationManager
+    return (
+      message.priority === OutputMessagePriority.Error ||
+      message.priority === OutputMessagePriority.Fatal
+    );
+  };
+
+  const notifyStatus = () =>
+    messages.some((msg) => isProblemStatus(msg)) ? "negative" : "primary";
```
GerardasB, line 52: Why a function?
```suggestion
  const notifyStatus = messages.some((msg) => isProblemStatus(msg)) ? "negative" : "primary";
```

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+  const handleMessagesUpdatedEvent = () => {
+    setNotify(notifyStatus());
```
GerardasB, line 55: Seems like currently status will be `negative` if there were previously error messages in the message manager. Status should be calculated based on **new** messages only?

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,155 @@
...
+    setMessages(MessageManager.messages);
+  };
+
+  React.useEffect(() => {
```
GerardasB, line 59: Provide a dependency array to useEffect. `addListener` returns a cleanup, which can be inlaid: 
```
React.useEffect(() => {
  return event.addListener(() => { /* do some stuff */ });
}, []);
```

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,155 @@
...
+
+/** Message Center Field React component.
+ * @public
+ */
+export const MessageCenterField: React.FC = () => {
+  const [messages, setMessages] = React.useState(MessageManager.messages);
+  const [notify, setNotify] = React.useState("");
+  const [isOpen, setIsOpen] = React.useState(false);
+
+  const indicator = React.createRef<HTMLDivElement>();
```
GerardasB, line 35: nit: naming
```suggestion
  const indicatorRef = React.createRef<HTMLDivElement>();
```

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,155 @@
...
+  const handleMessagesUpdatedEvent = () => {
+    setNotify(notifyStatus());
+    setMessages(MessageManager.messages);
+  };
+
+  React.useEffect(() => {
+    MessageManager.onMessagesUpdatedEvent.addListener(
+      handleMessagesUpdatedEvent
+    );
+    MessageManager.registerAnimateOutToElement(indicator.current);
```
GerardasB, line 63: I think this should be moved into ref callback instead. Btw, indicator seems to not be set ATM.

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,155 @@
...
+      MessageManager.onMessagesUpdatedEvent.removeListener(
+        handleMessagesUpdatedEvent
+      );
+    };
+  });
+
+  const getMessages = (tab: string): React.JSX.Element[] => {
+    return [...messages]
+      .reverse()
+      .map((details: NotifyMessageDetailsType, index: number) => {
```
GerardasB, line 75: (FYI: I have resolved previous comments that were not addressed, I assume it's because GH sometimes decides to collapse unresolved comments). Use implicit types.
```suggestion
      .map((details, index) => {
```

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,155 @@
...
+    </>
+  );
+
+  return (
+    <Popover
+      content={
+        <>
+          <TitleBar title={title}></TitleBar>
+
+          <Tabs.Wrapper type="pill" role="tablist">
```
GerardasB, line 122: Isn't `role` attribute set by `Tabs.TabList`?

GerardasB (changes requested):
Getting very close, some of these are nit-picky, nice job! :)

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
AdamMeza-Bentley: the empty dependency array was throwing an error from the linter

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
AdamMeza-Bentley: as in use useRef instead of useEffect here?

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
AdamMeza-Bentley: yeah I was wondering about that. is "new"/ unread messages somehing that messageManager already handles? otherwise we would need to add logic to differentiate between a message that is considered new/unread and all the others. this seems like it might get too complicated? or maybe im misunderstanding

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
AdamMeza-Bentley: fixed that

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
GerardasB: `useRef` will create a `RefObject`, but callback can be provided as a `Ref` as well https://react.dev/reference/react-dom/components/common#ref-callback

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
GerardasB: This needs investigation, but wouldn't message count be enough to differentiate new/old messages?

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
GerardasB: Inline the `handleMessagesUpdatedEvent` in the addListener.

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
GerardasB: `registerAnimateOutToElement` is missing now. That should be needed to add animate-out toast messages animation to message center.

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
GerardasB: `notifyStatus` should change based on new messages only, instead of all messages

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
AdamMeza-Bentley: I removed it and didn't notice any change in functionality/animation. would you be able to provide a video of what I should be looking for?

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
AdamMeza-Bentley: > Inline the handleMessagesUpdatedEvent in the addListener.

im sorry but I don't follow. Isnt the `handleMessagesUpdatedEvent` already inline with the addListener?

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
GerardasB: animate-out-to

ui/appui-react/src/appui-react/statusfields/MessageCenterField.tsx
```diff
@@ -0,0 +1,165 @@
...
+import { MessageCenterMessage } from "../layout/footer/message-center/Message";
+import { MessageManager } from "../messages/MessageManager";
+import { TitleBar } from "../layout/footer/dialog/TitleBar";
+import type { NotifyMessageDetailsType } from "../messages/ReactNotifyMessageDetails";
+import "../layout/footer/message-center/Dialog.scss";
+
+/** Message Center Field React component.
+ * @public
+ */
+export const MessageCenterField: React.FC = () => {
```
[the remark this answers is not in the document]
AdamMeza-Bentley: are we just not reaching for arrow functions?

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
AdamMeza-Bentley: fixed!

ui/appui-react/src/appui-react/layout/footer/message-center/Message.scss
```diff
@@ -3,19 +3,15 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 @import "~@itwin/core-react/lib/cjs/core-react/style/themecolors";
```
ignas-k, line 5: If you renamed the `Message.tsx` file to `MessageCenterMessage.tsx` then you should probably rename this one too for consistency and clarity.

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,150 @@
...
+    return [...messages].reverse().map((message, index) => {
+      if (messages.length > 0) {
+        const iconClassName = MessageManager.getIconClassName(message);
+        const iconSpec = MessageManager.getIconSpecFromDetails(message);
+        const text: MessageType = message.briefMessage;
+        if ((tab === "error" && isProblemStatus(message)) || tab === "all") {
+          return (
+            <MessageCenterMessage
+              key={index.toString()}
+              message={text}
```
ignas-k, line 76: I think you should write this as `message.briefMessage`. No need to define a variable here that you use once. Especially since one line below you wrote `message.detailedMessage`. Should be consistent.

ui/appui-react/src/appui-react.ts
```diff
@@ -175,7 +175,7 @@ export * from "./appui-react/statusbar/useDefaultStatusBarItems";
 
 export * from "./appui-react/statusfields/tileloading/TileLoadingIndicator";
 export * from "./appui-react/statusfields/ActivityCenter";
-export * from "./appui-react/statusfields/MessageCenter";
+export * from "./appui-react/layout/footer/message-center/MessageCenterField";
```
ignas-k, line 178: Gerardas told me that we should keep field components in the `/statusfields` folder so you should move this back and probably add an inner folder for it like `/statusfield/messagecenter/`.

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,150 @@
...
+            </div>
+          </Tabs.Panel>
+        );
+      })}
+    </>
+  );
+
+  return (
+    <>
+      <div ref={indicatorRef} />
```
ignas-k, line 110: This should work if you move this to line 128 (pass to the `Button` as a `ref` prop. This div would not be needed then. Also if this is moved to the button, the definition on line 35 needs to be changed to `const indicatorRef = React.createRef<HTMLButtonElement>();` The variable name could also be changed but not necessarily.

ignas-k (changes requested):
Just some simple clean up suggestions.

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
AdamMeza-Bentley: that does work better!

docs/storybook/src/components/MessageCenterField.stories.tsx
```diff
@@ -4,7 +4,7 @@
  *--------------------------------------------------------------------------------------------*/
 import type { Decorator, Meta, StoryObj } from "@storybook/react";
 import { AppUiDecorator, InitializerDecorator } from "../Decorators";
-import { MessageCenterField } from "@itwin/appui-react/src/appui-react/statusfields/MessageCenter";
+import { MessageCenterField } from "@itwin/appui-react/src/appui-react/statusfields/message-center/MessageCenterField";
```
GerardasB, line 7: The visuals look great, however it would be great if we could see different states of the message center field (i.e. stories for empty, different types of messages, etc): https://itwin.github.io/appui/746/?path=/story/components-messagecenterfield--basic
Additionally it would be great to add a story which would contain some UI to manually add/clear the messages (similarly to what we have in test-app Component Examples), i.e. adding a sb decorator to one of the stories https://storybook.js.org/docs/writing-stories/decorators

AdamMeza-Bentley:
While updating the storybook I realized there was something odd going on with the notification marker. The logic I had could differentiate a cleared messages event from a message added event and would have a notifcaation pop up for a cleared messages event. Additionally there were typing issues being thrown by the iTwinUI components. Fixed this by adding a new state. I'd love some feedback on the solution I have as it feels clunky despite working as intended. @GerardasB @ignas-k 

Additionally, I realized a few things about `MessageCenterField` and wanted to ask you guys what your thoughts are on it:

- Why is `MessageCenterField` named the way it is? Would something simpler like `MessageCenter` work? 
- Should messages be able to clear themselves? For example an x button that links to the MessageManager. or even clear all button at the top of the `MessageCenterField` component.

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const determineStatus = () =>
+    messages.some((msg) => isProblemStatus(msg))
+      ? setStatus("negative")
+      : setStatus("primary");
+
+  React.useEffect(() => {
+    MessageManager.registerAnimateOutToElement(indicatorRef.current);
+
+    return MessageManager.onMessagesUpdatedEvent.addListener(() => {
+      MessageManager.messages.length > 0 ? setNotify(true) : setNotify(false);
```
ignas-k, line 66: You can move this to a `useEffect` that updates when `messages` state changes (related to next comment). Then you can remove `MessageManager` from `messages.length`.

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,152 @@
...
+      setMessages(MessageManager.messages);
+      determineStatus();
```
ignas-k, line 68: This has to be moved to a `useEffect` that updates when `messages` state changes. In `determineStatus` function you use `messages` state which is not guaranteed to be updated by the time the function starts executing (if you're not sure why, I can explain further).

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,152 @@
...
+import { MessageManager } from "../../messages/MessageManager";
+import { TitleBar } from "../../layout/footer/dialog/TitleBar";
+import type { MessageType } from "@itwin/core-react";
+import type { NotifyMessageDetailsType } from "../../messages/ReactNotifyMessageDetails";
+import "./MessageCenterField.scss";
+
+/**
+ * Type for Status state to satisfy NotificationMarker type checking
+ */
+type status = "primary" | "negative";
```
ignas-k, line 30: Rename this to something more clear e.g. `NotificationMarkerStatus`

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,152 @@
...
+  const isProblemStatus = (message: NotifyMessageDetailsType): boolean => {
+    // See priority values in DgnPlatform defined in NotificationManager
+    return (
+      message.priority === OutputMessagePriority.Error ||
+      message.priority === OutputMessagePriority.Fatal
+    );
+  };
+
+  const determineStatus = () =>
+    messages.some((msg) => isProblemStatus(msg))
```
ignas-k, line 58: Let's use full variable names when they are not too long (use `message` instead of `msg`)

ignas-k (changes requested):
Not sure why you think your solution is clunky? Seems fine to me. Just left a few suggestions.

To answer the first question: the name ends with `Field` because these components are meant for status bar and all components there are called fields (even the folder is `statusfields`). Not sure if having `Field` in the name is good but I believe we cannot change the name because it is a public component (I think the correct procedure would be to deprecate `MessageCenterField` and export the new `MessageCenter` component but we probably don't want to do that).

Can't answer the second question, sorry.

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.scss
```diff
@@ -0,0 +1,22 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+@import "~@itwin/core-react/lib/cjs/core-react/z-index";
+@import "../../layout/footer/variables.scss";
+
+.mc-footer-messageCenter-dialog {
```
GerardasB, line 8: Should be `.uifw-statusFields-messageCenter-messageCenterField`

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.scss
```diff
@@ -0,0 +1,22 @@
...
+  height: 200px;
+  overflow-y: scroll;
+  min-width: $nz-footer-popup-min-width;
+  max-width: $nz-footer-popup-max-width;
+  padding-left: 10px;
+
+  > .mc-message-prompt {
```
GerardasB, line 15:
```suggestion
  > .uifw-message-prompt {
```

docs/storybook/src/components/MessageCenterField.stories.tsx
```diff
@@ -4,7 +4,45 @@
...
 import { AppUiDecorator, InitializerDecorator } from "../Decorators";
-import { MessageCenterField } from "@itwin/appui-react/src/appui-react/statusfields/MessageCenter";
+import { MessageManager, MessageCenterField } from "@itwin/appui-react";
+import {
+  NotifyMessageDetails,
+  OutputMessagePriority,
+} from "@itwin/core-frontend";
+import { DropdownButton, MenuItem } from "@itwin/itwinui-react";
+
+const menuItems = [
```
GerardasB, line 14: Looks great. Now let's add more stories:
- Empty (instead of Basic) - renders a component w/o any messages
- Information (renders a component with a couple of information messages)
- Error (renders a component with a couple of error messages)
- Detailed (I think there could be an advanced message, i.e. w/ sublabel, let's showcase that)

Additionally, please update `onClick` handlers (this will help us test more scenarios and see if everything is working correctly):
- `Add info` should add w/o clearing messages
- `Add error` should add w/o clearing messages
- `Clear` should clear all messages

common/api/appui-react.api.md
```diff
@@ -13,7 +13,11 @@ import { AccuDraw } from '@itwin/core-frontend';
 import type { ActionButton } from '@itwin/appui-abstract';
 import { ActivityMessageDetails } from '@itwin/core-frontend';
 import { ActivityMessageEndReason } from '@itwin/core-frontend';
+<<<<<<< HEAD
+// import { BadgeType } from '@itwin/appui-abstract';
+=======
 import { BadgeType } from '@itwin/core-react';
+>>>>>>> ce27d92a4905a08bfdcde14f282ab40e9f752288
```
GerardasB, line 20:
```suggestion
import { BadgeType } from '@itwin/core-react';
```

GerardasB (changes requested):

ui/appui-react/src/appui-react/layout/footer/message-center/MessageCenterField.tsx
GerardasB: We can test if this works correctly after you update sb.

package.json
```diff
@@ -30,5 +30,11 @@
     "url": "http://www.bentley.com"
   },
   "//dependencies": "NOTE: dependencies will be empty here and instead specified in each package.",
-  "//devDependencies": "NOTE: devDependencies will be empty here and instead specified in each package."
+  "//devDependencies": "NOTE: devDependencies will be empty here and instead specified in each package.",
+  "dependencies": {
+    "@storybook/preview-api": "^8.0.0"
```
GerardasB, line 35: Revert these changes, these dependencies should only be specified in `docs/storybook/package.json`.

GerardasB:
Since this is a visual change, please add some before/after screenshots :)

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,157 @@
...
+              {getMessages(tabType)}
+            </div>
+          </Tabs.Panel>
+        );
+      })}
+    </>
+  );
+
+  return (
+    <>
```
GerardasB, line 116: Unused JSX.Fragment

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,157 @@
...
+      <Popover
```
GerardasB, line 117: I think we need a default placement. Should expand to the top and be centered w/ field if possible.

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,157 @@
...
+        style={{ width: "305px" }}
```
GerardasB, line 118: Do we need a default min height as well?

GerardasB (changes requested):

docs/storybook/src/components/MessageCenterField.stories.tsx
GerardasB: - Looks like detailed message styling is not right (it renders message and details in two columns). You can see how one looks in test-app -> Messages Tool -> Sticky Message
- Stories need some refinement: Notification, Error, Detailed should have some messages by default once the story is opened.
- Buttons to `Add Info`, `Add Error`, `Add Detailed`, `Clear` should all be visible (otherwise we can't fully test what happens when new messages of different types are added). And it should be possible to add multiple messages, I can only add 1 now. (you can actually create a separate story w/ these buttons, other stories don't need the buttons then).

docs/storybook/src/components/MessageCenterField.stories.tsx
```diff
@@ -34,4 +41,153 @@ const meta = {
...
+export const Error: Story = {
+  decorators: ErrorDecorator,
+};
+
+//Detailed Message
+
+const DetailedDecorator: Decorator = (Story) => {
+  React.useEffect(() => {
+    MessageManager.clearMessages();
+    [1, 2, 3, 4].forEach((num) => {
```
GerardasB, line 96: Why only last message color is green & others are rendered w/ black color?
Screenshot 2024-04-04 at 10 08 05

docs/storybook/src/components/MessageCenterField.stories.tsx
```diff
@@ -34,4 +41,153 @@ const meta = {
...
+            new NotifyMessageDetails(
+              1,
+              "This is the brief message",
+              "This is the detailed message"
+            )
+          );
+        }}
+      >
+        Detailed Message
+      </Button>
```
GerardasB, line 124:
```suggestion
```
`Detailed` story is the odd one out now..

docs/storybook/src/components/MessageCenterField.stories.tsx
```diff
@@ -34,4 +41,153 @@ const meta = {
...
+    <div>
+      <Story />
+      <DropdownButton menuItems={menuItems} styleType="borderless">
+        Add Messages
+      </DropdownButton>
+    </div>
+  );
+};
+
+export const Dynamic: Story = {
```
GerardasB, line 191: Open dynamic story: https://itwin.github.io/appui/746/?path=/story/components-messagecenterfield--dynamic

Steps:
- Click `Add message` -> `Info` (green bubble is rendered as excepted)
- Click `Add message` -> `Error` (red bubble is rendered as expected)
- Click on `Messages` field to open the popover (no bubble is rendered as expected)
- Click `Add message` -> `Info` (OUCH! red bubble is rendered, but green is expected)

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,152 @@
...
+            </div>
+          </Tabs.Panel>
+        );
+      })}
+    </>
+  );
+
+  return (
+    <Popover
+      style={{ width: "305px", minHeight: "600px" }}
```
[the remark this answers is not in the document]
GerardasB: Screenshot 2024-04-04 at 10 15 16

GerardasB (approved):

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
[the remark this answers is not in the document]
GerardasB: Default size is good now, however the dialog keeps growing in size once I add multiple messages (needs scrollbar/overflow).

docs/storybook/src/components/MessageCenterField.stories.tsx
GerardasB: I.e. here's how detailed message looks in the test-app currently:
Screenshot 2024-04-05 at 16 37 03

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,162 @@
...
+      }
+    }
+    return;
+  };
+
+  React.useEffect(() => {
+    MessageManager.registerAnimateOutToElement(indicatorRef.current);
+
+    return MessageManager.onMessagesUpdatedEvent.addListener(() => {
+      messages.length > 0 ? setNotify(true) : setNotify(false);
```
ignas-k, line 77: I guess this will use old (it is updated just after this) `messages` value? Is this how it is supposed to be?

GerardasB: That's a nice catch!
We got lucky, since state uses the same array reference.
```suggestion
      MessageManager.messages.length > 0 ? setNotify(true) : setNotify(false);
```

ignas-k (approved):

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,169 @@
...
+            </div>
+          </Tabs.Panel>
+        );
+      })}
+    </>
+  );
+
+  return (
+    <Popover
+      visible={isOpen}
```
mayank99, line 132: When using `visible`, you need to also use `onVisibleChange`, to avoid throwing away the built-in Popover triggers.

```suggestion
      visible={isOpen}
      onVisibleChange={setIsOpen}
```

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,169 @@
...
+              <Tabs.Tab label="Error" key="error" value="error" />
+            </Tabs.TabList>
+            {tabs}
+          </Tabs.Wrapper>
+        </>
+      }
+      applyBackground
+    >
+      <Button
+        onClick={() => setIsOpen(!isOpen)}
```
mayank99, line 151: This one shouldn't be necessary. The click handler is already wired up by Popover.
```suggestion
```

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,169 @@
...
+        ref={indicatorRef}
+        styleType="borderless"
+        labelProps={
+          <span>
+            `${messages.length} ${title}`
+          </span>
+        }
```
mayank99, line 158: This `labelProps` value looks invalid. What is the intent here?

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.scss
```diff
@@ -0,0 +1,28 @@
...
+
+.uifw-statusFields-messageCenter-popover {
+  width: 305px;
+  min-height: 200px;
+}
+
+.uifw-statusFields-messageCenter-messageCenterField {
+  width: 305px;
+  max-height: 200px;
+  overflow-y: scroll;
```
mayank99, line 16: `scroll` should pretty much never be used. I believe you meant to use [`auto`](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow#auto) here.

You can also add [`scrollbar-gutter`](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter) to prevent layout shift.

```suggestion
  overflow: auto;
  scrollbar-gutter: stable;
```

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,170 @@
...
+      }
+    }
+    return;
+  };
+
+  React.useEffect(() => {
+    return MessageManager.onOpenMessageCenterEvent.addListener(() => {
+      handleOpenChange(true);
+    });
+  });
```
mayank99, line 77: I don't know if it's intentional but these `useEffect`s are being called on every single render because they don't have a [dependency array](https://react.dev/learn/synchronizing-with-effects#step-2-specify-the-effect-dependencies)

test-apps/appui-test-app/standalone/public/locales/en-PSEUDO/AppuiTestProviders.json
```diff
@@ -0,0 +1,102 @@
+{
```
GerardasB, line 1: Remove these pseudo-localization files.

ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
```diff
@@ -0,0 +1,164 @@
+/*---------------------------------------------------------------------------------------------
```
GerardasB, line 1: Add a note to NextVersion.md

GerardasB (approved):

docs/changehistory/NextVersion.md
```diff
@@ -16,3 +16,8 @@ Table of contents:
 - `FlexWrapContainer` component in favor of [iTwinUI Flex](https://itwinui.bentley.com/docs/flex). [#795](https://github.com/iTwin/appui/pull/795)
 - `Gap` component in favor of [iTwinUI size variables](https://itwinui.bentley.com/docs/variables#size). [#795](https://github.com/iTwin/appui/pull/795)
 - `ScrollView` component in favor of [overflow property](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow). [#795](https://github.com/iTwin/appui/pull/795)
+- `Tab` component in favor of [iTwinUI Tabs](https://itwinui.bentley.com/docs/tabs). [#746](https://github.com/iTwin/appui/pull/746)
```
GerardasB, line 19: Good initiative, but remove this line. `Tab` is not actually deprecated, since `@deprecated` tag was not added to `Tab` component from `@itwin/core-react`

docs/changehistory/NextVersion.md
```diff
@@ -16,3 +16,8 @@ Table of contents:
...
+
+### Changes
+
+- `MessageCenter` has been renamed `MessageCenterField`. It now primarily uses iTwinUI components. There are a number of visual changes. [#746](https://github.com/iTwin/appui/pull/746)
```
GerardasB, line 23: We can't just rename a component that has a `@public` tag. Looking at `appui-react.api.md` hopefully that's not the case. Just mention that `MessageCenterField` was converted to a function component and mention what enhancements or visual changes it received :)

GerardasB (approved):

docs/changehistory/NextVersion.md
ignas-k: Yes, you renamed the file but not the component itself. The public component was also called `MessageCenterField` before.

GerardasB:
The dialog size is fixed now to prevent resizing when switching between `All`/`Errors` tabs.

PR #750, opened 2024-02-29
Remove status bar popup arrow

ignas-k:
## Changes
- Fixed error in `ToolbarComposer` story.
- Removed arrow from `StatusBarPopup`. The gap between the popup and field components is smaller because the arrow is not used. This makes it more consistent with other components that use the `Popup` without the arrow. Additionally, the bigger gap without the arrow makes the popup look slightly disconnected from the field component and other stuff (e.g. text) can be easily seen through the gap which is distracting.

Before:
image

After:
image

## Testing
Added a story for `MessageCenterField` which has the `StatusBarPopup`. Updated related snapshots. Tested in standalone test-app.


ui/appui-react/src/appui-react/statusbar/Popup.tsx
```diff
@@ -14,5 +14,5 @@ import { FooterPopup } from "../layout/footer/Popup";
  * @beta
  */
 export function StatusBarPopup(props: Partial<PopupProps>) {
-  return <FooterPopup {...props} />;
+  return <FooterPopup showArrow={false} {...props} />;
```
GerardasB, line 17: Ouch, `showArrow` is a public prop in `PopupProps` and used by `StatusBarPopup`...
I think we want to disallow that configurability aspect? I.e. currently someone might still add arrow to the `StatusBarPopup` since the props are spread after the `showArrow`.

ui/appui-react/src/appui-react/layout/footer/Popup.tsx
```diff
@@ -18,12 +18,12 @@ import { Popup } from "@itwin/core-react";
  * @internal
  */
 export function FooterPopup(props: Partial<PopupProps>) {
-  const { className, ...other } = props;
+  const { className, showArrow, ...other } = props;
```
GerardasB, line 21: Let's merge internal `FooterPopup` to `StatusBarPopup` (remove `FooterPopup`).

GerardasB (changes requested):
Looks good! Would you mind adding some before and after screenshots (ideally used in a test-app) to the PR since it's a visual change?
Btw, seems like there's no spacing/gap between popup and field components? (I've only checked the storybook)

ui/appui-react/src/appui-react/statusbar/Popup.tsx
ignas-k: Switched the order.

ignas-k:
@GerardasB the gap is smaller now because before the arrow was extending it. Is that a problem?

GerardasB:
> @GerardasB the gap is smaller now because before the arrow was extending it. Is that a problem?

I think we'd want to keep the gap instead of arrow. Take a look at images in the issue.

ignas-k:
As discussed, I will the leave the smaller gap for now because this makes it more consistent with other components that use the Popup without the arrow. Additionally, the bigger gap without the arrow makes the popup look slightly disconnected from the field component and other stuff (e.g. text) can be easily seen through the gap which is distracting.

GerardasB (approved):

PR #758, opened 2024-03-08
Refactor `UnitsPopup`

ignas-k:
## Changes
Changed `UnitsPopup` from a modal dialog to a simple popup. Also made some visual changes to it because it is just one Select so simplified the whole component. It is a component that is solely used in test-app so no breaking changes here.

Before:
image

After:
image

## Testing


ignas-k:
@GerardasB I also found this component that is currently a modal dialog but in my opinion this is fine because it is used for inputting a lot of data. What do you think?

image

GerardasB:
> @GerardasB I also found this component that is currently a modal dialog but in my opinion this is fine because it is used for inputting a lot of data. What do you think?
> 
> image

Yes it should be fine, I think it was added to test the dialog APIs. Let's only address status bar fields in this PR.

ui/appui-react/src/appui-react/statusfields/unitsfield/UnitsField.tsx
```diff
@@ -0,0 +1,94 @@
...
+interface UnitsFieldProps extends CommonProps {
+  label?: string;
+  title: string;
+  options: SelectOption<string>[];
+}
+
+/** Component used for setting the active unit system.
+ * @internal
+ */
+export function UnitsField(props: UnitsFieldProps) {
```
GerardasB, line 30: As discussed, let's keep this in the test-app (appui-test-providers).

GerardasB (changes requested):

ui/appui-react/src/appui-react/statusfields/unitsfield/UnitsField.tsx
ignas-k: Done

GerardasB (approved):

test-apps/appui-test-app/appui-test-providers/src/ui/dialogs/unitsfield/UnitsField.tsx
```diff
@@ -0,0 +1,100 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
```
GerardasB, line 2: nit: move outside of `dialogs` dir.

ignas-k:
@GerardasB I updated the snapshot and it seems that it got bigger by 4 pixels and the snap mode field moved slightly. Not sure why this happened as I didn't make any changes to it. Do you think this is significant?

Issue #717, opened 2024-02-01
Fix visual quirks with `SolarTimeline` component


GerardasB:
**Is your feature request related to a problem? Please describe.**
`SolarTimeline` has some visual quirks: missing gaps between UI elements. The component is not responsive and leaks elements outside of the container when shrunk.

**Describe the solution you'd like**
Convert the `SolarTimeline` from a class to a function component.
Fix visual quirks.
Re-use iTwinUI components.

PR #763, opened 2024-03-12
Use iTwinUI-React `<Select>` with native & borderless props

FlyersPh9:
## Changes

- Bump iTwinUI-React version.
- Use iTwinUI-React `<Select>` with native & borderless props, allowing to remove temporary CSS.
- Fix incorrect class name for slider placement within the grid.

## Testing

- Visually confirmed in browser.


ui/imodel-components-react/src/imodel-components-react/timeline/SolarTimeline.tsx
```diff
@@ -544,18 +543,18 @@ export class SolarTimeline extends React.PureComponent<
...
               value={currentSpeed}
-            >
-              {this._speeds.map((speed) => (
-                <option key={speed} value={speed}>
-                  {speed}x
-                </option>
-              ))}
-            </select>
+              options={this._speeds.map((speed) => ({
+                key: speed,
```
AdamMeza-Bentley, line 553: is there a reason for the key prop here? My understanding is that it's used primarily for JSX elements

ui/imodel-components-react/src/imodel-components-react/timeline/SolarTimeline.tsx
```diff
@@ -544,18 +543,18 @@ export class SolarTimeline extends React.PureComponent<
...
           </VisuallyHidden>
           <Tooltip content={this._speedLabel}>
-            <select
-              className="solar-timeline_speed"
+            <Select
+              native
+              styleType="borderless"
               name="speed"
-              onChange={(e) => this._onSpeedChange(Number(e.target.value))}
+              onChange={(newValue) => this._onSpeedChange(Number(newValue))}
```
AdamMeza-Bentley, line 550: where is `newValue` coming from? Would it be possible to convert it to a Number prior to this?

AdamMeza-Bentley (commented):
Generally looks good! small nit pick changes. Refraining from going further as you mentioned this component might not even be used

ui/imodel-components-react/src/imodel-components-react/timeline/SolarTimeline.tsx
ignas-k: I think iTwinUI `Select` already returns `newValue` as type `number` based on `options` values.  So converting it again is redundant and `Number()` can be removed.

ui/imodel-components-react/src/imodel-components-react/timeline/SolarTimeline.tsx
ignas-k: This is unnecessary and can be removed.

GerardasB: Removed `key` property.

ui/imodel-components-react/src/imodel-components-react/timeline/SolarTimeline.tsx
GerardasB: In case of a non-native Select, the value type is generic and could be a number w/o a need for the conversion. However, when `native` variant is used, the value is always typed as `string`.

GerardasB: Fixed.

GerardasB (approved):

GerardasB:
#717

PR #767, opened 2024-03-14
Custom renderer support for non-primitive properties in VirtualizedPropertyGrid

jasdom:
## Changes
Fixed custom renderers not being used for non-primitive properties in `VirtualizedPropertyGrid`.

## Testing
Added unit tests and tested in test-app.


ui/components-react/src/components-react/propertygrid/internal/flat-properties/FlatPropertyRenderer.tsx
```diff
@@ -104,17 +116,29 @@ export const FlatPropertyRenderer: React.FC<FlatPropertyRendererProps> = (
...
     // eslint-disable-next-line no-fallthrough
     case PropertyValueFormat.Struct:
+      // If a custom renderer is registered for the struct property, render it as a primitive property
+      if (hasCustomRenderer) {
+        return (
+          <PrimitivePropertyRenderer
+            highlight={highlight?.applyOnLabel ? highlight : undefined}
+            renderValueElementForNonPrimitiveProperty={true}
+            {...primitiveRendererProps}
+          />
```
grigasp, line 138: Maybe, instead of using `PrimitivePropertyRenderer` with special flag to render non-primitive properties, we could either - modify `FlatNonPrimitivePropertyRenderer` to take custom renderer into account,
- or, create a new "custom property renderer" that accepts only properties with custom renderer and uses it to render the property.

jasdom: When the struct property is null its type is set to primitive and so it is rendered as a primitive property, so only modifying `FlatNonPrimitivePropertyRenderer` would result in inconsistency depending if the property has a value or not.

Creating a new renderer that would handle all properties with custom renderers would work, but would end up being a duplicate of both `PrimitivePropertyRenderer` and `PropertyView` with the only difference of having the same effect as the current `renderValueElementForNonPrimitiveProperty` flag.

grigasp (commented):
The issue says:

> Custom renderer should be used for struct properties as well. Ideally, there should be a way to use default rendering too (e.g. if StructProperty.MemberValue = true, render custom stuff, else use default struct rendering).

Do these changes support using default rendering?

ui/components-react/src/components-react/propertygrid/internal/flat-properties/FlatPropertyRenderer.tsx
```diff
@@ -98,42 +98,43 @@ export const FlatPropertyRenderer: React.FC<FlatPropertyRendererProps> = (
...
-    property.renderer &&
-    rendererManager?.getRegisteredRenderer(property.renderer.name);
-  const typeRenderer = rendererManager?.getRegisteredRenderer(
-    property.typename
+  const customRenderer =
+    (property.renderer &&
+      rendererManager?.getRegisteredRenderer(property.renderer.name)) ||
+    rendererManager?.getRegisteredRenderer(property.typename);
+  const hasValidCustomRenderer = !!(
+    customRenderer && customRenderer.canRender(passthroughProps.propertyRecord)
```
saskliutas, line 106: Maybe `rendererManager` could have method that checks if any custom renderer applies based on property record? Something like `rendererManager.hasCustomRenderer(propertyRecord)`?

docs/storybook/src/components/PropertyGrid.stories.tsx
```diff
@@ -46,3 +54,279 @@ export const Basic: Story = {
...
+                value: "string value",
+              },
+              property: {
+                name: "simpleProperty",
+                displayLabel: "Simple property",
+                typename: "string",
+              },
+              copyWithNewValue: () => ({} as unknown as PropertyRecord),
+              getChildrenRecords: () => [],
+            },
```
saskliutas, line 145: You should create `PropertyRecord` class instance instead of constructing plain object. `new PropertyRecord({ ...value props... }, { ...property props... })`

ui/components-react/src/components-react/propertygrid/component/VirtualizedPropertyGrid.tsx
```diff
@@ -514,6 +517,16 @@ const FlatGridItemNode = React.memo(
       [onItemHeightChanged, index, node.key]
     );
 
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    React.useLayoutEffect(() => {
+      assert(divRef.current !== null);
+      const currentHeight = Math.max(divRef.current.offsetHeight, 27);
```
saskliutas, line 523: 27 probably should be defined as a constant as it used in multiple places.

ui/components-react/src/components-react/properties/ValueRendererManager.tsx
```diff
@@ -171,6 +171,17 @@ export class PropertyValueRendererManager {
     return this._propertyRenderers.get(rendererType);
   }
 
+  /** Check whether a property record has a custom renderer registered that can render it */
+  public hasCustomRenderer(record: PropertyRecord): boolean {
+    const property = record.property;
+    const customRenderer =
```
saskliutas, line 177: Looking at `selectRenderer` method it merged value renderer has higher priority than renderer based on `typename` Maybe registered custom renderer lookup could be shared between both methods? This part:

```typescript
if (
  record.property.renderer &&
  this._propertyRenderers.has(record.property.renderer.name)
)
  return this._propertyRenderers.get(record.property.renderer.name);
if (this._defaultMergedValueRenderer.canRender(record))
  return this._defaultMergedValueRenderer;
if (this._propertyRenderers.has(record.property.typename))
  return this._propertyRenderers.get(record.property.typename)!;
```

ui/components-react/src/test/propertygrid/component/internal/FlatPropertyRenderer.test.tsx
```diff
@@ -168,7 +168,107 @@ describe("FlatPropertyRenderer", () => {
...
+      selectorMatches(
+        [
+          ".components-nonprimitive-property-label-renderer",
+          ".components-property-label-renderer",
+        ].join(" ")
+      )
+    );
+  });
+
+  it("renders array as a primitive value when a custom property renderer is registered", () => {
```
saskliutas, line 196: I guess it would make more sense to check that custom renderer is used instead of checking for `.components-primitive-property-label-renderer`. Same for struct properties.

saskliutas (changes requested):

ui/components-react/src/test/propertygrid/component/internal/FlatPropertyRenderer.test.tsx
saskliutas: I think same changes should be applied when testing custom rendering based on `typename`.

jasdom: Not sure how I missed them, fixed.

ui/components-react/src/components-react/properties/renderers/PrimitivePropertyRenderer.tsx
```diff
@@ -6,76 +6,29 @@
...
-   */
+  /** Properties used for highlighting */
   highlight?: HighlightingComponentProps;
 }
 
 /** React Component that renders primitive properties
  * @public
  */
-export class PrimitivePropertyRenderer extends React.Component<PrimitiveRendererProps> {
+export class PrimitivePropertyRenderer extends CustomizablePropertyRenderer {
```
GerardasB, line 30: Prefer composition over inheritance.

ui/components-react/src/components-react/properties/renderers/CustomizablePropertyRenderer.tsx
```diff
@@ -0,0 +1,80 @@
...
+  /** Multiplier of how much the property is indented to the right */
+  indentation?: number;
+  /** Properties used for highlighting */
+  highlight?: HighlightingComponentProps;
+}
+
+/** React Component that renders customizable properties
+ * @public
+ */
+export class CustomizablePropertyRenderer extends React.Component<CustomizablePropertyRendererProps> {
```
GerardasB, line 35: Prefer functional components for new components.

GerardasB (changes requested):

saskliutas (approved):

saskliutas (approved):

Issue #766, opened 2024-03-13
Avoid exporting `@internal` APIs from the barrel file


GerardasB:
### Is your feature request related to a problem? Please describe.

Currently we are using star re-export for most of the modules which causes all of the exported APIs to be re-exported from the barrel (even ones that are tagged with `@internal`). This increases risk of consumers using internal APIs and is adding unnecessary noice when using the packages. I.e.:

```
// Module.tsx
/** @internal */
export function test() {
}

/** @public */
export function Component() {
}

---

// Current barrel file
export * from "./Module"; // OUCH: both test and Component are accessible from barrel file.

---

// Proposed change:
export { Component } from "./Module"; // test is no longer re-exported in the barrel file
```

### Describe the Solution you'd like

Update barrel files to re-export explicit APIs instead. Additional re-export modules can be created if needed i.e. https://github.com/iTwin/presentation/blob/master/packages/components/src/presentation-components/properties/editors/index.ts

### Describe alternatives you've considered

Alternatively, we could move internal APIs to a module that is not re-exported i.e. public APIs stay in `Module.tsx`, but internal ones are moved to `_Module.tsx`. Barrel file re-exports `Module.tsx` only.

PR #768, opened 2024-03-15
Fix `UiFramework.frontstages.onWidgetStateChangedEvent` event to correctly emit widget state changes

### LINKED ISSUE #757: UiFramework.frontstages.onWidgetStateChangedEvent is not being fired when widget state is changed from floating to open
### Describe the bug

When widget state changes from `WidgetState.Floating` directly o `WidgetState.Open`, UiFramework.frontstages.onWidgetStateChangedEvent is not being fired.

### To Reproduce

1. Go to a frontstage and move any widget to floating state, while keeping it open.
2. Set everything up in such a way that after clicking "Send to panel" button the widget will be sent to panel but will remain open in the panel.
3. Click "Send to panel" button.

### Expected Behavior

Widget state changes from `WidgetState.Floating` to `WidgetState.Open` and UiFramework.frontstages.onWidgetStateChangedEvent IS fired.

### Desktop

- OS: Windows 11
- Browser: electron 25.8.1
- iTwin.js AppUI Version: 4.10.0
- iTwin.js Version: 4.4.3

GerardasB:
## Changes

Fixes #757 
`UiFramework.frontstages.onWidgetStateChangedEvent` event will now correctly emit widget state changes.
Values emitted by the event will match `WidgetState` returned by `WidgetDef.state`.

## Testing

Added additional unit tests.

Tested via standalone test-app:
- Open <http://localhost:3000/?frontstage=appui-test-providers:WidgetApi>
- Open `Layout Info` widget
- Select a widget to inspect i.e. `FW-1`
- Interact with `FW-1` widget, note how `state={WidgetState}` output changes

### DIFF
--- a/ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
+++ b/ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
@@ -581,25 +581,25 @@ export class WidgetDef {
 }
 
 /** @internal */
-export function getWidgetState(widgetDef: WidgetDef, state: NineZoneState) {
-  const tab = state.tabs[widgetDef.id];
+export function getWidgetState(widgetDef: WidgetDef, nineZone: NineZoneState) {
+  const tab = nineZone.tabs[widgetDef.id];
   if (tab && tab.unloaded) {
     return WidgetState.Unloaded;
   }
 
-  if (state.draggedTab?.tabId === widgetDef.id) {
+  if (nineZone.draggedTab?.tabId === widgetDef.id) {
     return WidgetState.Closed;
   }
 
-  const toolSettingsTabId = state.toolSettings?.tabId;
+  const toolSettingsTabId = nineZone.toolSettings?.tabId;
   if (
     toolSettingsTabId === widgetDef.id &&
-    state.toolSettings?.type === "docked"
+    nineZone.toolSettings?.type === "docked"
   ) {
-    return state.toolSettings.hidden ? WidgetState.Hidden : WidgetState.Open;
+    return nineZone.toolSettings.hidden ? WidgetState.Hidden : WidgetState.Open;
   }
 
-  const location = getTabLocation(state, widgetDef.id);
+  const location = getTabLocation(nineZone, widgetDef.id);
   if (!location) {
     return WidgetState.Hidden;
   }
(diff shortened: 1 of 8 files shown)


ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -103,12 +106,10 @@ export class WidgetDef {
 
   public get state(): WidgetState {
     const frontstageDef = UiFramework.frontstages.activeFrontstageDef;
-    if (frontstageDef && frontstageDef.findWidgetDef(this.id)) {
-      const currentState = frontstageDef.getWidgetCurrentState(this);
-      // istanbul ignore else
-      if (undefined !== currentState) return currentState;
-    }
-    return this.defaultState;
+    const state = frontstageDef?.nineZoneState;
```
ignas-k, line 109: The name `state` seems a bit generic here since the property method is already called `state` and we also pass `state` to `getWidgetState` function.

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
```diff
@@ -578,3 +579,44 @@ export class WidgetDef {
     });
   }
 }
+
+/** @internal */
+export function getWidgetState(widgetDef: WidgetDef, state: NineZoneState) {
```
ignas-k, line 584: Is there a reason you want this to be exported as a separate method? It seems that it fits perfectly inside `WidgetDef` class since we need to pass an object of that type anyway.

ui/appui-react/src/test/frontstage/FrontstageDef.test.tsx
```diff
@@ -187,35 +183,96 @@ describe("FrontstageDef", () => {
...
-        sections: {
-          start: [
-            {
-              id: "test-widget",
-              defaultState: WidgetState.Hidden,
-            },
-          ],
+  describe("onWidgetStateChangedEvent", () => {
+    it("should open a hidden widget", async () => {
+      const def = new FrontstageDef();
```
ignas-k, line 188: nitpick: maybe let's use a name that is more specific e.g. `frontstageDef`. Same applies to tests below.

ignas-k (changes requested):

ui/appui-react/src/test/frontstage/FrontstageDef.test.tsx
GerardasB: Renamed.

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
GerardasB: Renamed to `nineZone`.

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
GerardasB: With #766 this way we might keep the API a bit cleaner. Long term our strategy should be aimed at deprecating these classes..

ui/appui-react/src/appui-react/widgets/WidgetDef.tsx
ignas-k: I see you changed it in `getWidgetState` but it would be good to change it here as well.

GerardasB: Renamed.

PR #773, opened 2024-03-18
Fix event listener disposal to address viewport memory leak.

hl662:
## Changes

Link to issue in backlog: https://github.com/iTwin/itwinjs-backlog/issues/1065
As part of a memory leak investigation, I'm looking at places that may have references (pointers) to viewports and whether those places have been properly disposed. One case is with event listeners and javascript closures. The handleWindowUnload() event listener is a closure, that has a reference to a screenViewport object. If the event listener is never removed, the javascript garbage collector can't dispose of variables that event listener / closure uses, including the parent object of the event listener. This leads to a memory leak, because we can't dispose of any references that the ViewportComponent uses (like the screenViewportRef as an example)

Attached is an example of a memory snapshot on chrome devtools, which shows the handleWindowUnload() still existing after multiple page refreshes (a page refresh counts as a window.unload event), and causes the viewportRef to still exist.
image

Line 258 was not removing the event listener properly - the third argument to the addEventListener, when its type is boolean
, is used to tell the event dispatcher to [prioritize](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#usecapture) this event handler before any others on unload. We can remove the event listener by changing the third argument to an object, and add the `once` opt.
## Testing
N/A, tests already exist to make sure the event listener triggers (found in FrontStage.test.tsx), but it's very unlikely we can verify that an event listener was **removed** without access to chrome DevTools within unit tests.

### DIFF
--- a/ui/imodel-components-react/src/imodel-components-react/viewport/ViewportComponent.tsx
+++ b/ui/imodel-components-react/src/imodel-components-react/viewport/ViewportComponent.tsx
@@ -255,7 +255,7 @@ export function ViewportComponent(props: ViewportProps) {
       ? screenViewportOverride
       : ScreenViewport;
     const parentWindow = parentDiv.ownerDocument.defaultView as Window;
-    parentWindow.addEventListener("unload", handleWindowUnload, true); // listener clear after being called
+    parentWindow.addEventListener("unload", handleWindowUnload, {once: true, capture: true}); // by specifying 'once', listener clears after being called
     ViewportComponentEvents.initialize();
     ViewportComponentEvents.onDrawingViewportChangeEvent.addListener(
       handleDrawingViewportChangeEvent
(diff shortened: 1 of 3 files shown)


GerardasB (approved):
Seems ok for a quick fix, but we should probably refactor. I.e. seems like this will not fix a use case, where we simply re-mount w/o unloading a page.
All the listeners should be added in `useEffect`, removed in it's cleanup fn.

ui/imodel-components-react/src/imodel-components-react/viewport/ViewportComponent.tsx
```diff
@@ -255,7 +255,7 @@ export function ViewportComponent(props: ViewportProps) {
       ? screenViewportOverride
       : ScreenViewport;
     const parentWindow = parentDiv.ownerDocument.defaultView as Window;
-    parentWindow.addEventListener("unload", handleWindowUnload, true); // listener clear after being called
+    parentWindow.addEventListener("unload", handleWindowUnload, {once: true, capture: true}); // by specifying 'once', listener clears after being called
```
GerardasB, line 258: Add a note to `NextVersion.md`

PR #776, opened 2024-03-20
Refactor translations stage 1

ignas-k:
## Changes
This is stage 1 of translations refactoring:

- removed memoization for translated strings in components.
- tried to isolate translations for each package so they don't depend on other packages. This includes moving around some translation strings between packages.
- standardized translations (where possible) in preparation for stage 2 refactoring (automatically re-render components when language is changed). Added `useTranslation` to every package and used it in functional components.

## Testing

Note: not every component could be fully fixed or tested because it might be a class component or it is memoized etc.


test-apps/appui-test-app/appui-test-providers/src/ui/frontstages/ComponentExamples.tsx
```diff
@@ -25,8 +27,8 @@ export interface ComponentExampleCategory {
  */
 export class ComponentExamplesModalFrontstage implements ModalFrontstageInfo {
   public static stageId = "appui-test-app:componentExamplesStage";
-  public title: string = UiFramework.localization.getLocalizedString(
-    "SampleApp:componentExamplesStage.examples"
+  public title = AppUiTestProviders.translate(
```
GerardasB, line 30: This is analogous to `useRef` in a functional component.
I.e. this will be called once to initialize the instance. Move this to `render`. Not adding comments for other similar cases.

test-apps/appui-test-app/appui-test-providers/src/Translation.ts
```diff
@@ -0,0 +1,22 @@
+/*---------------------------------------------------------------------------------------------
```
GerardasB, line 1: Should we rename the file to `useTranslation`?

test-apps/appui-test-app/appui-test-providers/src/Translation.ts
```diff
@@ -0,0 +1,22 @@
...
+
+interface Localization {
+  translate: (key: string) => string;
+}
+
+/**
+ * Returns a localization object.
+ * @internal
+ */
+export function useTranslation(): Localization {
```
GerardasB, line 16: Should we rename `Localization`? Or use an an implicit type (remove `Localization`, specify nothing as return type in `useTranslation() {}`

Either way, I think we need a consistent naming strategy for hooks (might be a good addition to CONTRIBUTING).
How about `useTranslation(): UseTranslationResult`. If we need additional options/arguments either list separate arguments, or if it's an object: `useTranslation(opts?: UseTranslationOptions)`

GerardasB (changes requested):

test-apps/appui-test-app/appui-test-providers/src/ui/frontstages/ComponentExamples.tsx
ignas-k: Good suggestions just not in this case as this member is inherited from `ModalFrontstageInfo` and cannot be moved. Applied this change for classes where possible though.

test-apps/appui-test-app/appui-test-providers/src/Translation.ts
ignas-k: I thought about it but was 50/50 on it. Since you mentioned it, I decided to rename it.

test-apps/appui-test-app/appui-test-providers/src/Translation.ts
ignas-k: Looking at your suggestions, for now I went with `UseTranslationResult` but I am not sure about it. Talking about consistent strategy, I do not want to fully decide something right now. We already talked about me wanting to create some code styling rules and I think this could be resolved with that because it needs some investigating and thinking to come up with a good solution (I created an issue for that whole topic #779). What do you think?

test-apps/appui-test-app/appui-test-providers/src/ui/frontstages/ComponentExamples.tsx
GerardasB: whoops, sorry about that, mistaken this for a class component :)

GerardasB (approved):

PR #781, opened 2024-03-22
Add "os" type to ColorTheme for type-safe setting of theme

GintV:
## Changes

Add "os" type to ColorTheme for type-safe setting of theme. With this change, you can use type-safe themes like so:

```
type Theme = `${ColorTheme}`;

const setColorTheme = (colorTheme: Theme) => {
    UiFramework.setColorTheme(colorTheme);
};
```


GerardasB:
There is already a public `ThemeId` type that you can use. `ThemeId` for backwards compatibility takes any string while at the same time allows us to add additional themes if needed.
There is already an alias for (previously weakly typed) "os" which is "SYSTEM_PREFERRED".

Issue #72, opened 2022-10-13
Review i18n use for missuse


raplemie:
As seen in `ui/appui-react/src/appui-react/settings/ui/UiSettingsPage.tsx`, using a ref is a mistake because it will remove the benefit of being able to change language, while always doing the work of getting the full translation on each render...

See what we want to do with that, but we probably can optimize this everywhere in the different packages.

raplemie:
Along with this check, we should also review the many different ways we have to use localization, IE, we have `UiFramework.localization` (internal) which is a pass through of `IModelApp.localization`, and we have `UiFramework.translate` (internal), which is calling directly `IModelApp.localization.getLocalizedString` (not UiFramework's...) with the namespace key... In test app we have calls to `UIFramework.localization` which is internal, we shouldnt.

We have some component that directly call `IModelApp.localization.getLocalizedString("UiFramework:...")`, not using the namespace... So we basically have multiple ways of doing the same thing, and whenever we will try to fix any of them, we will certainly forget one, so we should standardise and probably remove the unneeded ones.

PR #782, opened 2024-03-22
Add `LocalizationProvider` to replace static initializer

GerardasB:
## Changes

This PR adds `LocalizationProvider` to replace `UiCore.initialize()`. Additionally, this change enables components to correctly update translation strings when localization is changed (i.e. due to a language change). 
All usages of `UiCore.translate()` are replaced with `useTranslation` hook.
See `NextVersion.md` for a detailed explanation and usage examples.

I plan to look into `components-react` (and other  package) initializers in a separate PR. Seems like there are a couple of places, where localization is used outside of React element tree, which might complicate the static initializer deprecation, but should still be doable.

## Testing

Added additional unit tests.


GerardasB:
#72

ui/core-react/src/core-react/l10n/usePackageTranslation.tsx
```diff
@@ -0,0 +1,73 @@
...
+  namespace: string;
+  fallback: (key: string) => string | undefined;
+  defaults: object;
+}) {
+  const localization = useLocalization();
+  const [registered, setRegistered] = React.useState(false);
+  React.useEffect(() => {
+    if (!localization) return;
+
+    let ignore = false;
```
saskliutas, line 29: I think `ignore` is not needed in this case. It's ok to call `setState` on unmounted component when it is called only once. The warning about setting state in unmounted component is for cases where set state is called from some king of subscription that is not unsubscribed after unmount: https://github.com/facebook/react/pull/22114

saskliutas (approved):

test-apps/appui-test-app/standalone/src/frontend/Localization.tsx
```diff
@@ -0,0 +1,60 @@
...
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
+import { LocalizationProvider } from "@itwin/core-react";
+import { IModelApp } from "@itwin/core-frontend";
+import { Localization } from "@itwin/core-common";
+import { Language, LanguageSelect } from "@itwin/appui-test-providers";
+
+export const LanguageContext = React.createContext<
+  | {
```
ignas-k, line 12: This `|` seems unnecessary.

ui/core-react/src/core-react/searchbox/SearchBox.tsx
```diff
@@ -182,3 +175,28 @@ export class SearchBox extends React.Component<SearchBoxProps, SearchBoxState> {
...
+
+const SearchBoxInput = React.forwardRef<
+  HTMLInputElement,
+  React.ComponentProps<typeof Input>
+>(function SearchBoxInput({ placeholder, ...props }, ref) {
+  const { translate } = useTranslation();
+  return (
+    <Input
+      ref={ref}
+      placeholder={placeholder ? placeholder : translate("general.search")}
```
ignas-k, line 187:
```suggestion
      placeholder={placeholder ?? translate("general.search")}
```

ui/core-react/src/core-react/dialog/Dialog.tsx
```diff
@@ -293,76 +294,23 @@ export class Dialog extends React.Component<DialogProps> {
     }
   }
 
+  /** @deprecated in 4.11.x. Use iTwinUI Button instead. */
```
ignas-k, line 297:
```suggestion
  /** @deprecated in 4.12.x. Use iTwinUI Button instead. */
```

ui/core-react/src/core-react/dialog/Dialog.tsx
```diff
@@ -386,3 +334,77 @@ export class Dialog extends React.Component<DialogProps> {
...
+      usePrimaryStyleType = true;
+      break;
+    case DialogButtonType.Previous:
+      buttonText = translate("dialog.previous");
+      buttonClass = classnames(buttonClass, button.buttonStyle);
+      usePrimaryStyleType = true;
+      break;
+  }
+
+  buttonText = button.label ? button.label : buttonText;
```
ignas-k, line 398:
```suggestion
  buttonText = button.label ?? buttonText;
```

ui/core-react/src/core-react/searchbox/SearchBox.tsx
```diff
@@ -84,26 +83,20 @@ export class SearchBox extends React.Component<SearchBoxProps, SearchBoxState> {
           onKeyDown={this._handleKeyDown}
           onPaste={this._trackChange}
           onCut={this._trackChange}
-          placeholder={
-            this.props.placeholder
-              ? this.props.placeholder
-              : UiCore.translate("general.search")
-          }
```
ignas-k, line 91: Shouldn't placeholder be passed here as `placeholder={this.props.placeholder}`? You added the property to `SearchBoxInput` but don't pass it.

ui/core-react/src/core-react/UiCore.ts
```diff
@@ -62,27 +56,13 @@ export class UiCore {
...
-    return UiCore._localization;
-  }
-
   /** The internationalization service namespace. */
   public static get localizationNamespace(): string {
     return "UiCore";
   }
 
   /** Calls localization.getLocalizedString with the "UiCore" namespace. Do NOT include the namespace in the key.
+   * @deprecated in 4.11.x. Do not use this internally, this is replaced by `useTranslation`.
```
ignas-k, line 65:
```suggestion
   * @deprecated in 4.12.x. Do not use this internally, this is replaced by `useTranslation`.
```

ignas-k (changes requested):
Look good, just left some cleanup suggestions

ui/core-react/src/core-react/l10n/useTranslation.tsx
```diff
@@ -0,0 +1,27 @@
...
+ * @internal
+ */
+export function useTranslation() {
+  const fallback = React.useCallback((key: string) => {
+    if (!UiCore.initialized) {
+      return undefined;
+    }
+
+    // eslint-disable-next-line deprecation/deprecation
+    return UiCore.translate(key);
```
ignas-k, line 20: 1. Why did you deprecate an internal function?
2. If this method is deprecated, what is the the future plan here? Do we want to remove this fallback?
3. Why did you deprecate the translate method only in core-react and not the other packages?

ui/core-react/src/core-react/dialog/Dialog.tsx
GerardasB: I had this change initially, but wasn't sure about empty string, since operators will returns different results.

ui/core-react/src/core-react/searchbox/SearchBox.tsx
GerardasB: Same as Dialog.tsx https://www.typescriptlang.org/play?#code/GYVwdgxgLglg9mABFApgZygCgwJwFyK4xgDmiAPouACYrDErUCUiA3gFDuKIQIaIAPRAF5CUHIgD8YiQQBEtYAEMQAGyhyA3Fx58oiAJ4iZU6Qror1WnbzBo4qlADpVcEpg7duAgDSGdAL5M2gGcqBiYckoARhBywezhWDR0DMzaSZHxmogA9LmI1DDAwCg4KGBQQA

test-apps/appui-test-app/standalone/src/frontend/Localization.tsx
GerardasB: Auto formatted.

ui/core-react/src/core-react/searchbox/SearchBox.tsx
ignas-k: In that case we can use `||`. I just want this to be more clean

ui/core-react/src/core-react/searchbox/SearchBox.tsx
GerardasB: Yes, definitely, good catch!

ui/core-react/src/core-react/l10n/useTranslation.tsx
GerardasB: - I deprecated it so we wouldn't use it by accident anywhere else other than in `useTranslation` for a fallback.
- The future plan is to see if we can make corresponding changes to all our other packages. Then we can deprecate  those static initializers like `UiCore.initialize()`.
- I planned to look into other packages separately, since this PR is already getting huge

test-apps/appui-test-app/appui-test-providers/src/ui/components/LanguageSelect.tsx
```diff
@@ -4,12 +4,18 @@
...
-export function LanguageSelect() {
-  const [language, setLanguage] = React.useState("en-US");
+export type Language = "en-US" | "en-PSEUDO";
+
+export function LanguageSelect({
+  language,
+  onChange,
+}: {
+  language: Language;
+  onChange?: (newLanguage: Language) => void;
```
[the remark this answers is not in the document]
GerardasB: Good catch, passing down of `onChange` is enough.

ui/core-react/src/core-react/l10n/usePackageTranslation.tsx
GerardasB: Great suggestion, but I think we'll have to wait, since this change is only on React18, while we still support v17.
I've added this as a task item: https://github.com/iTwin/appui/issues/786

ui/core-react/src/core-react/dialog/Dialog.tsx
GerardasB: Replaced with logical OR.

ui/core-react/src/core-react/searchbox/SearchBox.tsx
GerardasB: Replaced with logical OR.

ignas-k (approved):

ui/core-react/src/core-react/l10n/usePackageTranslation.tsx
```diff
@@ -0,0 +1,73 @@
...
+    setRegistered(false);
+    void (async () => {
+      await localization.registerNamespace(namespace);
```
jason-crow, line 32: does this just become a no-op if its already registered? seems like this would get called quite a bit

GerardasB: yes, `registerNamespace` returns the same promise for each namespace.

PR #784, opened 2024-03-26
Panel stage percentage size

ignas-k:
## Changes
Added ability to set `StagePanel` size as percentage value. This involves deprecating old `size` and `minSize` getters/setters and introducing new `sizeSpec` and `minSizeSpec` getters/setters. The same was done for `StagePanelConfig` as well.

## Testing
Tested in standalone test-app. Tried with old `size`, `minSize` and `maxSize` properties, then with new properties, various combinations etc.

### DIFF
--- a/ui/appui-react/src/appui-react/layout/state/PanelState.ts
+++ b/ui/appui-react/src/appui-react/layout/state/PanelState.ts
@@ -13,15 +13,13 @@ import type {
 } from "../widget-panels/PanelTypes";
 import { isHorizontalPanelSide } from "../widget-panels/Panel";
 import type { WidgetState } from "./WidgetState";
-
-/** @internal */
-export type PanelMaxSizeState = number | { readonly percentage: number };
+import type { StagePanelSizeSpec } from "../../stagepanels/StagePanelConfig";
 
 /** @internal */
 export interface PanelState {
   readonly collapseOffset: number;
   readonly collapsed: boolean;
-  readonly maxSize: PanelMaxSizeState;
+  readonly maxSize: StagePanelSizeSpec;
   readonly minSize: number;
   readonly pinned: boolean;
   readonly resizable: boolean;
--- a/ui/appui-react/src/appui-react/layout/state/internal/PanelStateHelpers.ts
+++ b/ui/appui-react/src/appui-react/layout/state/internal/PanelStateHelpers.ts
@@ -20,7 +20,6 @@ import type { NineZoneState } from "../NineZoneState";
 import type {
   HorizontalPanelState,
   PanelsState,
-  PanelState,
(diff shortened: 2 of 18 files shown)


ui/appui-react/src/appui-react/layout/state/internal/PanelStateHelpers.ts
```diff
@@ -94,29 +94,37 @@ export function updatePanelState<K extends keyof PanelsState>(
...
 }
 
 /** @internal */
 export function getPanelSize(
-  preferredSize: number | undefined,
+  preferredSizeSpec: StagePanelSizeSpec | undefined,
   side: PanelSide,
-  minSize: PanelState["minSize"],
+  minSizeSpec: StagePanelSizeSpec,
   maxSizeSpec: PanelState["maxSize"],
```
GerardasB, line 115: Consistency
```suggestion
  maxSizeSpec: StagePanelSizeSpec,
```

ui/appui-react/src/appui-react/layout/state/PanelState.ts
```diff
@@ -13,15 +13,13 @@ import type {
...
-export type PanelMaxSizeState = number | { readonly percentage: number };
+import type { StagePanelSizeSpec } from "../../stagepanels/StagePanelConfig";
 
 /** @internal */
 export interface PanelState {
   readonly collapseOffset: number;
   readonly collapsed: boolean;
-  readonly maxSize: PanelMaxSizeState;
+  readonly maxSize: StagePanelSizeSpec;
   readonly minSize: number;
```
GerardasB, line 23: Why are we keeping `minSize`&`size` as a number while `maxSize` is a spec?

GerardasB (approved):
Looks good, thanks for adding the tests!

ui/appui-react/src/appui-react/layout/state/PanelState.ts
ignas-k: `minSize` and `size` are used directly in components for calculations or defining styling e.g. in `Panel.tsx`, `Grip.tsx`. So it makes sense to keep them as pixel values (unless we want to do more refactoring). To have these properties as pixels, `NineZoneStateReducer` converts any percentages to pixels before setting `minSize` and `size` values for PanelState. That means `minSize` and `size` will always be pixels so having these properties as `StagePanelSizeSpec` does not make sense because they will never be set as percentage values (and when we access them in components mentioned above, we would have to write `panelState.size as number` which unneccesarily complicates the code).

GerardasB: I see, we can leave it as is then.

Issue #630, opened 2023-12-11
_handleContainerComit is called to commit changed values when `Escape` button is clicked


rohitnayakbentley:
**Describe the bug**
_handleContainerComit is called to commit changed values when `Escape` button is clicked, instead onPressEscape should be called.

**To Reproduce**

1. Click on any property record 
2. Press make some change in property and press escape button,
3. Then it hits breakpoint at onPressEscape and it removes recently made changes, which is expected. 
4. But if I don’t put breakpoint at onPressEscape, and then press escape button then following code is executed, which ultimately updated property values.  [We expect to ignore changes if escape is clicked instead of updating]

**Expected behavior**
Changes from property editor record should not be committed via _handleContainerComit  when escape button is clicked. Instead it should just revert changes made and remove focus. 

**Screenshots**
![image](https://github.com/iTwin/appui/assets/94962331/dc863a6e-9abf-45ce-83ba-40c83daa79be)

![image](https://github.com/iTwin/appui/assets/94962331/22a65bd0-5a12-4bd4-99bc-a61f90e981c3)

**Desktop (please complete the applicable information):**

- OS: Windows
- Browser : Using with iTwin studio/ mango app
- Version: iTwinStudio 2023.11.27-1
- iTwin.js AppUI Version: 4.2.3
- iTwin.js Version: 4.1.6

PR #787, opened 2024-03-28
Refactor Editor Container 

AdamMeza-Bentley:
## Changes

This PR makes a number of changes to the `EditorContainer` component. 
- Migrates away from class based to functional React syntax as [outlined in the contributing docs](https://github.com/iTwin/appui/blob/master/CONTRIBUTING.md#common-issues-to-solve). 

## Testing

Component functionality has been tested in standalone and storybook. 

## Notes
This PR began by addressing a bug outlined in - Addresses the issues outline in #630.  However, after refactoring the component and attempting the solution provided it still displayed some odd behavior with the escape key event. This bug will be solved in a future PR


AdamMeza-Bentley:
I noticed the use of `React.cloneElement`. The [React Docs](https://react.dev/reference/react/cloneElement#alternatives) strong urge against using it. I'm looking into the other ways they give. Are there any prefernces? @GerardasB. 

Screenshot 2024-03-28 101055

AdamMeza-Bentley:
> > I noticed the use of `React.cloneElement`. The [React Docs](https://react.dev/reference/react/cloneElement#alternatives) strongly urge against using it. I'm looking into the other ways they give. Are there any prefernces?
> 
> Not sure if we can do much w/o breaking existing editors. I'd leave it as a separate task for later.

Ok sweet! I can make a separate issue

ui/components-react/src/components-react/editors/EditorContainer.tsx
```diff
@@ -130,78 +145,104 @@ export class EditorContainer extends React.PureComponent<EditorContainerProps> {
...
 
     return clonedNode;
-  }
+  };
 
-  private _handleEditorBlur = (_e: React.FocusEvent) => {
+  /** Event Handlers
+   * @internal
+   */
+  const handleCancel = () => onCancel();
```
GerardasB, line 153: Remove alias, use `onCancel` instead.

ui/components-react/src/components-react/editors/EditorContainer.tsx
```diff
@@ -218,91 +259,43 @@ export class EditorContainer extends React.PureComponent<EditorContainerProps> {
...
+      className="components-editor-container"
+      onBlur={handleContainerBlur}
+      onKeyDown={handleKeyDown}
+      onClick={handleClick}
+      onContextMenu={handleClick}
+      title={title}
+      data-testid="editor-container"
+      role="presentation"
+    >
+      {createEditor()}
```
GerardasB, line 298: Remove this function call, inline it in the component body instead.

ui/components-react/src/components-react/editors/EditorContainer.tsx
```diff
@@ -130,78 +145,104 @@ export class EditorContainer extends React.PureComponent<EditorContainerProps> {
...
+        newValue,
+      });
+    }
   };
 
-  private _handleContainerBlur = (e: React.FocusEvent) => {
-    e.stopPropagation();
+  const onPressEscape = (): void => {
+    // istanbul ignore else
+    if (propertyEditor && propertyEditor.containerHandlesEscape) handleCancel();
```
GerardasB, line 172: This seems to match the previous handler.
I can't seem to reproduce the #630 here https://itwin.github.io/appui/storybook/?path=/docs/components-editors--docs
What was the cause and what changed to fix it?

GerardasB (approved):

ui/components-react/src/components-react/editors/EditorContainer.tsx
AdamMeza-Bentley: DO you mean inline the whole editor creation? that seem excessive as its 30+ lines.

ui/components-react/src/components-react/editors/EditorContainer.tsx
AdamMeza-Bentley: I'm not sure to be honest. I wasn't able to recreate it either but I made the switch that was recommended

ui/components-react/src/components-react/editors/EditorContainer.tsx
```diff
@@ -91,37 +91,52 @@ export interface TypeEditor {
...
+    setFocus,
+    onCommit,
+    onCancel,
+  } = props;
+
+  let propertyEditor: PropertyEditorBase;
+
+  const editorRef = React.useRef<TypeEditor | null>(null);
+
+  const setEditorRef = (ref: TypeEditor | null) => (editorRef.current = ref);
```
ignas-k, line 111: Seems like a good place to use `React.useState`

ui/components-react/src/components-react/editors/EditorContainer.tsx
```diff
@@ -130,78 +145,104 @@ export class EditorContainer extends React.PureComponent<EditorContainerProps> {
...
-  private _handleContainerBlur = (e: React.FocusEvent) => {
-    e.stopPropagation();
+  const onPressEscape = (): void => {
+    // istanbul ignore else
+    if (propertyEditor && propertyEditor.containerHandlesEscape) handleCancel();
+  };
+
+  const onPressEnter = (e: React.KeyboardEvent): void => {
+    // istanbul ignore else
+    if (propertyEditor && propertyEditor.containerHandlesEnter) {
```
ignas-k, line 177: Maybe we can write `propertyEditor?. propertyEditor.containerHandlesEnter` for a cleaner solution? You should make this change to all similar places

ignas-k (approved):

ui/components-react/src/components-react/editors/EditorContainer.tsx
AdamMeza-Bentley: yeah thats a good catch. i'll change it in the conditionals that use that syntax

ui/components-react/src/components-react/editors/EditorContainer.tsx
GerardasB: It's 30lines + function declaration (that's used once) either way.

ui/components-react/src/components-react/editors/EditorContainer.tsx
GerardasB: Should we update the PR description then if we are not sure that it fixes the #630?
Were you able to reproduce that issue before your changes?

ui/components-react/src/components-react/editors/EditorContainer.tsx
AdamMeza-Bentley: thas a good point! I just wanted to double check before I made that change

ui/components-react/src/components-react/editors/EditorContainer.tsx
ignas-k: I am actually curious about this myself. Could you try `const [editorRef, setEditorRef] = React.useState(React.useRef<TypeEditor | null>(null));` and see if it works fine?

GerardasB: Please don't, what's wrong with the ref?

ignas-k: Nothing wrong, just thought that we could combine the two things into one

ui/components-react/src/components-react/editors/EditorContainer.tsx
AdamMeza-Bentley: I changed the PR description to address that. I did try the suggested change but it wasn't successful

ui/components-react/src/components-react/editors/EditorContainer.tsx
```diff
@@ -218,91 +216,77 @@ export class EditorContainer extends React.PureComponent<EditorContainerProps> {
...
+      className="components-editor-container"
+      onBlur={handleContainerBlur}
+      onKeyDown={handleKeyDown}
+      onClick={handleClick}
+      onContextMenu={handleClick}
+      title={title}
+      data-testid="editor-container"
+      role="presentation"
+    >
+      {() => {
```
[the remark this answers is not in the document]
AdamMeza-Bentley: fixed this.

Issue #785, opened 2024-03-26
Active viewport border is not visible


raplemie:
### Describe the bug

When there are multiple viewport, a brown bar usually appear at the top of the active viewport to indicate that it is active. In our case, it works well in our widgets since we use `FloatingViewportContent`. However, we no longer see the brown bar in the main view.

After some investigation, it turns out that the bar is there, but it is hidden behind the docked tool settings.

Note that pinning the top panel causes the bar to appear as expected.

### To Reproduce

Have multiple views in a frontstage
Have docked tool settings displayed
Notice that the top bar is not visible.

### Expected Behavior

The top bar should always be visible when the view is active, regardless of tool settings and pinned top panel state.

### Screenshots

https://github.com/iTwin/appui/assets/1904889/a1d5851b-bb4e-40ae-956e-19fe29268b8f

### Desktop (please complete the applicable information)

AppUI 4.10.0

### Additional context

This issue was likely introduced by #577, or #705, sorry.
We are not using any of the preview features yet.

ignas-k:
Fixed in #793.

PR #793, opened 2024-04-04
Fix standard layout to prevent content overlays

GerardasB:
## Changes

This PR fixes an issue where tool settings and status bar would overlay the content (i.e. viewport).
This would cause certain negative side effects like content view-overlay hiding below tool settings or imodeljs-icon hiding below status bar.

  | Before  | After |
  | ------------- | ------------- |
  | Screenshot 2024-04-04 at 17 56 18 | Screenshot 2024-04-04 at 17 56 38 |

This unintentional behavior was first introduced with #736 (released in [4.10.0](https://github.com/iTwin/appui/releases/tag/release%2F4.10.0)).

## Testing

Tested via standalone test-app.

### DIFF
--- a/common/changes/@itwin/appui-react/fix-tool-settings-overlay_2024-04-04-14-47.json
+++ b/common/changes/@itwin/appui-react/fix-tool-settings-overlay_2024-04-04-14-47.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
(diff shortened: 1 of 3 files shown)


wilmaier:
I tested this change in pineapple and it works well.

wilmaier (approved):
Works well with the change

ignas-k (approved):
Nice, I think this fixes #785. I will test it.

PR #803, opened 2024-04-10
React 18 bugs and type fixes

MichaelBelousov:
fix usages that break in react 18 including fixing somethings that stricter types revealed.
Also contains some FIXME comments instead of an actual fix due to possible API breaks


ui/appui-react/src/appui-react/layout/widget/Button.tsx
```diff
@@ -8,9 +8,11 @@
 
 import * as React from "react";
 import { IconButton } from "@itwin/itwinui-react";
+import { PolymorphicForwardRefComponent } from "@itwin/itwinui-react/cjs/core/utils";
```
ignas-k, line 11: Is this needed?

ignas-k (approved):

ui/appui-react/src/appui-react/layout/widget/Button.tsx
GerardasB: Removed

PR #806, opened 2024-04-22
Deprecate UiEvent subclasses

ignas-k:
## Changes
Deprecated all UI event classes because they are only emitter classes and therefore should not be exported. I did not create separate properties to replace current public properties that use these event classes as type because users really should have no use for these classes (except `instanceof UiVisibilityChangedEvent` but they should see the deprecation for this use case and apply the suggested change `instanceof BeUiEvent<UiVisibilityEventArgs>`). Because behaviour does not change and users should not really be using these classes anyway (instanceof is a questionable use case for them), we can just change public properties to use `BeUiEvent` types in AppUI 6. 

## Testing
N/A

### DIFF
--- a/common/api/appui-react.api.md
+++ b/common/api/appui-react.api.md
@@ -176,7 +176,7 @@ export interface AccuDrawDialogProps extends CommonProps {
     orientation?: Orientation;
 }
 
-// @beta
+// @beta @deprecated
 export class AccuDrawGrabInputFocusEvent extends BeUiEvent<{}> {
 }
 
@@ -205,7 +205,7 @@ export class AccuDrawPopupManager {
     static showMenuButton(id: string, el: HTMLElement, pt: XAndY, menuItemsProps: CursorMenuItemProps[]): boolean;
 }
 
-// @beta
+// @beta @deprecated
 export class AccuDrawSetCompassModeEvent extends BeUiEvent<AccuDrawSetCompassModeEventArgs> {
 }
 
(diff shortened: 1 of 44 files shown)


GerardasB (approved):
LGTM!

ui/components-react/src/components-react/tree/controlled/TreeEvents.ts
```diff
@@ -26,42 +26,48 @@ import type { Observable, Subscription } from "./Observable";
...
   onCheckboxStateChanged?(
-    event: TreeCheckboxStateChangeEventArgs
+    event: TreeCheckboxStateChangeEventArgs // eslint-disable-line deprecation/deprecation
   ): Subscription | undefined;
 }
 
 /**
  * Data structure that describes tree node event payload.
  * @public
+ * @deprecated in 4.13.x. Event args are inferred from a listener. If explicit type is needed use a type helper.
```
saskliutas, line 70: I don't think these `Tree*EventArgs` types should be deprecated. They are not used with `BeEvent` or BeUIEvent` to describe listeners args. They are used as parameter types for functions of `TreeEvents`.

saskliutas (changes requested):

ui/components-react/src/components-react/tree/controlled/TreeEvents.ts
ignas-k: Reverted

GerardasB:
Event args were deprecated as well, since that would allow us to add additional required properties if needed. Example type helper: `type ListenerType<TEvent extends BeEvent<Listener>> = TEvent extends BeEvent<infer TListener> ? TListener : never;`

PR #810, opened 2024-04-23
Fix rush audit GHSA-8cp3-66vr-3r4c

GerardasB:
## Changes

Remove unused and undocumented usage of `@itwin/core-mobile` to fix https://github.com/advisories/GHSA-8cp3-66vr-3r4c
Mobile specific parts of the UI can be tested w/o running a mobile app backend.

Deprecate `UiFramework.isMobile()` in favor of `ProcessDetector`.

## Testing

N/A


ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.tsx
```diff
@@ -147,6 +147,7 @@ export class ToolAssistanceField extends React.Component<
   constructor(p: ToolAssistanceFieldProps) {
     super(p);
 
+    // eslint-disable-next-line deprecation/deprecation
     const mobile = UiFramework.isMobile();
```
ignas-k, line 151: Seems like `UiFramework.isMobile()` is just `ProcessDetector.isMobileBrowser`. Can't we just replace it with that here and in other places? We won't need to fix in the future and we won't need `eslint-disable` then.

docs/changehistory/NextVersion.md
```diff
@@ -3,13 +3,18 @@
...
   - [Changes](#changes)
+- [@itwin/core-react](#itwincore-react)
+  - [Deprecations](#deprecations-1)
+  - [Changes](#changes-1)
 
 ## @itwin/appui-react
 
+### Deprecations
+
+- `UiFramework.isMobile()` method in favor of `ProcessDetector.isMobileBrowser()` from `@itwin/core-bentley`. [#810](https://github.com/iTwin/appui/pull/810)
```
ignas-k, line 16:
```suggestion
- `UiFramework.isMobile()` method in favor of `ProcessDetector.isMobileBrowser` from `@itwin/core-bentley`. [#810](https://github.com/iTwin/appui/pull/810)
```
Seems like it is just a getter.

ignas-k (changes requested):

ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.tsx
GerardasB: Replaced the calls.

ignas-k (approved):

Issue #612, opened 2023-12-01
Widget with `ThemeProvider` cannot pop content outside of floating widget


raplemie:
With popout rendering change, or with iTwinUI 3.0 content, widgets that uses popouts (Select component for example) now need to include a ThemeProvider within the body of their widget, however, when doing so, the popout is rendered "within" the widget when it is floating, and the poped out content is no longer expanding beyond the widget.

It would be nice to:
* Not have to include `ThemeProvider` for widgets that are using AppUI provided iTwinUI version.
  We initialy thought of providing directly, but because of the overflow issue, we decided to not do it.
* If a `ThemeProvider` is used, the popout would be rendered as if there was none
  Initial investigation shows that `.nz-widget-widget` class is responsible for the `overflow: hidden` rule, we need to investigate if this is required.

Visual of the issue.
No provider: 
<img width="824" alt="image" src="https://github.com/iTwin/appui/assets/1904889/10033528-e0e9-4478-8ea4-0a4de1c8233b">

With provider:
<img width="903" alt="image" src="https://github.com/iTwin/appui/assets/1904889/4f5bdbd5-25ba-4a95-a4fa-4a7882f02c17">

GerardasB:
This doesn't seem to be an issue with iTwinUI 3.0, because a [portalContainer](https://github.com/iTwin/iTwinUI/pull/1477) can control where the floating elements are appended.

PR #820, opened 2024-05-03
Add `reparentPopoutWidgets` preview feature

GerardasB:
## Changes

This PR adds the `reparentPopoutWidgets` preview feature which will reparent widget content to a popout content container when enabled.

## Testing

Manual testing with a standalone test app.

### DIFF
--- a/ui/appui-react/src/appui-react/layout/widget/PopoutWidget.tsx
+++ b/ui/appui-react/src/appui-react/layout/widget/PopoutWidget.tsx
@@ -9,14 +9,14 @@
 import * as React from "react";
 import * as ReactDOM from "react-dom";
 import { WidgetIdContext } from "./Widget";
-import { useContainersStore } from "./ContentManager";
 import { WidgetContentContainer } from "./ContentContainer";
+import { usePopoutsStore } from "../../childwindow/usePopoutsStore";
 
 /** @internal */
 export function PopoutWidget() {
   const widgetId = React.useContext(WidgetIdContext);
-  const popoutContainer = useContainersStore((state) =>
-    widgetId ? state.popoutContainers[widgetId] : undefined
+  const popoutContainer = usePopoutsStore((state) =>
+    widgetId ? state.popouts[widgetId] : undefined
   );
   if (!popoutContainer) return null;
   return ReactDOM.createPortal(<WidgetContentContainer />, popoutContainer);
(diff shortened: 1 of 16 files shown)


ui/appui-react/src/appui-react/widget-panels/Content.tsx
```diff
@@ -51,12 +60,37 @@ export function WidgetContent() {
...
-      <ErrorBoundary FallbackComponent={WidgetFallback}>
-        {widget?.reactNode}
-      </ErrorBoundary>
-    </ScrollableWidgetContent>
+    // Theme providers are required to open floating/popover elements in a popout widget window (instead of a main window).
+    <ThemeProvider
+      portalContainer={popoutContainer ?? undefined}
+      className="uifw-widgetPanels-content_themeProvider"
+    >
+      <ThemeProviderV2
```
GerardasB, line 83: There is an issue with this: https://github.com/iTwin/appui/issues/612
Should we remove `ThemeProviderV2` and add a prereq: compatible with iTwinUI v3 only? Looking around if there's another option..

ui/appui-react/src/appui-react/layout/widget/PopoutWidget.tsx
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+import { WidgetIdContext } from "./Widget";
+import { WidgetContentContainer } from "./ContentContainer";
+import { usePopoutsStore } from "../../childwindow/usePopoutsStore";
+
+/** @internal */
+export function PopoutWidget() {
```
GerardasB, line 16: Rename as there are 2 components with the same name.

ui/appui-react/src/appui-react/widget-panels/Content.tsx
ignas-k: Would that mean that the new popout widgets would only work with iTwinUI v3? But old ones would work fine with v2?

ignas-k (approved):

ui/appui-react/src/appui-react/layout/widget/PopoutWidget.tsx
GerardasB: Renamed.

ui/appui-react/src/appui-react/widget-panels/Content.tsx
GerardasB: Just FYI I had some success, simply wrapping each widget content in a v2 ThemeProvider, however, that involved re-parenting of ThemeProvider to the child window and would probably lead to other issues...

ignas-k: Sounds good. Anyway, how long do we need to support iTwinUI v2 for?

PR #821, opened 2024-05-06
Use iTwinUI labels for `StatusBar` and `ToolSettings`

ignas-k:
## Changes
In this PR I used iTwinUI for `StatusBar` and `ToolSettings`. This mainly means iTwinUI labels and scss variables. Additionally, I added a DockedBar internal component which wraps both bars now so their size, background color, borders etc. are the same. Furthermore, I decreased div nesting and removed unused/duplicate css settings and adjusted some gaps/spacing in `StatusBar` and `ToolSettings` components. Noticeable visual changes are iTwinUI labels (so labels are now semi-bold) and spacing adjustments: 

### 1. Docked tool settings:
Before
image
After
image

### 2. Floating tool settings 
Before
image
After
image

### 3. Narrow floating tool settings
Before
image
After
image

### 4. Status bar
Before
image
After
image

## Testing


ui/appui-react/src/appui-react/widget-panels/DockedBar.scss
```diff
@@ -2,10 +2,20 @@
...
+
+  &[data-placement="top"] {
+    border-block-end: 1px solid var(--iui-color-border);
+  }
+
+  &[data-placement="bottom"] {
+    border-block-start: 1px solid var(--iui-color-border);
+  }
+
+  label {
```
ignas-k, line 18: @GerardasB  what do you think about having styling for all labels here. On one hand I am not a big fan of putting styling on all elements of label type, however, it is the easiest way to ensure that all labels within `StatusBar` and `ToolSettings` look the same. Besides, labels should only be used to describe input elements and nothing else so maybe it is fine to have this styling. What do you think?

ui/appui-react/src/appui-react/layout/tool-settings/Docked.tsx
```diff
@@ -138,7 +138,7 @@ export function DockedToolSettings(props: DockedToolSettingsProps) {
   const PanelContainer = props.panelContainer
     ? props.panelContainer
     : DefaultPanelContainer;
-  const className = classnames("nz-toolSettings-docked", props.className);
+  const className = classnames("uifw-toolSettings-docked", props.className);
```
ignas-k, line 141: @FlyersPh9  what do you think about deprecating `UiFramework.animateToolSettings`? I noticed that tool settings have some optional animations but personally I don't see any benefit from it for a few reasons: it happens only when the window is resized and this is the only place (lines 152-166 in this file) where such animations are used, so is it really needed at all?

ui/appui-react/src/appui-react/statusfields/SelectionScope.tsx
```diff
@@ -55,9 +55,9 @@ function SelectionScopeFieldComponent(props: SelectionScopeFieldProps) {
       )}
       style={props.style}
     >
-      <label className="uifw-statusFields-selectionScope-label">
+      <Label className="uifw-statusFields-selectionScope-label">
```
ignas-k, line 58: @FlyersPh9  I also noticed something about iTwinUI labels when working on this item. It looks like label text and text inside the input is not vertically aligned. For example, in the screenshot I used `LabeledInput` without any additional styling (except I disabled the semi-bold font so it's easier to compare). So both label and input are from iTwinUI and they're wrapped together by iTwinUI (although the same thing happens if I use `Label` and `Input` separately). You can see that the label is slightly lower than the text inside the input. This is also the case in [Ladle docs](https://itwin.github.io/iTwinUI/react/?story=label--inline). However, in [these docs](https://itwinui.bentley.com/docs/input#inline) the label and input text seems to align properly. Could you take a look?

P.S. it aligns if I wrap the `LabeledInput` in a div with `display:flex` and odd height e.g. `height: 35px`.

image

ui/appui-react/src/appui-react/widget-panels/DockedBar.scss
GerardasB: Let's avoid that since it's generally a bad idea to target an element you are not in control.

ignas-k: Alright, I moved the margin to `uifw-default-label` class so at least all labels in tool settings will be the same.

GerardasB (approved):

ui/appui-react/src/appui-react/layout/tool-settings/Docked.tsx
FlyersPh9: Unfortunately I think this needs to stay, at least for now.  This animation was added because users were not noticing the tool settings changing when switching between tools.  Even with the animation I've still heard of some users not noticing the tool settings changing.

Personally, I'm of the belief that a user may not notice this their first couple times using the application but will learn it over time making the animation no longer necessary.  Perhaps onboarding or something else could teach the user where tool settings are?

ignas-k: I see, it makes sense if the animation is supposed to help users notice the tool settings. But I also agree with your point that users would easily learn where the to find the tool settings on their own. And when the users know where to find them, the animation becomes basically useless. We can leave this as it is for now.

PR #848, opened 2024-05-22
Refactor `StatusBar`

ignas-k:
## Changes

- Cleaned up css of `StatusBar` and main field components. Removed smaller font on smaller screens for these components as well. Added expand indicators to buttons with popovers.
- Removed/Cleaned up `StatusBar` field components. Switched field indicators which were custom AppUI buttons to proper iTwinUI buttons.
- Started using iTwinUI `Popover` instead of AppUI `Popup` for status bar field components.
- Deprecated `StatusBar`, `StatusBarLabelIndicator`, `StatusBarIndicator`. These are very simple components that the user can replicate themselves.
- Added `StatusBarPopover` which is a simple wrapper for iTwinUI `Popover` but it helps us control the offset of the popover so it will be consistent for all status bar fields.
- Added `StatusBarPopover.ExpandIndicator` which the users can add to their buttons to indicate that is has expandable content (popover).
- Slightly improved the status bar overflow panel. Added some formatting and gaps between fields. Also using the iTwinUI `Popover` fixed the issue where components with popovers would not work if they were in the overflow panel.

**Old:**

image

image

**New:**

image

image

## Testing


ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.scss
```diff
@@ -16,34 +18,18 @@ $key-size-medium: 18px;
...
-
-  @include for-tablet-portrait-down {
-    font-size: $font-size-medium;
-  }
-
-  @include for-phone-only {
-    font-size: $font-size-small;
-    width: 12em;
-  }
+.temp {
```
ignas-k, line 21: Rename and update when styling is clear.

ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.tsx
```diff
@@ -541,15 +523,25 @@ export class ToolAssistanceField extends React.Component<
...
             {dialogContent}
           </ToolAssistanceDialog>
-        </StatusBar.Popup>
-      </>
+        }
+      >
+        <Button
+          styleType="borderless"
+          startIcon={<>{toolIcon}</>}
+          className={classnames("temp", this.props.className)}
```
ignas-k, line 531: @FlyersPh9 I have some questions about `ToolAsisstanceField` styling. 
1. First of all, it has a little triangle at the bottom right corner which I suppose symbolizes that it is expandable. I see the same concept used in toolbars. However, I was wondering why this is the only component in `StatusBar` that uses this triangle even though many other are also button that open a popup e.g. `MessageCenterField`, `SnapModeField`. Do we want to keep it then?
2. Should this field have a constant size (like it is now)? I suppose this was done so the other `StatusBar` fields do not move when changing the active tool (since names of tools are of different length).
3. Also this field had some additional styling for smaller screens e.g. the width of this field would decrease at smaller screens and the text would wrap into multiple lines. I would like to get rid of the wrapping because all other field components have text without wrapping so that would be just consistent. Not too sure about the width getting smaller on smaller screens though. What do you think?
image

FlyersPh9: For the 2nd question: I think we should have a consistent width to prevent content shift.  I don't know how that width is set currently, but we should set the width of the label to use the `ch` unit.

For the 1st and 3rd, can we take a look in a call?

ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.scss
ignas-k: Fixed.

ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.scss
```diff
@@ -16,33 +16,14 @@ $key-size-medium: 18px;
...
-      margin-right: 4px;
-    }
+.uifw-statusFields-toolAssistance-toolAssistanceField {
+  width: 30ch;
+  justify-content: flex-start;
+
+  .prompt {
+    white-space: nowrap;
+    overflow: hidden;
+    text-overflow: ellipsis;
```
ignas-k, line 26: @GerardasB Jon and I decided to show ellipsis for overflown text here. That means that some prompts will never be fully displayed because they are too long (although they will be fully displayed in the opened popover). Do you think this is fine? We also considered leaving comment somewhere saying that instructions above certain length will not be fully displayed in the `ToolAssistanceField` but I did not find a good place because the `ToolAssistance.createInstrunction` seems to be in `appui-abstract`.

ui/appui-react/src/appui-react/statusbar/popup/StatusBarPopover.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * @module StatusBar
+ */
+
+import * as React from "react";
+import { Popover } from "@itwin/itwinui-react";
+import { ButtonExpandIndicator } from "./ExpandIndicator";
+
+/** Popover component used in [[StatusBar]] component.
+ * This component should wrap the element that triggers the popover.
+ * @note Add the {@link StatusBarPopover.ExpandIndicator} to popover trigger buttons.
```
ignas-k, line 15: @GerardasB do you think it's ok to have a note like this? I want to encourage users to use the expand indicator but not sure what is the best way to do that.

ui/appui-react/src/appui-react/statusbar/popup/StatusBarPopover.tsx
```diff
@@ -0,0 +1,39 @@
...
+        offset: 4,
+      }}
+    />
+  );
+}
+
+/** Components used in StatusBarPopover.
+ * @public
+ */
+export namespace StatusBarPopover {
```
GerardasB, line 34: Do we need a separate namespace definition? Can we assign additional properties on the function itself i.e. `StatusBarPopover.ExpandIndicator = ButtonExpandIndicator;`

docs/changehistory/NextVersion.md
```diff
@@ -2,11 +2,34 @@
...
+## @itwin/appui-react
+
+### Deprecations
+
+- Deprecated `StatusBar` component in favor of `StatusBarComposer`. [#848](https://github.com/iTwin/appui/pull/848)
+- Deprecated `StatusBar` namespace. Please use the components directly. [#848](https://github.com/iTwin/appui/pull/848)
+  - Deprecated `StatusBar.Popup` and `StatusBarPopup` in favor of `StatusBarPopover`.
+  - Deprecated `StatusBar.Field` and `StatusBarField` in favor of [iTwinUI Button](https://itwinui.bentley.com/docs/button).
+- Deprecated `StatusBarIndicator` in favor of [iTwinUI Button](https://itwinui.bentley.com/docs/button) (or other components) and AppUI `StatusBarPopup`. [#848](https://github.com/iTwin/appui/pull/848)
+- Deprecated `StatusBarLabelIndicator` in favor of [iTwinUI Label](https://itwinui.bentley.com/docs/typography#label) and AppUI `Icon`. [#848](https://github.com/iTwin/appui/pull/848)
```
GerardasB, line 24: Let's not advocate the `AppUI` `Icon` which is mainly for handling those outdated `IconSpec` types internally. Instead, just point to iTwinUI react package components. (applies to component documentation as well)

docs/changehistory/NextVersion.md
```diff
@@ -2,11 +2,34 @@
...
 
+## @itwin/appui-react
+
+### Deprecations
+
+- Deprecated `StatusBar` component in favor of `StatusBarComposer`. [#848](https://github.com/iTwin/appui/pull/848)
+- Deprecated `StatusBar` namespace. Please use the components directly. [#848](https://github.com/iTwin/appui/pull/848)
+  - Deprecated `StatusBar.Popup` and `StatusBarPopup` in favor of `StatusBarPopover`.
+  - Deprecated `StatusBar.Field` and `StatusBarField` in favor of [iTwinUI Button](https://itwinui.bentley.com/docs/button).
+- Deprecated `StatusBarIndicator` in favor of [iTwinUI Button](https://itwinui.bentley.com/docs/button) (or other components) and AppUI `StatusBarPopup`. [#848](https://github.com/iTwin/appui/pull/848)
```
GerardasB, line 23: `StatusBarPopup` is deprecated, just point to `StatusBarPopover` directly.

docs/changehistory/NextVersion.md
```diff
@@ -2,11 +2,34 @@
...
+- Deprecated `StatusBarLabelIndicator` in favor of [iTwinUI Label](https://itwinui.bentley.com/docs/typography#label) and AppUI `Icon`. [#848](https://github.com/iTwin/appui/pull/848)
+
+### Additions
+
+- Added `StatusBarPopover` to replace `StatusBar.Popup`. The component uses [iTwinUI Popover](https://itwinui.bentley.com/docs/popover) with some consistent styling for all `StatusBar` fields. `StatusBarPopover` should wrap the element that triggers it (e.g. a button). [#848](https://github.com/iTwin/appui/pull/848)
```
GerardasB, line 28: Mention the addition, usage and use-cases of `StatusBarPopover.ExpandIndicator`.

ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.scss
GerardasB: Is it only me, or do we need to add some additional spacing to the expand indicator?

docs/changehistory/NextVersion.md
ignas-k: I mentioned it below already. I added some examples now

docs/changehistory/NextVersion.md
ignas-k: My mistake, I wrote the wrong thing. Fixed now.

ui/appui-react/src/appui-react/statusbar/popup/StatusBarPopover.tsx
ignas-k: I tested it. If I write `StatusBarPopover.ExpandIndicator = ButtonExpandIndicator;` it will still export `StatusBarPopover` as an undocumented namespace so it seems like there is no way around it.

ui/appui-react/src/appui-react/statusbar/popup/StatusBarPopover.tsx
ignas-k: I switched to using ` `` ` because I could not get the docs to work properly.

ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.scss
ignas-k: It still has a capped size. On line 20 the width is set to a static value.

Do you mean additional spacing away from the sides or away from the content of the button?

docs/changehistory/NextVersion.md
ignas-k: I suggested using `AppUI` `Icon` for two reasons that are closely related. The first one is that the deprecated `StatusBarLabelIndicator` used `AppUI` `Icon` and the second one is that our `Icon` adds some additional styling. For example, if we use iTwinUI `Icon` or just svgs directly it will have a muted color which is not what we want.

GerardasB (approved):

ui/appui-react/src/appui-react/statusbar/popup/StatusBarPopover.tsx
```diff
@@ -0,0 +1,39 @@
...
+import * as React from "react";
+import { Popover } from "@itwin/itwinui-react";
+import { ButtonExpandIndicator } from "./ExpandIndicator";
+
+/** Popover component used in `StatusBar` component.
+ * This component should wrap the element that triggers the popover.
+ * @note Add the `StatusBarPopover.ExpandIndicator` to popover trigger buttons.
+ * @public
+ */
+export function StatusBarPopover(props: React.ComponentProps<typeof Popover>) {
```
GerardasB, line 18: I'd prefer to keep the props type "smaller", we can add additional properties if needed.

ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.scss
ignas-k: BTW I could increase the spacing by moving the triangle closer to the edge by 1px. My only concern that it might be harder to notice when closer to the edge (although that may not be a bad thing). I cannot increase the spacing in another way because this is a border with absolute positioning. The position closer to the edge is what was used by `ToolAssistanceField` before so maybe we should use that. What do you think?

Current:
image

Moved (and used by `ToolAssistanceField` before:
image

GerardasB: I see what you mean, that's why I asked - in case visual change like this is needed in the future +@FlyersPh9

ui/appui-react/src/appui-react/statusbar/popup/StatusBarPopover.tsx
ignas-k: I added a props interface with the required props.

ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.scss
ignas-k: @FlyersPh9 Jon, what do you think? Should the indicator be closer to the edge or not?

ui/appui-react/src/appui-react/statusbar/StatusBarComposer.tsx
```diff
@@ -355,12 +386,18 @@ export function StatusBarComposer(props: StatusBarComposerProps) {
   const handleEntryResize = React.useCallback(
     (key: string) => (w: number) => {
       const oldW = entryWidths.current.get(key);
-      if (undefined === oldW || Math.abs(oldW - w) > 2) {
+      if (oldW === undefined || Math.abs(oldW - w) > 2) {
         entryWidths.current.set(key, w);
+        if (oldW === undefined) {
+          const widths = verifiedMapEntries(entryWidths.current);
+          if (widths !== undefined)
+            entryWidths.current = sortWidths(widths, statusBarItems);
```
ignas-k, line 394: @GerardasB could you take a look at `fix overflow order` commit? I managed to reproduce the issue of items going to overflow in the wrong order. I reproduced it by adding a new item provider after a timeout. My fix is basically sorting the widths every time a new status bar item is added. It is not a clean solution but this is the best I could do since the code is very interconnected and old (and needs a refactoring but I don't want to do that right now because this file is >500 lines). I tried a few different ways to fix this issue e.g. resetting `entryWidths`, sorting in a `useMemo` but none of them worked that well.

GerardasB: Seems good and might be worth mentioning in the `NextVersion`.
As discussed - long term we should have a single place where we handle overflow logic, since we have multiple components that need it.

ui/appui-react/src/appui-react/statusfields/toolassistance/ToolAssistanceField.scss
FlyersPh9: That 1px change is hardly even noticeable.  I think I prefer current over moved example because the bottom of the triangle aligns with the bottom of the dividing lines.

PR #849, opened 2024-05-22
Fix property grid text selection

FlyersPh9:
## Changes

- Moved `user-select: none` from property grid to element separator to allow a user to highlight the text values.
- Add `tabindex: -1` to element separator so it is not keyboard focusable. It already had `:focus { outline: 0; }` which hides the focus, this now disables focus.
- Changed `.components-property-label-renderer-colon` from a `<span>` to an `::after`.

## Testing

- Confirmed that resizing columns does not unintentionally highlight text.
- Used the <kbd>tab</kbd> to confirm that the resizer is no longer taking focus.

### DIFF
--- a/ui/components-react/src/components-react/properties/renderers/PropertyView.tsx
+++ b/ui/components-react/src/components-react/properties/renderers/PropertyView.tsx
@@ -59,11 +59,14 @@ export class PropertyView extends React.Component<
   };
 
   private _onContextMenu = (e: React.MouseEvent) => {
-    if (this.props.onContextMenu)
+    if (this.props.onContextMenu) {
       this.props.onContextMenu(this.props.propertyRecord, e);
-    if (this.props.onRightClick)
+      e.preventDefault();
+    }
+    if (this.props.onRightClick) {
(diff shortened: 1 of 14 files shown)


saskliutas (approved):

ui/components-react/src/components-react/properties/renderers/PropertyView.tsx
```diff
@@ -113,6 +116,7 @@ export class PropertyView extends React.Component<
         {needElementSeparator ? (
           // eslint-disable-next-line deprecation/deprecation
           <ElementSeparator
+            style={{ margin: 0 }}
```
ignas-k, line 119: Looks good. This is needed so it doesn't apply the negative margins? This is just what I gathered from storybook. Ignore if it doesn't make sense

ignas-k (approved):

ui/components-react/src/components-react/properties/renderers/PropertyView.tsx
GerardasB: Yes, I was hesitant to change it at element separator level due to unknown consequences.

PR #860, opened 2024-06-03
Open 'pw' scheme links in a new window

### LINKED ISSUE #857: Make model source links in PropertyGrid open to a new tab, rather than modifying the current window's href.

image

After selecting an element that has a source path in the viewer, the property grid will show at the bottom a tab like above. If you click on the path, the link handler within propertyGrid tries to modify the current window's [url](https://github.com/iTwin/appui/blob/050ebc9c832e2b6c8d8aaaef4ef444d82e77ff18/ui/components-react/src/components-react/propertygrid/component/PropertyGridCommons.ts#L138). This will conflict with a behavior within the viewer where it will teardown the IModelApp when it detects a possible window redirect. This is done to avoid memory leaks when refreshing the window that has the viewer open.

I'd like it if clicking on the path results in a new tab being opened, rather than the window with the viewer within changing its href.

### Describe alternatives you've considered

As documented in this [PR](https://github.com/iTwin/viewer/pull/316), I tried using `visibilitychange` event, but that will cause the IModelApp shutdown to occur when you switch tabs. And the `unload` event is deprecated, so there is no long term viable alternatives...

### Additional context

_No response_

GerardasB:
## Changes

This PR fixes #857 by updating handlers of `VirtualizedPropertyGrid` and `UrlPropertyValueRenderer` components to open links with `pw` scheme in a new window.

## Testing

Updated unit tests and added story.

### DIFF
--- a/common/changes/@itwin/components-react/issue-857_2024-06-03-16-26.json
+++ b/common/changes/@itwin/components-react/issue-857_2024-06-03-16-26.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/components-react",
+      "comment": "Update `VirtualizedPropertyGrid` and `UrlPropertyValueRenderer` to open links with `pw` scheme in a new window.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/components-react"
+}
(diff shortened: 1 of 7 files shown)


aruniverse (approved):

ignas-k (approved):

grigasp:
As far as I remember, `mailto` and `pw` schemes were handled differently, because that allowed associated apps take over the "open" request. I think it needs to be confirm that a local ProjectWise app still opens the link after the change.

GerardasB:
> As far as I remember, `mailto` and `pw` schemes were handled differently, because that allowed associated apps to take over the "open" request. I think it needs to be confirmed that a local ProjectWise app still opens the link after the change.

I've tested this change with the `mailto` protocol as well and can confirm, that it gets handled correctly.

PR #868, opened 2024-06-11
Fix `ElementTooltip` to correctly render in a popup window

### LINKED ISSUE #862: Tooltip is always displayed in main window
### Describe the bug

As shown in the image below, the tooltip is always displayed in the main window where it should be displayed in the popout view.

### To Reproduce

1. Open a widget that has a popout with a viewport
2. Place a `Marker` that has the `title` property set. (It's the value of the tooltip.)
3. Hover the marker. Notice the tooltip doesn't show at the right location.

### Expected Behavior

I would expect the tooltip to show in the right window.

### Screenshots

image

Chrome 125.0.6422.114 (Official Build) (64-bit) (cohort: Control)
appui version 4.11, but tested with 4.13.4 and getting the same issue.

### Additional context

_No response_

GerardasB:
## Changes

This PR fixes #862 
The class component is refactored to a function component.

## Testing

Updated `PopupTestView` of a test app.

### DIFF
--- a/common/api/appui-react.api.md
+++ b/common/api/appui-react.api.md
@@ -1702,11 +1702,6 @@ export class DrawingNavigationAidControl extends NavigationAidControl {
 
 // @public
 export class ElementTooltip extends React_2.Component<CommonProps, ElementTooltipState> {
-    constructor(props: CommonProps);
-    // (undocumented)
-    componentDidMount(): void;
-    // (undocumented)
-    componentWillUnmount(): void;
     // (undocumented)
     static hideTooltip(): void;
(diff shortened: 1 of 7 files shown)


ignas-k (approved):
I suppose you saw this and went with `useEffect` for a reason? It doesn't work with `useEffect`?

image

GerardasB:
> I suppose you saw this and went with `useEffect` for a reason? It doesn't work with `useEffect`?
> 
> image

Yes, it's intentional to avoid flickering between preferred and calculated positions.

PR #873, opened 2024-06-11
FilterBuilder: Add allowEmptyFilter options for buildFilter function in usePropertyFilterBuilder

AnDuong249:
## Changes

- Add `allowEmptyRules` options for buildFilter in the useFilterPropertyBuilder hook so that the function will return the filter even if there are some errors when validating the rules instead of returning undefined.
- Add `allowInitialEmptyRules` props for the useFilterPropertyBuilder that modifies `getRangeRuleItems` behavior to return a filter even if both from and to value is undefined instead of returning undefined when one of them is undefined.

## Testing

Added new unit tests and visually checked component in Storybook.


saskliutas (changes requested):
After looking over changes I think that behavior controlled by these newly introduced props should be the default and there is no need for those props. `convertFilterToState` and `buildPropertyFilter` should not ignore or skip over rules that do not have values. 

Default rule validator will take care of empty values and if you want to allow empty values you can pass custom rule validator.

PR #879, opened 2024-06-14
Portal editor to same container used to portal other components

saskliutas:
## Changes

Container used to portal components had higher z-index than `Popup` components. Changed editors to use same portal container. Required to fix https://github.com/iTwin/viewer-components-react/issues/921.

## Testing

Tested manually

### DIFF
--- a/common/api/core-react.api.md
+++ b/common/api/core-react.api.md
@@ -1230,6 +1230,7 @@ export interface PopupProps extends CommonProps {
     onOpen?: () => void;
     onOutsideClick?: (e: MouseEvent) => void;
     onWheel?: (e: WheelEvent) => void;
+    portalTarget?: HTMLElement;
     position: RelativePosition;
     // @beta
     repositionOnResize?: boolean;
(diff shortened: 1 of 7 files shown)


common/api/core-react.api.md
```diff
@@ -1230,6 +1230,7 @@ export interface PopupProps extends CommonProps {
     onOpen?: () => void;
     onOutsideClick?: (e: MouseEvent) => void;
     onWheel?: (e: WheelEvent) => void;
+    portalTarget?: HTMLElement;
```
GerardasB, line 1233: API addition, document in NextVersion

GerardasB (approved):

common/api/core-react.api.md
saskliutas: Should be advertise changes to deprecated components?

GerardasB: I don't think it's an issue. Changelogs should be complete with all the changes to avoid any surprises.

saskliutas: Updated changelog

PR #883, opened 2024-06-19
Dialog changes

ignas-k:
## Changes
In this PR:

- Deprecated `ContentDialog`, `ModelessDialog` and `UiDataProvidedDialog` in favor of iTwinUI `Dialog`.
- Deprecated `dialogId` prop of `AccuDrawDialogProps`. This prop is no longer needed because `AccuDrawDialog` was changed to use iTwinUI `Dialog` instead of `ModelessDialog`.
- Deprecated `dialogs` prop of `FrameworkContent`. Consumers should use `UiFramework.dialogs.modal` or `UiFramework.dialogs.modeless` instead to control dialogs.
- Changed `StandardMessageBox` from a class to a functional component.
- Changed a few components to use iTwinUI `Dialog` instead of various AppUI's dialogs. 

## Testing


ui/appui-react/src/appui-react/framework/FrameworkContent.ts
```diff
@@ -196,6 +197,7 @@ export interface FrameworkContent {
   /**
    * Manage dialogs displaying managed content.
    * @beta
+   * @deprecated in 4.15.0. Use `modeless` or `modal` properties of {@link UiFramework.dialogs} instead.
    */
   readonly dialogs: FrameworkStackedDialog<ContentDialogInfo> & {
```
ignas-k, line 202: @GerardasB do you think it is fine to deprecate this? I did it since we decided that content dialogs do not make sense and should just function as other modeless dialogs

GerardasB: Yes, it should be fine.

GerardasB (approved):
Do we have a way to enforce that a dialog is modal/modeless from within `UiFramework.dialog.modal` & `UiFramework.dialog.modeless` APIs?

ignas-k:
Not really. Both of them accept `ReactNode` so it is all up to the consumer. I suppose if iTwinUI implements the changing `z-index` functionality maybe `modal` and `modeless` properties could be deprecated and merged unless there are some other reasons for keeping track which dialogs are modal or modeless.

PR #884, opened 2024-06-20
FilterBuilder: Allow adding custom item for PropertyFilterBuilderActions

AnDuong249:
## Changes
- Modify item parameter type in `addItem` function in `PropertyFilterBuilderActions` to `"RULE_GROUP" | "RULE" | PropertyFilterRule | PropertyFilterRuleGroup` so that users can pass in custom item to initialize newly created rule/group instead of always creating empty one.

## Testing
- Added new unit tests and visually checked component in Storybook.


ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
```diff
@@ -110,15 +110,25 @@ export class PropertyFilterBuilderActions {
   }
 
   /** Adds new rule or group of rules to the group specified by path. */
-  public addItem(path: string[], itemType: "RULE_GROUP" | "RULE") {
+  public addItem(
+    path: string[],
+    itemType: "RULE_GROUP" | "RULE",
+    itemData?: PropertyFilterBuilderRule | PropertyFilterBuilderRuleGroup
```
saskliutas, line 116:
```suggestion
    item: "RULE_GROUP" | "RULE" | PropertyFilterRule | PropertyFilterRuleGroup,
```

I think `addItem` should accept `PropertyFilterRule/Group` and create `id`s inside.

saskliutas (changes requested):

ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
AnDuong249: Added

saskliutas (approved):
LGTM. You will need to run `rush change` to generate changelog and mention your changes in `docs/NextVersion.md`

GerardasB (approved):

PR #890, opened 2024-07-02
Show property record description next to array indices

Yato333:
## Changes

Part of https://github.com/iTwin/presentation/issues/654
Property record description (if present) will be shown next to the index for array property items.
This should improve user experience when browsing through the property grid having many related instances.

## Testing

A unit test was added.

### DIFF
--- a/common/changes/@itwin/components-react/master_2024-07-02-06-46.json
+++ b/common/changes/@itwin/components-react/master_2024-07-02-06-46.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/components-react",
+      "comment": "Show label next to index for array property items",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/components-react"
+}
--- a/common/changes/@itwin/components-react/master_2024-07-02-06-46.json
+++ b/common/changes/@itwin/components-react/master_2024-07-02-06-46.json
@@ -2,7 +2,7 @@
   "changes": [
     {
       "packageName": "@itwin/components-react",
-      "comment": "Show label next to index for array property items",
+      "comment": "Property grid: Show array items' description next to the index when the items are non-primitive.",
       "type": "none"
     }
   ],
--- a/ui/components-react/src/components-react/propertygrid/internal/flat-items/MutableCategorizedArrayProperty.ts
+++ b/ui/components-react/src/components-react/propertygrid/internal/flat-items/MutableCategorizedArrayProperty.ts
@@ -54,14 +54,7 @@ export class MutableCategorizedArrayProperty
     this._children = record.getChildrenRecords().map((child, index) => {
       const newName = `${child.property.name}_${index}`;
       let newDisplayLabel = `[${index + 1}]`;
-
-      if (
-        child.description &&
-        ((child.value.valueFormat === PropertyValueFormat.Array &&
-          child.value.items.length > 1) ||
-          (child.value.valueFormat === PropertyValueFormat.Struct &&
-            Object.keys(child.value.members).length > 1))
-      ) {
+      if (child.description && child.value.valueFormat !== PropertyValueFormat.Primitive) {
         newDisplayLabel += ` ${child.description}`;
       }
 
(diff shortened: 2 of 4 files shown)


ui/components-react/src/components-react/propertygrid/internal/flat-items/MutableCategorizedArrayProperty.ts
```diff
@@ -50,7 +53,18 @@ export class MutableCategorizedArrayProperty
...
-      const newDisplayLabel = `[${index + 1}] ${child.property.displayLabel}`;
+      let newDisplayLabel = `[${index + 1}]`;
+
+      if (
+        child.description &&
+        ((child.value.valueFormat === PropertyValueFormat.Array &&
+          child.value.items.length > 1) ||
+          (child.value.valueFormat === PropertyValueFormat.Struct &&
+            Object.keys(child.value.members).length > 1))
+      ) {
```
grigasp, line 64: why was this condition necessary?

Yato333: Because labels for primitive values are taken from the description, so there would be duplicate labels for primitive array items like this:
[1] Description:    Description

Yato333: And if I recall correctly, structs and arrays that have one element are flattened to display that element instead?

grigasp: > Because labels for primitive values are taken from the description, so there would be duplicate labels for primitive array items like this:
> [1] Description: Description

Yes, and I think that's the only check we should do - if the array item is not primitive.

> And if I recall correctly, structs and arrays that have one element are flattened to display that element instead?

That happens before the property data gets to `@itwin/components-react`. Once it's here, no more modifications are made to it.

Yato333: Gotcha, removed extra conditions

common/changes/@itwin/components-react/master_2024-07-02-06-46.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/components-react",
+      "comment": "Show label next to index for array property items",
```
grigasp, line 5:
```suggestion
      "comment": "Property grid: Show array items' description next to the index when the items are non-primitive.",
```

ui/components-react/src/test/propertygrid/component/internal/flat-items/CategorizedArrayProperty.test.ts
```diff
@@ -176,6 +176,81 @@ describe("CategorizedArrayProperty", () => {
         expect(overrideDisplayLabel).toEqual(expectedOverrideDisplayLabel);
       });
     });
+
+    it("Should append property description to non-primitive items that have more that 1 items", () => {
```
grigasp, line 180:
```suggestion
    it("Should append property description to non-primitive items", () => {
```

ui/components-react/src/test/propertygrid/component/internal/flat-items/CategorizedArrayProperty.test.ts
```diff
@@ -176,6 +176,81 @@ describe("CategorizedArrayProperty", () => {
...
+              "CADID1_6_2",
+              "test"
+            ),
+          }),
+          TestUtils.createArrayProperty("CADID1_7", [
+            TestUtils.createPrimitiveStringProperty("CADID1_7_1", "test"),
+            TestUtils.createPrimitiveStringProperty("CADID1_7_2", "test"),
+          ]),
+        ].map((record, idx) => {
+          record.description = `${idx}`;
```
grigasp, line 200: Maybe make this `Description_${idx}`? While here it's pretty clear what you're doing, down below, where you do the checks, the assertion is quite confusing: `[${index + 1}] ${index}`.

Yato333: Done

grigasp (approved):

GerardasB (approved):

GerardasB (commented):
It would be great to mention it in the `NextVersion.md` and add before/after screenshots to the PR description.

Yato333:
@GerardasB there goes the NextVersion.md 😄
Added screenshots and updated NextVersion.md in #897

PR #902, opened 2024-07-09
Remove `ConditionalIconItem.isConditionalIconItem()` workaround

### LINKED ISSUE #900: Conditional icons fail to render for built in toolbar buttons
### Describe the bug

We are experiencing a bug in AppUI based viewer app that conditional icons are failing to render and only a blank button appears on screen (without icon). The problem is reproducible with built in toggle camera toolbar button available with StandardNavigationToolsUiItemsProvider which is using [ToolbarItems.createToggleCameraView](https://github.com/iTwin/appui/blob/f92b456aa301882a94e93e82b087b1dd87b97229/ui/appui-react/src/appui-react/ui-items-provider/StandardNavigationToolsUiItemsProvider.ts#L151) to draw the toolbar button.

### To Reproduce

No special steps. When application loads up, the vertical toolbar for navigation tools is missing the toggle camera icon. 

### Expected Behavior

Toggle camera icon should appear on the toolbar button

### Screenshots

image

Application is running on windows 11, 
It is an ElectronJS based app
@iTwin/appui-abstract: ^4.7.4
@iTwin/appui-react: ^4.12.0
@iTwin/components-react: ^4.12.0

iTwinJS version: 4.7.4

### Additional context

The problem seems to be with conditional icons which are using [CondictionalIconItem ](https://github.com/iTwin/appui/blob/f92b456aa301882a94e93e82b087b1dd87b97229/ui/core-react/src/core-react/icons/ConditionalIconItem.tsx#L14)class of AppUI. 

Apparently the class has a function [isConditionalIconItem ](https://github.com/iTwin/appui/blob/f92b456aa301882a94e93e82b087b1dd87b97229/ui/core-react/src/core-react/icons/ConditionalIconItem.tsx#L33)to check if a given item is an instance of this class and it seems to be failing. 

The function is using following statement to check the given instance

    if (itemPrototype.constructor.name !== "ConditionalIconItem") return false;

From what i observed, the evaluated constructor name of its instances are always prefixed with _ so they are _ConditionalIconItem and not ConditionalIconItem. The [ActionItem](https://github.com/iTwin/appui/blob/f92b456aa301882a94e93e82b087b1dd87b97229/ui/components-react/src/components-react/toolbar/InternalToolbarComponent.tsx#L297) component uses [IconHelper.getIconReactNode](https://github.com/iTwin/appui/blob/f92b456aa301882a94e93e82b087b1dd87b97229/ui/core-react/src/core-react/utils/IconHelper.tsx#L42) which checks if icon is of type CondictionalIconItem using function above and returns null since above function incorrectly returns false. 

In getIconReactNode function, if I evaluate the icon using 'icon instanceof ConditionalIconItem', i get true. 

The fix is to request the AppUI team to either

- Fix the function isConditionalIconItem to consider _ prefixed constructor name
- Or remove it in favor of instanceof checks

Note other similar classes like ConditionalStringValue do not have any such function and all checks are using instanceof throughout AppUI code.

GerardasB:
## Changes

This PR fixes #900 by removing a `ConditionalIconItem.isConditionalIconItem()`. It was added as a workaround for #315, which I suspect might have happened due to the usage of **both** ESM and CJS modules.

## Testing

Tested in standalone test-app and storybook.

### DIFF
--- a/common/api/core-react.api.md
+++ b/common/api/core-react.api.md
@@ -224,8 +224,6 @@ export class ConditionalIconItem {
     static getValue(conditionalValue: ConditionalIconItem | string | undefined): IconSpec | undefined;
     // (undocumented)
     readonly iconGetter: () => IconSpec;
-    // @internal
-    static isConditionalIconItem(item: any): item is ConditionalIconItem;
     refresh(): boolean;
     static refreshValue(conditionalValue: ConditionalIconItem | string | undefined, eventIds: Set<string>): boolean;
     // (undocumented)
--- a/common/changes/@itwin/appui-react/issue-900_2024-07-09-07-50.json
+++ b/common/changes/@itwin/appui-react/issue-900_2024-07-09-07-50.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/appui-react"
+}
--- a/common/changes/@itwin/components-react/issue-900_2024-07-09-07-50.json
+++ b/common/changes/@itwin/components-react/issue-900_2024-07-09-07-50.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/components-react",
+      "comment": "",
+      "type": "none"
+    }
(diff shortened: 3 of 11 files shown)


GerardasB:

ignas-k (approved):

PR #912, opened 2024-07-15
Fix `PopupContextMenu` in `StatusBarPopover`

### LINKED ISSUE #909: Regression with opening PopupContextMenu from within Status bar popover
### Describe the bug

When PopupContextMenu is invoked by clicking a button inside an opened status bar popover component, it briefly renders the context menu behind the parent component and immediately closes both.

### To Reproduce

1. Create a status bar popover component with a button that triggers a PopoverContextMenu from @itwin/core-react
2. Open the status bar popover component from the status bar
3. Click the button to trigger the context menu
4. Both, the context menu, and the popover component are closed

### Expected Behavior

Context menu should be rendered on top of the parent status bar popover component and not close the popover.
Last iTwin.js AppUI version where it seems to be working correctly is 4.13.3.

### Screenshots

_No response_

- OS: Windows 11
- Browser: Chromium (within Electron)
- Browser version: v124
- iTwin.js AppUI version: v4.14.1
- iTwin.js version: v4.7.3

### Additional context

_No response_

ignas-k:
## Changes
When `PopupContextMenu` is used as a nested popup in `StatusBarPopover` it is problematic because the logic of iTwinUI (Popover) and AppUI (Popup) clash since they portal to different places which then causes various issues (e.g. the nested `Popup` displays behind the main `Popover` or opening the nested `Popup` would close the `Popover`). As a simple fix I used `portalTarget` prop which will portal the `Popup` into the `Popover` (more specifically, the trigger button which is inside the Popover).

Additionally, `PopupContextMenu` will be deprecated in AppUI 4.16.0 and can be easily replaced with `DropdownMenu` or `DropdownButton` from iTwinUI. 

Fixes #909.

## Testing

### DIFF
--- a/common/changes/@itwin/core-react/fix-popup-context-menu_2024-07-15-06-05.json
+++ b/common/changes/@itwin/core-react/fix-popup-context-menu_2024-07-15-06-05.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/core-react",
+      "comment": "Fix `PopupContextMenu` to work in `StatusBarPopover`.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/core-react"
+}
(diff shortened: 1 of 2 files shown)


ignas-k:
I tested this with other `Popup` components and even when it is not used in another popup. Everything seems to be working the same as before.

PR #916, opened 2024-07-18
Fix popout widget changing position on second monitor

### LINKED ISSUE #896: Electron: windows are "running away" on consecutive widget popouts when they were first moved to another monitor
### Describe the bug

I reproduced it on electron environment, not sure if applicable elsewhere.

This issue happens when I pop out a widget and I move it to another monitor for the first time. After closing window and popping it out consecutively, with each iteration it goes further left/right (depending what direction second monitor is). Even if I move it back to original monitor it keeps shifting, eventually it is escapes all monitors completely and it can only be brought back with "maximize" option on Windows toolbar (but still escapes afterwards).

### To Reproduce

1. Pop out any widget.
2. Move widget to another monitor. Remember its position there.
3. Close that widget window to return it.
4. Pop out that widget and close it again (should be ok for now and stay in the same place at another monitor).
5. Pop out that widget again.
6. Note that it shifted.
7. Keep repeating this until widget is outside of all monitors.

### Expected Behavior

Widgets should stay in same place when repeatedly popped out at other monitors and not shift. 

### Screenshots

Stages of widget running away:
image
image
image

**OS:** Windows 11
**Browser:** Microsoft Edge
**Browser version:** Version 126.0.2592.68 (Official build) (64-bit)
**Appui version:** 4.15.0 (reproduced on appui source - standalone electron app)

### Additional context

Would need this fixed before our Modeler's release, our feature freeze is at beggining of next month. Sometimes its confusing for user, it looks like widgets are minimized and they are cumbersome to restore.

ignas-k:
## Changes
Moved `deltaLeft` and `deltaTop` calculations after rendering the content. Seems like when the popout widget is on the second monitor `left` and `top` values are set correctly only after the content is rendered.

Fixes #896.

## Testing
Tested in electron. Regular browsers are not relevant because they will always open the child window on the primary monitor.

### DIFF
--- a/common/changes/@itwin/appui-react/moving-popout-widget-second-monitor-fix_2024-07-18-05-55.json
+++ b/common/changes/@itwin/appui-react/moving-popout-widget-second-monitor-fix_2024-07-18-05-55.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "Fixed popout widget changing position on a secondary montitor.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/appui-react"
+}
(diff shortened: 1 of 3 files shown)


ignas-k:

Juliakas:
Hi @ignas-k , do you think we can even backport this to 4.14.x? Would help us out a lot!

ignas-k:
@Mergifyio backport release/4.14.x

GerardasB (approved):
Nice!

Issue #915, opened 2024-07-17
The commit behavior of the PropertyPane editor needs some modifications


Rock2000:
### Describe the bug

When a commit is triggered from the UI, the component will propagate the change event even if the 'backing' value is exactly the same. This results in a transaction on the undo stack when a user simply enters edit mode and then clicks off (changes nothing). Also the ESC key seems to commit, which it should not.

### Expected Behavior

The ESC key should be an easy one. If ESC is hit, the formatted original value should be restored to the widget, and I think edit mode of the property should be ended.

Fixing the other issue brings up a few more situations to worry about. For example, a backing value of 17.0026 could be displayed as 17.0. But it is possible that the user actually wants to edit the value and set it to exactly 17. So I think we do need to support that case also, but probably only via an explicit Enter key. Here is what I think the Booster team would like to see for behavior once in edit mode:

if (loseFocus) {
  if (edit string != originalFormattedString)
    tryCommit()
} else if (EnterKey)
    tryCommit()
}
 
tryCommit() {
  newValue = parse(editText)
  if (isValid(newValue) && newValue != originalValue)
    commitForReal()
}

### Desktop (please complete the applicable information)

iTwin.js 4.7.2
AppUI 4.12.0

grigasp:
The fix released with `@itwin/components-react@4.16.0`

PR #928, opened 2024-07-23
Do not commit edited value if it contains no changes

jasdom:
## Changes

Closes https://github.com/iTwin/appui/issues/915. Updated `EditorContainer` to not commit the value when `Escape` is pressed or when the value is identical to the original (for primitive properties).

## Testing

Added and updated unit tests, manually tested in test app.


GerardasB (commented):
How did you test the ToolSettingsPopup? Would you mind adding it to storybook?

jasdom:
> How did you test the ToolSettingsPopup? Would you mind adding it to storybook?

I tested it by using the setup of the unit test I modified in a test app. Moved the code into storybook.

GerardasB (approved):

grigasp:
The fix released with `@itwin/components-react@4.16.0`

PR #929, opened 2024-07-24
Fix UI visibility when `autoHideUi` is toggled

### LINKED ISSUE #927: AutoHideUI and collapse panels toggle sometimes don't work
### Describe the bug

While testing user setting options for auto hide toolbars and collapse panels, it came to notice that the toggle fails to update the UI sometimes. This is not reproducible each time but within 2-3 attempts of toggling, the issue is reproducible. 

### To Reproduce

Toggle autohide ui and collapse panels via 

- UiFramework.visibility.autoHideUi (toggle true/false a couple time)
- <ConfigurableUiContent collapsePanels= {collapse} - Update collapse via state change between true/false

### Expected Behavior

Toolbars should appear and stay on screen if autoHideUi is false
Unpinned panels should remain visible if collapsePanels was called with false

### Screenshots

_No response_

Windows 11, Chrome, iTwinUI 4.15.1, iTwinJS 4.7.4

### Additional context

The problem does not occur if we have debugging being done, only happens if no breakpoints exist so seems like a timing/sync issue.

GerardasB:
## Changes

This PR fixes #927 by resetting UI visibility when `autoHideUi` is disabled.

## Testing

Added additional story and unit test.


GerardasB:

ignas-k (approved):

PR #934, opened 2024-07-26
Fix standard content tools throwing uncaught exception with transient elements

### LINKED ISSUE #926: `StandardContentToolsUiItemsProvider` tools throws unhandled exception when used on temporary elements
### Describe the bug

Temporary elements created by "measurement tools" are selectable, and user can try to "isolate/emphasize/hide selected models/categories" via isolation tools provided by `StandardContentToolsUiItemsProvider`.

However, such action results in unhandled promise rejection.

### To Reproduce

See this sandbox: https://www.itwinjs.org/sandboxes/ImVeryLost/candide-the-magenta-vulture-473

1. Select measurement tools > distance
2. Add some measurement. Make sure the "distance line" is selected.
3. Select isolate tools > isolate models
4. Following exception is thrown : "Uncaught (in promise) Not Found: reading element={id: 0xffffff0000000005 federationGuid: undefined, code: undefined}"

Similar error is seen after "isolate categories", "hide models/categories" or "emphasize models/categories" is selected.

### Expected Behavior

Isolation and other tools provided by the `StandardContentToolsUiItemsProvider` should not throw unhandled exceptions.

### Screenshots

firefox_WZCPHLt1To

Reproducible both in sandbox and in local environment.

**Local setup**

browser: electron
"@itwin/appui-react": 4.12.0 
"@itwin/measure-tools-react": "0.14.2",

### Additional context

Related internal item: https://dev.azure.com/bentleycs/iModelTechnologies/_workitems/edit/1472900

ignas-k:
## Changes

Added a filter to remove transient elements when calling `IModelConnection.elements.getProps()` as it was returning a rejected promise because the transient elements do not exist in the iModel.

Fixes #926.

_Note_: to determine that the element is geometric it is possible to do a query: https://github.com/iTwin/DevicePainter/blob/ee0af3202c8faaea7dcaaef269c73b9e1783a16e/src/frontend/api/IModelHelper.ts#L140

## Testing
Tested in a separate test project.


ui/appui-react/src/appui-react/selection/HideIsolateEmphasizeManager.ts
```diff
@@ -330,8 +330,20 @@ export class HideIsolateEmphasizeManager extends HideIsolateEmphasizeActionHandl
     vp.changeCategoryDisplay(ids, true);
   }
 
+  private static async getSelectionSetElementProps(iModel: IModelConnection) {
```
GerardasB, line 333: Something like `toPersistentSet(elements: Id64Set): Id64Set` might be easier to understand, reuse and test.

ui/appui-react/src/appui-react/selection/HideIsolateEmphasizeManager.ts
```diff
@@ -340,9 +352,10 @@ export class HideIsolateEmphasizeManager extends HideIsolateEmphasizeActionHandl
   private static async getSelectionSetElementCategories(
     iModel: IModelConnection
   ) {
-    const props = (await iModel.elements.getProps(
-      iModel.selectionSet.elements
-    )) as GeometricElementProps[];
+    const props =
+      (await HideIsolateEmphasizeManager.getSelectionSetElementProps(
+        iModel
+      )) as GeometricElementProps[];
```
GerardasB, line 358: We are getting lucky with this typecast, since `prop.category` is false if the required property doesn't exist...

GerardasB (approved):

ui/appui-react/src/appui-react/selection/HideIsolateEmphasizeManager.ts
ignas-k: What is the problem here? If it returns false, then the category from the prop does not get added to the set so it will be irrelevant but nothing will go wrong.

GerardasB: Yes, if we are certain that only geometric elements are in the selection set when transient ids are filtered out.

ui/appui-react/src/appui-react/selection/HideIsolateEmphasizeManager.ts
ignas-k: Usually I would agree but I think in this case that is not needed because we can't test this method either way (because it is private) and for reusability I like to use YAGNI (you ain't gonna need it), so if there is no need for reusability now or in the near future, why do it. The methods is perfect for those two other places where it is called. This is just my opinion though.

ui/appui-react/src/appui-react/selection/HideIsolateEmphasizeManager.ts
ignas-k: Decided to leave it as it is for now as there seems to be no simple way to ensure that the element is geometric. Possible alternative would be something like this: https://github.com/iTwin/DevicePainter/blob/ee0af3202c8faaea7dcaaef269c73b9e1783a16e/src/frontend/api/IModelHelper.ts#L140

PR #936, opened 2024-07-29
Change selection count padding

### LINKED ISSUE #935: Padding for Selection Count right side is to small
### Describe the bug

With changes in https://github.com/iTwin/appui/pull/848/files#diff-e0669cc7693186c148947d0caa5c718339685bf6d1c94293cef144535c861a5f
for  
[ui/appui-react/src/appui-react/statusfields/SelectionCount.scss](https://github.com/iTwin/appui/pull/848/files#diff-e0669cc7693186c148947d0caa5c718339685bf6d1c94293cef144535c861a5f) the padding around the selection count was completely removed when some padding is still needed

AppUi 4.14.x
 
image

AppUi 4.13.3
image

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Screenshots

_No response_

_No response_

### Additional context

_No response_

ignas-k:
## Changes
Adjusted `SelectionCount` styling to improve visuals:

- Added `min-width` value so that the component will not shift all other status bar fields when another digit is added/removed. It should show 5 digits without shifting.
- Added `margin-right` when the component is the last child in `statusBar-right` so that it will be further from the corner. This just shows a limitation of our provider system because users cannot adjust the styling easily when using this component from `StandardStatusbarUiItemsProvider` and we cannot apply the same styling in all situations because the margin is applied automatically if the component is not the last child. Applying the margin for all components that are last in the section also does not make sense because buttons e.g. overflow button looks fine being in the corner.
- Added `margin-left` when the component is the first child in `statusBar-left`. Same as above.

Fixes #935.

**Before:**
image 
image

**After:**
image
image

## Testing


GerardasB (approved):

ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
```diff
@@ -6,9 +6,22 @@
...
-  padding: 0 var(--iui-size-2xs) 0 var(--iui-size-2xs);
+  min-width: var(--iui-size-2xl);
+  padding-left: var(--iui-size-2xs);
 
   .icon {
     color: var(--iui-color-icon);
   }
 }
+
+.uifw-statusBar-right
```
GerardasB, line 17: Looks good for a quick fix, but ideally the component should add a class name for these conditions (i.e. you could wrap each item in React context with needed info). Please add some screenshots as well.

ignas-k: Screenshots added. I suppose you don't want me to add the context now?

GerardasB: If you could, I'd prefer to do it now

ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
```diff
@@ -6,9 +6,18 @@
...
   align-items: center;
-  padding: 0 var(--iui-size-2xs) 0 var(--iui-size-2xs);
+  min-width: var(--iui-size-2xl);
+  padding-left: var(--iui-size-2xs);
 
   .icon {
     color: var(--iui-color-icon);
   }
+
+  &.left-statusBar-corner {
```
GerardasB, line 16: `uifw-` prefixes still needed. Fine to simplify, since it's a nested element i.e.: `.uifw-left-corner`

ui/appui-react/src/appui-react/statusbar/StatusBarComposer.tsx
```diff
@@ -541,14 +542,80 @@ export function StatusBarComposer(props: StatusBarComposerProps) {
     >
       <StatusBarSpaceBetween className={mainClassName}>
         <StatusBarLeftSection className={leftClassName}>
-          {leftItems}
+          {leftItems.map((item, index) => (
```
GerardasB, line 545: Can this be refactored/simplified, looks like spaghetti.

GerardasB (approved):

ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
ignas-k: Updated.

ui/appui-react/src/appui-react/statusbar/StatusBarComposer.tsx
ignas-k: Yeah, my bad. I forgot that in previous PRs I already sorted the status bar items. I think now it is pretty clean.

PR #940, opened 2024-07-31
Correctly copy styles of complex shorthand properties

### LINKED ISSUE #893: Popouts: Sometimes stylesheets don't get copied over correctly to new window
### Describe the bug

Sometimes when styling components and combining shorthand properties with css variables and adding other related longhand properties to override something, document.stylesheet produces unexpected results that are then copied over to popout window and wrong styles are getting applied.

One example being:
```scss
.border-test {
  --border-color: red;
  border: 5px solid var(--border-color);
  border-bottom: 5px solid blue;
  height: 20px;
  width: 100px;
}
```
In main window it looks like so:
image

And when popping out to new window, this gets produced upon inspecting: 
image
image

One interesting thing is that it works correctly when I use static value instead of variable or when I remove `border-bottom`.

This seems to be intended spec behavior that serializes `cssText` property like that. I found this chromium rejected issue:
https://issues.chromium.org/issues/40804066 and spec was linked to indicate intended behavior: https://drafts.csswg.org/css-variables/#pending-substitution-value

### To Reproduce

1. Add a simple div to a widget that can popout with a classname that defines the following styles:
    ```scss
    .border-test {
      --border-color: red;
      border: 5px solid var(--border-color);
      border-bottom: 5px solid blue;
      height: 20px;
      width: 100px;
    }
    ```
2. Open application, note applied borders for that div inside a widget, all 4 of them are applied with bottom one being different color.
3. Popout that widget.
4. Notice how only bottom border is applied.
5. Open devtools, inspect that div and note that stylesheets have deformed and lost information on remaining 3 borders.

### Expected Behavior

It is expected that when popping out widgets to a new window, all styles should be preserved as is and should not lose information. In this case it is expected that all borders retain same css properties as it had in main window.

### Screenshots

_No response_

**OS:** Windows 11
**Browser:** Microsoft Edge
**Browser version:** Version 126.0.2592.68 (Official build) (64-bit)
**Appui version:** 4.15.0 (reproduced on source)

### Additional context

I debugged source at `CopyStyles.ts` and noticed deformation was happening there. In my case I worked around this issue by being very verbose with longhand css properties and it seemed to work, but it can still cause issues as we just haven't discovered problems in other areas, maybe even in external code that we can't control.

GerardasB:
## Changes

This PR fixes #893 by refactoring `copyStyles` to avoid using `cssText` property when copying `document.styleSheets`. 

Additionally, a visual element is added to the test-app widget that is styled using various property variations with both `styleSheets` and `adoptedStyleSheets`. 

Seems like `cssText` is **serialized as expected** when used from within a contructable style sheet.

## Testing

Added a visual test.


ignas-k (approved):

test-apps/appui-test-app/standalone/src/frontend/appui/frontstages/TestPopoutFrontstage.tsx
```diff
@@ -0,0 +1,49 @@
...
+  }
+};
+
+(() => {
+  const sheet = new CSSStyleSheet();
+  sheet.replaceSync(`
+    #border-test {
+      --border-top-color: yellow;
+      border-top: 5px solid var(--border-top-color);
+      border-right: 5px solid green;
```
mayank99, line 45: Did you mean to use a `border` shorthand here? Because otherwise the problem with `cssText` will not be seen.

ui/appui-react/src/appui-react/childwindow/CopyStyles.ts
```diff
@@ -2,39 +2,34 @@
...
 ) {
-  const stylesheets = sourceDoc.adoptedStyleSheets
-    ? Array.from([...sourceDoc.styleSheets, ...sourceDoc.adoptedStyleSheets])
-    : Array.from(sourceDoc.styleSheets);
+  const styleSheets = Array.from(sourceDoc.styleSheets);
+  styleSheets.forEach(({ ownerNode }) => {
+    // Copy `link` and `style` elements.
+    if (!ownerNode) return;
+    const clonedNode = targetDoc.importNode(ownerNode, true);
+    targetDoc.head.appendChild(clonedNode);
```
mayank99, line 18: Have you tested whether `<link>` elements reuse the same cached resource or make a new network request?

test-apps/appui-test-app/standalone/src/frontend/appui/frontstages/TestPopoutFrontstage.tsx
GerardasB: Ah, good catch, that's why it serialized correctly :) I'll need a separate element.

ui/appui-react/src/appui-react/childwindow/CopyStyles.ts
```diff
@@ -2,39 +2,34 @@
...
-        targetDoc.head.appendChild(newStyleElement);
-      }
-    }
+  const adoptedStyleSheets = Array.from(sourceDoc.adoptedStyleSheets ?? []);
+  adoptedStyleSheets.forEach((styleSheet) => {
+    // Adopted stylesheet have no ownerNode and can't be shared between multiple documents.
+    if (!targetDoc.defaultView) return;
+    const newStyleSheet = new targetDoc.defaultView.CSSStyleSheet();
+    Array.from(styleSheet.cssRules).forEach((rule) => {
+      newStyleSheet.insertRule(rule.cssText);
```
mayank99, line 27: [`insertRule`](https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleSheet/insertRule) inserts the rule at the beginning of the stylesheet, so I believe this will reverse the order of all CSS rules, which can lead to incorrect output.

You could specify the index manually, but I would actually suggest using the `stringifyStyleSheet` helper from https://github.com/iTwin/appui/issues/893#issuecomment-2218036702. As a bonus, it's more readable and would be easier to modify in the future (teaser: iTwinUI could one day provide a pre-serialized version of its stylesheet to bypass `cssText` altogether).

ui/appui-react/src/appui-react/childwindow/CopyStyles.ts
GerardasB: Yes, it seems like the browser memory cache is reused.

ui/appui-react/src/appui-react/childwindow/CopyStyles.ts
GerardasB: Good catch, fixed by passing in an index for now.

test-apps/appui-test-app/standalone/src/frontend/appui/frontstages/TestPopoutFrontstage.tsx
GerardasB: Confirmed, I've just left a comment here and in `CopyStyles`.

mayank99 (approved):

PR #943, opened 2024-08-01
Update release tags of frequently used APIs

GerardasB:
## Changes

Fixes #891 by updating release tags of frequently used APIs.
Fixes #671 by removing `@internal` tags from React lifecycle methods and properties.
Additionally `@internal` tags are removed from methods tagged with the `override` keyword.

## Testing

N/A

PR #944, opened 2024-08-01
Bump more APIs to `@public`

GerardasB:
## Changes

Bump more APIs to `@public` in addition to https://github.com/iTwin/appui/pull/943

## Testing

N/A

### DIFF
--- a/ui/appui-react/src/appui-react/syncui/SyncUiEventDispatcher.ts
+++ b/ui/appui-react/src/appui-react/syncui/SyncUiEventDispatcher.ts
@@ -77,8 +77,6 @@ export class SyncUiEventDispatcher {
   private static _connectionUnregisterFuncs = new Array<() => void>();
   private static _iModelConnection?: IModelConnection;
 
-  /** @internal - used for testing only */
-
   public static setTimeoutPeriod(period: number): void {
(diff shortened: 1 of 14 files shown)


ui/appui-react/src/appui-react/syncui/SyncUiEventDispatcher.ts
```diff
@@ -77,8 +77,6 @@ export class SyncUiEventDispatcher {
   private static _connectionUnregisterFuncs = new Array<() => void>();
   private static _iModelConnection?: IModelConnection;
 
-  /** @internal - used for testing only */
```
ignas-k, line 80: I guess the idea is not to have `internal` methods in `public` classes? But if this method is only intended for testing, is it ok to make it non-internal?

ignas-k (approved):

ui/appui-react/src/appui-react/syncui/SyncUiEventDispatcher.ts
GerardasB: Not really, you can have an internal method in a public class, however, one of our consumers is using this method, so I've decided to simply remove the internal tag and promote this method to public.

PR #954, opened 2024-08-09
952 allow to enable multiple snaps

mathieu-fournier:
### Multi snaps 
We are now able to activate multiple snap modes simultaneousely. 
We can see in the picture that, when multiple modes are active at the same time, we have multiple possible snaps location depending ont he cursor position. 
image

### Snap modes choices
The consuming app can now choose what snap modes are available in their app. 
Ex, Bisector option has been filtered out. 
image

## Testing
2 unit tests are added to check the functionnality. 
- Ensuring at least one snap mode active at all times.
- Validating if we are able to select what snap modes are available in the app.

### DIFF
--- a/common/changes/@itwin/appui-react/952-allow-to-enable-multiple-snaps_2024-08-09-18-12.json
+++ b/common/changes/@itwin/appui-react/952-allow-to-enable-multiple-snaps_2024-08-09-18-12.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "We are now able to activate multiple snap modes at once. We can also choose what snap modes are available in the app via SnapModeField  snapModes props. ",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/appui-react"
+}
--- a/test-apps/appui-test-app/standalone/src/frontend/index.tsx
+++ b/test-apps/appui-test-app/standalone/src/frontend/index.tsx
@@ -37,6 +37,7 @@ import {
   SafeAreaContext,
   SafeAreaInsets,
   SessionStateActionId,
+  SnapModeField,
   StageUsage,
   StandardContentToolsUiItemsProvider,
   StateManager,
@@ -379,6 +380,25 @@ export class SampleAppIModelApp {
             id: "language",
             section: StatusBarSection.Right,
             content: <AppLanguageSelect />,
+          }),
+          StatusBarItemUtilities.createCustomItem({
+            id: "booster.SnapMode2",
+            section: StatusBarSection.Center,
+            content: (
+              <SnapModeField
+                snapMode={SnapMode.NearestKeypoint}
+                snapModes={
+                  Object.values(SnapMode).filter(
+                    (snapMode) => snapMode !== SnapMode.Bisector
+                  ) as SnapMode[]
+                }
+              />
+            ),
+          }),
+          StatusBarItemUtilities.createCustomItem({
+            id: "booster.SnapMode3",
+            section: StatusBarSection.Center,
+            content: <SnapModeField snapModes={[SnapMode.Bisector]} />,
           }),
         ],
       },
--- a/ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
+++ b/ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
@@ -61,6 +61,19 @@ function getIconFromIconName(iconName: string): IconSpec {
   return iconSpec;
 }
 
+/**
+ * Return the number of snap active in the bitmask.
+ */
+function getSnapCount(snapMode: SnapMode): number {
+  return Object.values(SnapMode).reduce<number>((count, snap) => {
+    if (typeof snap === "number" && snap & snapMode) {
+      // Check if the snap is present in the bitmask.
+      count++;
+    }
+    return count;
+  }, 0);
+}
+
 /** Return icon class name for a specific snapMode. */
 const getSnapModeIconNameFromMode = (
   modeEntries: SnapModeFieldEntry[],
@@ -78,6 +91,9 @@ const getSnapModeIconNameFromMode = (
 /** Defines properties supported by the SnapMode Field Component. */
 // eslint-disable-next-line deprecation/deprecation
 interface SnapModeFieldProps extends CommonProps {
+  /** The list of snap supported by the app. Default to all the modes.
+   */
+  snapModes?: SnapMode[];
   /** Uses redux store as a fallback. Defaults to {@link SnapMode.NearestKeypoint}.
    * @note Enum flags are supported.
    */
@@ -95,9 +111,11 @@ export function SnapModeField(props: SnapModeFieldProps) {
     // eslint-disable-next-line deprecation/deprecation
     (state) => state?.configurableUiState.snapMode
   );
-  const snapMode = props.snapMode ?? reduxSnapMode ?? SnapMode.NearestKeypoint;
 
-  const snapModes: SnapModeFieldEntry[] = [
+  const snapModeChoices = props.snapModes ?? [];
+  let snapMode = props.snapMode ?? reduxSnapMode ?? SnapMode.NearestKeypoint;
+
+  let snapModes: SnapModeFieldEntry[] = [
     {
       label: translate("snapModeField.keypoint"),
       value: SnapMode.NearestKeypoint as number,
@@ -134,6 +152,21 @@ export function SnapModeField(props: SnapModeFieldProps) {
       iconName: "snaps-bisector",
     },
   ];
+
+  if (snapModeChoices.length > 0) {
+    // Remove all snap mode not present in snapModeChoices.
+    snapModes = snapModes.filter((snapModeFieldEntry) =>
+      snapModeChoices.includes(snapModeFieldEntry.value)
+    );
+    if (
+      (!props.snapMode || !snapModeChoices.includes(props.snapMode)) &&
+      !snapModeChoices.includes(SnapMode.NearestKeypoint)
+    ) {
+      // The default snap mode have been filtered out. Default to the first one.
+      snapMode = snapModeChoices[0];
+    }
+  }
+
   const iconName = getSnapModeIconNameFromMode(snapModes, snapMode);
 
   const title = translate("snapModeField.snapMode");
@@ -149,8 +182,13 @@ export function SnapModeField(props: SnapModeFieldProps) {
                   props.onChange(item.value);
                   return;
                 }
-                // eslint-disable-next-line deprecation/deprecation
-                UiFramework.setAccudrawSnapMode(item.value);
+                // Bitwise XOR. Toggles the value in the bitmask. Allows multiple snap modes to be active at the same time.
+                const snapModeMask = snapMode ^ item.value;
+                // Checking for zero ensures that at least one mode remain active at all times.
+                if (snapModeMask !== 0) {
+                  // eslint-disable-next-line deprecation/deprecation
+                  UiFramework.setAccudrawSnapMode(snapModeMask);
+                }
               }}
               isActive={(snapMode & item.value) === item.value}
               icon={
@@ -170,8 +208,15 @@ export function SnapModeField(props: SnapModeFieldProps) {
       <Button
         styleType="borderless"
         title={title}
-        // eslint-disable-next-line deprecation/deprecation
-        endIcon={<Icon iconSpec={getIconFromIconName(iconName)} />}
+        endIcon={
+          getSnapCount(snapMode) > 1 ? (
+            // eslint-disable-next-line deprecation/deprecation
+            <Icon iconSpec={snapModeKeypoint} />
+          ) : (
+            // eslint-disable-next-line deprecation/deprecation
+            <Icon iconSpec={getIconFromIconName(iconName)} />
+          )
+        }
       >
         {title}
         <StatusBarPopover.ExpandIndicator />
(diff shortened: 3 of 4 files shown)


common/changes/@itwin/appui-react/952-allow-to-enable-multiple-snaps_2024-08-09-18-12.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "We are now able to activate multiple snap modes at once. We can also choose what snap modes are available in the app via SnapModeField  snapModes props. ",
```
GerardasB, line 5: Could you reword this to avoid `we are/can`, i.e.: "Added XYZ to `SnapModeField` component", "Multiple snap modes can be activated when using `SnapModeField` component.

test-apps/appui-test-app/standalone/src/frontend/index.tsx
```diff
@@ -380,6 +381,25 @@ export class SampleAppIModelApp {
...
+          StatusBarItemUtilities.createCustomItem({
+            id: "booster.SnapMode2",
+            section: StatusBarSection.Center,
+            content: (
+              <SnapModeField
+                snapMode={SnapMode.NearestKeypoint}
+                snapModes={
+                  Object.values(SnapMode).filter(
+                    (snapMode) => snapMode !== SnapMode.Bisector
+                  ) as SnapMode[]
```
GerardasB, line 393: This type-cast is not correct since the filtered array will include enum keys as strings https://www.typescriptlang.org/docs/handbook/enums.html#reverse-mappings

test-apps/appui-test-app/standalone/src/frontend/index.tsx
```diff
@@ -380,6 +381,25 @@ export class SampleAppIModelApp {
             section: StatusBarSection.Right,
             content: <AppLanguageSelect />,
           }),
+          StatusBarItemUtilities.createCustomItem({
```
GerardasB, line 384: Could you add these to the storybook instead, i.e. see `MessageCenterField.stories.tsx`

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
```diff
@@ -61,6 +61,19 @@ function getIconFromIconName(iconName: string): IconSpec {
   return iconSpec;
 }
 
+/**
+ * Return the number of snap active in the bitmask.
+ */
```
GerardasB, line 66:
```suggestion
/** Return the number of snap active in the bitmask. */
```

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
```diff
@@ -78,6 +91,9 @@ const getSnapModeIconNameFromMode = (
 /** Defines properties supported by the SnapMode Field Component. */
 // eslint-disable-next-line deprecation/deprecation
 interface SnapModeFieldProps extends CommonProps {
+  /** The list of snap modes displayed in the status bar. Default to all the modes.
+   */
```
GerardasB, line 95:
```suggestion
  /** The list of snap modes displayed in the status bar. Default to all the modes. */
```

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
```diff
@@ -78,6 +91,9 @@ const getSnapModeIconNameFromMode = (
...
+  snapModes?: SnapMode[];
```
GerardasB, line 96: The current property `snapMode` seems to already support multiple active snap modes via the enum flags. Although the PR description implies that it is something new we are adding?

This new property seems to allow filtering of available snap modes. Should we rename it to `availableSnapModes`?

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
```diff
@@ -149,8 +182,13 @@ export function SnapModeField(props: SnapModeFieldProps) {
                   props.onChange(item.value);
                   return;
                 }
-                // eslint-disable-next-line deprecation/deprecation
-                UiFramework.setAccudrawSnapMode(item.value);
+                // Bitwise XOR. Toggles the value in the bitmask. Allows multiple snap modes to be active at the same time.
+                const snapModeMask = snapMode ^ item.value;
```
GerardasB, line 186: We can't change the bavior like that. Please document `onChange` to clarify, that the value in the callback is a single enum flag toggled on or off. Consumers can do the bit magic on their own.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
```diff
@@ -134,6 +152,21 @@ export function SnapModeField(props: SnapModeFieldProps) {
...
+    );
+    if (
+      (!props.snapMode || !snapModeChoices.includes(props.snapMode)) &&
+      !snapModeChoices.includes(SnapMode.NearestKeypoint)
+    ) {
+      // The default snap mode have been filtered out. Default to the first one.
+      snapMode = snapModeChoices[0];
+    }
+  }
+
```
GerardasB, line 169: Can we refactor this? Start from the names, I'd suggest: `avilableSnapModes`, `snapModeEntries`, `availableSnapModeEntries`.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
```diff
@@ -170,8 +208,15 @@ export function SnapModeField(props: SnapModeFieldProps) {
       <Button
         styleType="borderless"
         title={title}
-        // eslint-disable-next-line deprecation/deprecation
-        endIcon={<Icon iconSpec={getIconFromIconName(iconName)} />}
+        endIcon={
+          getSnapCount(snapMode) > 1 ? (
```
GerardasB, line 212: I think this is unnecessary once we refactor and have the `availableSnapModeEntries`.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
```diff
@@ -95,9 +111,11 @@ export function SnapModeField(props: SnapModeFieldProps) {
     // eslint-disable-next-line deprecation/deprecation
     (state) => state?.configurableUiState.snapMode
   );
-  const snapMode = props.snapMode ?? reduxSnapMode ?? SnapMode.NearestKeypoint;
 
-  const snapModes: SnapModeFieldEntry[] = [
+  const snapModeChoices = props.snapModes ?? [];
```
GerardasB, line 115: I'd say if an empty array is provided - no snaps should be rendered.

ui/appui-react/src/test/statusfields/SnapMode.test.tsx
```diff
@@ -69,12 +69,62 @@ describe("SnapModeField", () => {
...
+    // Bisector snap mode should not be present.
+    expect(screen.queryByText("snapModeField.bisector")).toBeFalsy();
+
+    const iconContainer = container.querySelector(".icon");
+    expect(iconContainer).toBeTruthy();
+
+    const snaps = container.parentElement!.querySelectorAll(
+      ".nz-footer-snapMode-snap"
+    );
+    expect(snaps.length).to.eql(6); // 7 modes total, minus Bisector = 6 modes
```
GerardasB, line 126: Avoid styling/markup assertions, instead prefer byText assertions.

GerardasB (changes requested):

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
mathieu-fournier: It was supported, but the UI wasn't allowing it. 
The SnapModeField items acted like radio buttons, now they act like checkboxes. 

Agreed, I will rename it.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
mathieu-fournier: I thought a bit about that and having no snap rendered doesnt really makes sense. SnapModeField is a component to display snap modes. I cannot think of a use case for using it to display nothing. This sounds more like wrong usage of the component to me.
This, with the rule of having at least one snap mode active at all time, make me think that always displaying modes is a must.

ui/appui-react/src/test/statusfields/SnapMode.test.tsx
mathieu-fournier: When writing the tests, Bill used the `container` instead of using the `screen` variable. 
Like this, the test can be synchronous, but we cant get elements by text. 
I wonder if the performance gain from writing synchronous code is something we should consider, or if we should just go with the most robust selectors (getByText / Id) ?

GerardasB: Just go with getByText, which is best option and is available in this case

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
GerardasB: Could you clarify what "UI wasn't allowing it" means? Wasn't it rendered correctly or only the `onChange` handler would not do the bit-wise operations?

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
GerardasB: Well, it's up to the consumer to configure correctly. It will make your implementation simpler and produce the expected results.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
mathieu-fournier: The onChange would not do the bit-wise operation, so it acted like a radio button. 
In other words, when clicking a mode, it toggled OFF the other mode, so only one mode was ON.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
mathieu-fournier: Unfortunately, it makes the code more complex. 
It raises the case, what's the default snap mode, when no snap modes are available ?
image

Maybe, `undefined` or `Keypoint`, but both cases add complexity down the line. 
The code is simpler if we can assume that at least one snap mode is available at all time. 
I can still change it if you want, but I have the feeling that we would add complexity for a feature that has no value whatsoever.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
mathieu-fournier: I just changed the default behaviour of the component, not the way onChange worked for API consumers. 

`props.onChange` is still called with only the new snap mode that has been toggled on. 
It fit's the param name `newSnapMode`.
 `onChange?: (newSnapMode: SnapMode) => void;`

GerardasB: Oh, I see it now. However, behavior change is a breaking change as well i.e. current consumers might not want a multi-select behavior. Ideally, in such cases, an additional opt-in prop is added, however, I'd recommend that you use the `onChange` handler in your application and handle the bit-wise operations there since Redux state/dispatchers are deprecated.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
GerardasB: `SnapModeField` is just a component that renders some menu options. It should not be aware of application constraints, like how many snaps are active, it will still render depending on props passed to it (please ignore the redux store, which complicates this component). Let's render nothing if an empty array is passed in.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
mathieu-fournier: I supposed that supporting multi snap shoul've been the standard from the beginning. Since snapmode is a bitmask made with that purpose in mind. I dislike the idea of having the same component behaving differently in the apps. Especially components used for editing, like snaps. 
I guess this in more of a UX question. What should be the standard behaviour ?

I suggest we do the opposite, let consumers use the `onChange` handler if they dont want to support multi snap.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
mathieu-fournier: So, what's the snap mode when an empty array is passed in ? undefined of some default snap mode ?

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
GerardasB: The logic outside of `onChange` is considered deprecated and will be removed in 6.0 either way.
There's no way of knowing if any of the consumers depend on snap modes being single-select.

To be safe and to follow semantic versioning rules I'd suggest one of:
- Add an optional `multiple` flag to the component (will change both onChange and redux dispatch) - might not make sense long term if we want to support multi-select only (in the major version we could change the default value)
- Add a preview feature `multipleSnapModes` (once again, we can change the default in the next major version AND drop the preview feature)

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
GerardasB: The application determines what is the active snap mode and it should provide correct props to the component i.e. the component might not display "correct" snap mode depending on what props are provided.

mathieu-fournier:
In the light of the discussion, this PR will be splitted into 2. 
1 - Allow to control what snaps are available. 
2 - Set multisnap has the default behaviour <- breaking change for the next major version.

PR #964, opened 2024-08-13
Add router to merge connected and standalone test apps

GerardasB:
## Changes

This PR adds [TanStack Router](https://tanstack.com/router/latest) and merges the test apps under a single test-app
The functionality of previously existing `test-apps/appui-test-app/standalone` and `test-apps/appui-test-app/connected` apps is now accessible via different routes of `app/test-app`.

Environment keys were cleaned up as well, take a look at `.env.template` before moving your `.env` files.

## Testing

Ran the e2e tests and tested manually.


e2e-tests/tests/backstage/backstage.test.ts-snapshots/backstage-test-with-snapshotPath-1-chromium-linux.png
ignas-k, line 1: Is there a reason why `Custom Content`, `Synchronizer Floating View` and `Popout Windows` frontstages are not available in the backstage menu anymore?

ignas-k (changes requested):
Looks good, just left a few suggestions for documentation. Also for documentation purposes could you please add some before and after screenshots to the PR? I think widget-api fronstage and backstage menu should be good enough.

e2e-tests/tests/backstage/backstage.test.ts-snapshots/backstage-test-with-snapshotPath-1-chromium-linux.png
GerardasB: Added back the missing frontstages.

ignas-k (approved):

PR #967, opened 2024-08-14
Fix `useConditionalValue` to handle event ids as case insensitive

GerardasB:
## Changes

This PR fixes an issue where `useConditionalValue` would not handle the sync event ids correctly due to casing.

## Testing

Added additional unit test.


GerardasB:

ignas-k (approved):

PR #970, opened 2024-08-14
Render default navigation aid only if content control is not used

### LINKED ISSUE #965: 2D Navigation Aid appears when it shouldn't
### Describe the bug

- previously, when a `ViewportContentControl` overrides the `navigationAidControl` property to return an empty string, this would prevent any navigation aid from showing. This was used to suppress the default 2d navigation aid from showing. Now the 2d navigation aid appears despite the override.
- A new issue is showing where when a ViewportControl is opened in a widget the navigation aid of the viewport in the view is replaced with the one applicable to that new ViewportControl. This means if a 2d viewport is opened in viewport control inside a widget, the main view, which could still be 3d, will now show the 2d navigation aid
- both these issues are breaking changes introduced in 4.16.0

### To Reproduce

Default 2d Navigation Aid is no longer suppressed by navigationAidControl override
1. make the main frontstage use a viewport control with logic like
```typescript
override get navigationAidControl() {
    const viewClassName = this.viewport?.view.classFullName;
    if (
      !this.options?.featureOptions?.enable2dNavigationAidControl &&
      (viewClassName === DrawingViewState.classFullName ||
        viewClassName === SheetViewState.classFullName)
    ) {
      return ""; // disable navigation aid for 2d views
    }

    return super.navigationAidControl;
  }
```
2.  open a 2d view
3. notice the default 2d navigation cube appears
image
--------------
2d Navigation Aid shows in main view when opening a 2d view in a viewport control even when the main view remains 3d
1. have a frontstage with a widget using a viewport control
2. open a 3d view
3. open a 2d view in the widget with a viewport control
4. notice the main view, which is still showing a 3d view, now shows the 2d navigation aid
image

### Expected Behavior

Expected behavior is for the default 2d navigation cube to continue to be hidden when overriding the `navigationAidControl` to return an empty string and also for the navigation aid of the main view to be unaffected by the viewport control that a widget maybe using

### Screenshots

_No response_

_No response_

### Additional context

reproducible in 4.16.0

GerardasB:
## Changes

This PR fixes #965 a regression that prevented existing content controls from rendering a blank navigation aid.

## Testing

Added additional stories


jason-crow (approved):

ignas-k (approved):

GerardasB:

PR #974, opened 2024-08-14
Allow to set the available snap modes in SnapModeField.

mathieu-fournier:
## Snap modes choices
The consuming app can now choose what snap modes are available in their app.
Ex, Bisector option has been filtered out.
image

### DIFF
--- a/ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
+++ b/ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
@@ -15,7 +15,7 @@ import snapModeOrigin from "@bentley/icons-generic/icons/snaps-origin.svg";
 import snapModeKeypoint from "@bentley/icons-generic/icons/snaps.svg";
 import { SnapMode } from "@itwin/core-frontend";
 import { Icon } from "@itwin/core-react";
-import type { CommonProps, IconSpec } from "@itwin/core-react";
+import type { CommonProps } from "@itwin/core-react";
 import * as React from "react";
 import { UiFramework } from "../UiFramework";
 import { SnapModePanel } from "../layout/footer/snap-mode/Panel";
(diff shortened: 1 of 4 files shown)


ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
```diff
@@ -91,77 +106,46 @@ interface SnapModeFieldProps extends CommonProps {
...
+  const snapModeEntries: SnapModeFieldEntry[] = allSnapModeFieldEntries.filter(
+    (entry) => {
+      return props.availableSnapModes
+        ? props.availableSnapModes.some(
+            (availableSnap) => availableSnap === (entry.value as SnapMode)
+          )
+        : true;
+    }
+  );
+
```
ignas-k, line 119: If `props.availableSnapModes` is not provided then there is no reason to filter the entries. You can move that if condition outside of the filter.

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
```diff
@@ -29,50 +29,63 @@ import { StatusBarPopover } from "../statusbar/popup/StatusBarPopover";
...
+    }
+  );
 
-  return "placeholder";
+  // eslint-disable-next-line deprecation/deprecation
+  return modes.length === 1 ? (
+    <Icon iconSpec={modes[0].iconSpec} />
+  ) : (
+    <Icon iconSpec={snapModeKeypoint} />
+  );
```
ignas-k, line 88: Is this needed for this PR?

ui/appui-react/src/test/statusfields/SnapMode.test.tsx
```diff
@@ -69,12 +68,31 @@ describe("SnapModeField", () => {
...
+    const { container } = render(
+      <Provider store={TestUtils.store}>
+        <SnapModeField availableSnapModes={[SnapMode.Center]} />
+      </Provider>
+    );
+
+    await theUserTo.click(screen.getByText("snapModeField.snapMode"));
+    // Bisector snap mode should not be present.
+    expect(screen.queryByText("snapModeField.bisector")).toBeFalsy();
+    // Center snap mode be present.
```
ignas-k, line 90:
```suggestion
    // Center snap mode should be present.
```

ignas-k (changes requested):

ui/appui-react/src/appui-react/statusfields/SnapMode.tsx
mathieu-fournier: With the refactor I needed to rework the function to get the icon. 
I accounted for what will happen when multiple snap modes are active at the same time.

ignas-k: I understand that but this PR does not add multiple active snap modes. It's just a nitpick that this code would fit better in the next PR but it is not a big deal. Leaving the choice up to you, not  a big deal for us.

mathieu-fournier: It was something to always take into account because the consumer can set multiple snapmode via props. There was already a for multiple snap mode. 

The next PR, for the next major, will only change the default behaviour to allow multiple snaps.

ignas-k (approved):

PR #979, opened 2024-08-20
Increase `SelectionCountField` margins

### LINKED ISSUE #978: No padding for Selection Count right side when NOT on far right of status bar
### Describe the bug

As of appui 4.16.1, the selection count padding contains a fix which adds padding when it is the rightmost item on the status bar.  However, if it is in the middle somewhere, there is no padding added after the fixed min-width (5 digits) as shown below in screen capture.

### To Reproduce

Open any iModel and select more than 10000 elements.
Note that the status bar selection count field shows the count with no padding on right side

### Expected Behavior

Added this workaround which produces a good result  (shown in second screenshot below):

.uifw-statusFields-selectionCount {
  padding-right: var(--iui-size-xs) !important;
}

### Screenshots

Issue:
image

Suggested fix:
image

_No response_

### Additional context

_No response_

ignas-k:
## Changes
We had no padding on the right side of the `SelectionCountField` and it looked weird when there was a separator after it (the separator removes padding between different status bar fields). Changes:

- I added margins (`--iui-size-xs`) from both left and right sides now. I chose margins instead of paddings because padding is considered a part of the width so it basically decreased the `min-width` that we set (more specifically it would decrease the space reserved for more digits).
- Changed the `min-width` from `--iui-size-2xl` (equivalent to `4rem`) to `3.75rem`. This slightly decreases the min width of the component and is perfect to fit 5 digits without shifting other status bar content. I made this change because we previously had `padding-left` which is now switched to `margin-left` (again, the padding eats into the width and margin does not) so the freed up space was not needed anymore.

Adding some screenshots for comparison. Note that the first two images have separators.

### **Before:**
image
image
image

### **After:**
image
image
image

Fixes #978.

## Testing
Tested in test-app.

### DIFF
--- a/ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
+++ b/ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
@@ -7,7 +7,7 @@
   display: flex;
   align-items: center;
   min-width: 3.75rem;
-  margin: 0 var(--iui-size-xs) 0 var(--iui-size-xs);
+  margin: 0 var(--iui-size-xs);
 
   .icon {
     color: var(--iui-color-icon);
--- a/ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
+++ b/ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
@@ -6,8 +6,8 @@
   gap: var(--iui-size-2xs);
   display: flex;
   align-items: center;
-  min-width: var(--iui-size-2xl);
-  padding-left: var(--iui-size-2xs);
+  min-width: 3.75rem;
+  margin: 0 var(--iui-size-xs) 0 var(--iui-size-xs);
 
   .icon {
     color: var(--iui-color-icon);
--- a/common/changes/@itwin/appui-react/increase-selection-count-padding_2024-08-20-12-21.json
+++ b/common/changes/@itwin/appui-react/increase-selection-count-padding_2024-08-20-12-21.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "Increased margins of `SelectionCountField` component.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/appui-react"
+}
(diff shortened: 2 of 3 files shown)


ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
```diff
@@ -6,8 +6,8 @@
   gap: var(--iui-size-2xs);
   display: flex;
   align-items: center;
-  min-width: var(--iui-size-2xl);
-  padding-left: var(--iui-size-2xs);
+  min-width: 3.75rem;
+  margin: 0 var(--iui-size-xs) 0 var(--iui-size-xs);
```
FlyersPh9, line 10:
```suggestion
  margin: 0 var(--iui-size-xs);
```
Can be simplified.

ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
```diff
@@ -6,8 +6,8 @@
   gap: var(--iui-size-2xs);
   display: flex;
   align-items: center;
-  min-width: var(--iui-size-2xl);
-  padding-left: var(--iui-size-2xs);
+  min-width: 3.75rem;
```
FlyersPh9, line 9: Could use the [`ch` unit](https://caniuse.com/ch-unit).  If this is accounting for the icon as well you could try something like `calc(var(--iui-icon-size) + 4ch)` for example.

FlyersPh9 (approved):

ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
ignas-k: Applied.

ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
ignas-k: I think this would be a viable option but it seems that the `--iui-icon-size` (or `--iui-size-icon`) doesn't exist. Therefore using `ch` units does not really make sense when we have an icon. I will leave rem (as it is the unit used by iTwinUI) for now. We can change it later.

ignas-k:

ui/appui-react/src/appui-react/statusfields/SelectionCount.scss
FlyersPh9: Sorry, I didn't look up the actual variable and just gave that as an example before I signed off for the day.

PR #980, opened 2024-08-23
Fix new toolbars to handle event ids as case insensitive

ignas-k:
## Changes
Same problem as in #967. Noticed when debugging booster code. Studio recently enabled new toolbars so the problem now occured.

## Testing
Added test for `useConditionalProp`.


ignas-k:

Issue #981, opened 2024-08-23
EditorContainer does not commit when new value is set to the initial value


ignas-k:
### Describe the bug

When editor new value is being set to the initial/original value of the editor, it is not commited. The bug is caused by the code below because `propertyRecord.value` is always the same and therefore the condition passes. When switching to other values it works fine because `propertyRecord.value` and `newValue` are different.

### To Reproduce

1. Select the selection tool from the toolbar.
2. In tool settings, switch to another selection mode - crossing line or box.
3. Switch back to single selection mode.
4. Try to select elements in the view.
5. Notice that instead of using single selection mode, it's still selecting using previously chosen mode (crossing line or box). One of the ways to get back to single selection mode, is to clear selection, as that seem to reset the tool.

### Expected Behavior

Should commit the new value even if it the same as the initial/original value.

### Screenshots

![image](https://github.com/user-attachments/assets/26f87011-b0d7-4c8c-8f52-f801af3b6f37)

### Desktop (please complete the applicable information)

- AppUI version 4.16.0 and up.

### Additional context

Not really sure if `propertyRecord.value` is supposed to always stay the same or if it should change but the main issue is that when `newValue` is the same as the original `propertyRecord.value` it does not commit . Bug was introduced in #928.

Bug originally discovered in https://github.com/iTwin/itwinjs-backlog/issues/1219.

PR #982, opened 2024-08-26
Fixed tool setting editor not updating state after value change

saskliutas:
## Changes

Fixed tool setting property editor not updating state after value is changed. This resulted in editor component always having only initial value and not the current value. 

Fixes https://github.com/iTwin/appui/issues/981

## Testing

Did manual testing with various tools in standalone test-app. Fixed some tools to correctly update settings values as some of them were changing only in UI.


ignas-k (approved):

saskliutas:

PR #875, opened 2024-06-12
Deprecate APIs related to `ItemDefBase` class

GerardasB:
## Changes

This PR deprecates APIs related to the `ItemDefBase` class. As a replacement specific item types like `ToolbarItem` should be used instead. See `NextVersion.md` or the API summary for a full list of deprecations and their suggested replacements.

## Testing

Added additional story to showcase `NestedFrontstage`.

PR #989, opened 2024-08-29
Render `MenuItem` without `item` or `submenu`

### LINKED ISSUE #984: `CursorMenuItemProps` alternative for the deprecation of `item?: CommandItemProps;` not working with `UiFramework.openCursorMenu`
### Describe the bug

After updating our usage of `CursorMenuItemProps` to no longer reference the deprecated `item?: CommandItemProps;` we are finding that the `UiFramework.openCursorMenu` fails to render the items. We took the advise of the comment above the deprecated field which reads `Use properties of this object instead.`; however, despite moving the `label`, `icon` and `execute` properties to be inline with the rest of the `CurosrMenuItemProps` the issue persists. 

From debugging the issue, it appears that the `MenuItem` class that is leveraged to produce the menu items that `UiFramework.openCursorMenu` uses does not have logic to find the `label`, `iconSpec`, `badgeType`, `badgeKind`, and `isDisabled` unless the props can be found on the `item` field that is deprecated
see https://github.com/iTwin/appui/blob/b6a5e1810fc9c23327dfb348df5383e7f9ef8e74/ui/appui-react/src/appui-react/shared/MenuItem.tsx#L126

### To Reproduce

1. have a viewer instance running that is at least on appui v4.15 with a menu built from `CursorMenuItemProps` where the `item` field has been replaced by merging its props into the object itself.
2. have a way to visualize these menu items by passing them to `UiFramework.openCursorMenu` 
3. notice that the menu items do not appear and only a thin vertical line where the menu would otherwise be shown is visible

### Expected Behavior

The expectation is that the menu shows exactly the same as it does when the `CommandItemProps` are on the `item` field instead of merged inline with the rest of the `CursorMenuItemProps`

### Screenshots

Before:
image

After:
image

- AppUi v4.15 or newer

### Additional context

_No response_

ignas-k:
## Changes
In #875 the `item` prop of `CursorMenuItemProps` was deprecated and replacement props added. However, `MenuItemHelpers.createMenuItemNode` was still checking if the `MenuItem` had `item` or `submenu`. Using the new props this condition would fail and the function would return null instead of a component node.

I changed the function to render `ContextMenuItem` by default (without any condition) so this does change the behaviour slightly. Before if the `actionItem` or `submenu` was not set it would return null but now it will return a normal node. I believe this is fine because `MenuItem` constructor already has a check that at least one of `item`, `submenu` or `execute` prop has to be used or it throws an error. So technically it should not even be possible to pass `MenuItem` without required properties to the `MenuItemHelpers.createMenuItemNode` function.

Fixes #984.

## Testing
Added a test and changed component examples to adjust for the deprecation.

### DIFF
--- a/common/changes/@itwin/appui-react/fix-menu-item-rendering_2024-08-29-06-11.json
+++ b/common/changes/@itwin/appui-react/fix-menu-item-rendering_2024-08-29-06-11.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "Render `MenuItem` without `item` or `submenu` props.",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/appui-react"
+}
--- a/docs/changehistory/NextVersion.md
+++ b/docs/changehistory/NextVersion.md
@@ -4,9 +4,14 @@ Table of contents:
 
 - [@itwin/appui-react](#itwinappui-react)
   - [Changes](#changes)
+  - [Fixes](#fixes)
 
 ## @itwin/appui-react
 
 ### Changes
 
 - Allow to set the available snap modes in `SnapModeField` component. [#974](https://github.com/iTwin/appui/pull/974)
+
+### Fixes
+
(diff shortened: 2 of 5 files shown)


GerardasB (approved):

GerardasB:

PR #994, opened 2024-09-03
Fix link style load timing issue in window popups

### LINKED ISSUE #986: Popout styles load after a delay
### Describe the bug

I'm not 100% sure when this started, looks around 4.15, we dont control the application in which our packages are used. (Feels related to https://github.com/iTwin/appui/pull/940, but I don't know if this is published yet so maybe 940 would be a fix for it ?)

We are using viewport in popouts, and in order for these viewport to be working correctly, the surrounding css needs to be present so the display is flex (otherwise the viewport component does not expand all the height available) 

With numerous test trying to understand what's going on, I noticed that when it does work, the styles from the classes appears to be applied at the viewport creation, while otherwise the viewport height is 0 as no styles from classes appears to be applied.

### To Reproduce

I dont have an easy to reproduce path to recreate this, please contact me so I can demo and maybe we can recreate a clear path to reproduce. (Maybe a simple viewport in a popout widget would work)

### Expected Behavior

The styles (classes) are downloaded and applied before the widget content is rendered.

### Screenshots

_No response_

Chrome, 
AppUI 4.16.1

### Additional context

Testing with overriding the ViewportComponent minheight and maxheight (through chrome dev tools overrides) always cause this to work correctly. I have tested with providing css that would cause this through the classes, but this css does not appear to be loaded when the component is mounted. So no "class" solution works over this. Our solution to this exact problem (which was much less common before) would be to override the styles directly to the viewport component, but it requires us to make a lot of changes, so if this could be fixed in a patch in AppUI as it used to work, it would be nice...

Ultimately this is a timing problem.

GerardasB:
## Changes

This PR fixes #986 a timing issue where stylesheets are not loaded in a popout widget before the React lifecycle runs, causing issues in the Viewport `size != 0` assertion.

## Testing

Added additional e2e tests.


e2e-tests/tests/editors/editors.test.ts-snapshots/Editor-Primitive-number-numeric-input-UiAbstract-InputEditorSize-UiAbstract-Range-default-visual-1-chromium-linux.png
GerardasB, line 1: This is not a regression. Seems like some editors are not rendered with correct default values when strict mode is disabled.

PR #996, opened 2024-09-04
Add new tab dragging logic

ignas-k:
WIP

Know issues to resolve:
```[tasklist]
### Tasks
- [ ] Docked tool settings dragging is buggy with the new API. Probably need to make the docked tool settings draggable with drag and drop API instead of instantly switching it to a floating widget. 
- [ ] Missing implementation for dragging tabs to panels and sections.
- [ ] Esc press while dragging makes the widget pop out.
```


GerardasB:
Closing as stale, we'll need to revisit this in the future

Issue #829, opened 2024-05-06
Deprecate static package initializers


GerardasB:
### Describe the Solution you'd like

Deprecate static package initializers (i.e. `UiCore.initialize()`, `UiFramework.initialize()`). Components should use `LocalizationProvider` for translations.

PR #1000, opened 2024-09-05
Reduce `UIComponents.translate` static usage

saskliutas:
## Changes

Part of https://github.com/iTwin/appui/issues/829
Remove `UIComponents.translate` static usage in FilterBuilder and PropertyGrid components.

## Testing

Manually tested in storybook.


ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
```diff
@@ -358,7 +365,13 @@ function validateRules(
       const errorMessage = ruleValidator
         ? ruleValidator(item)
         : defaultPropertyFilterBuilderRuleValidator(item);
-      if (errorMessage) ruleIdsAndErrorMessages.set(item.id, errorMessage);
+      if (errorMessage)
+        ruleIdsAndErrorMessages.set(
+          item.id,
+          errorMessage.startsWith("filterBuilder.errorMessages")
```
GerardasB, line 371: Seems a bit exotic, would it make more sense to wrap/convert `defaultPropertyFilterBuilderRuleValidator` to hook and keep 2 args in `validateRules`?

common/api/components-react.api.md
```diff
@@ -2406,20 +2401,10 @@ export abstract class PropertyRecordDataFiltererBase extends PropertyDataFiltere
...
-    componentDidMount(): void;
-    // (undocumented)
-    componentDidUpdate(prevProps: PropertyRendererProps): void;
-    // (undocumented)
-    static getLabelOffset(indentation?: number, orientation?: Orientation, width?: number, columnRatio?: number, minColumnLabelWidth?: number): number;
-    // (undocumented)
-    render(): React_3.JSX.Element;
-    // (undocumented)
-    readonly state: Readonly<PropertyRendererState>;
-    updateDisplayValueAsEditor(props: PropertyRendererProps): void;
```
GerardasB, line 2421: Do we need `useImperativeHandle` here or are we hoping this is not used?

GerardasB (approved):

ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
saskliutas: `defaultPropertyFilterBuilderRuleValidator` might be used by FilterBuilder consumers (presentation-components uses it) so it can be simply converted into a hook. Added new `useDefaultPropertyFilterBuilderRuleValidator` and deprecated `defaultPropertyFilterBuilderRuleValidator`. I think it can be removed in 5.0 as it's only a beta. What do you think?

common/api/components-react.api.md
saskliutas: I hope this is not used. I could not find references to it in AppUI repo and I hope it is not used by any consumer. I even think `PropertyRenderer` is not used directly. Only inside PropertyGrid.

ui/components-react/src/components-react/filter-builder/FilterBuilderState.ts
GerardasB: Sounds good, we will break some of the existing `@beta` APIs with this PR as well, i.e. `getPropertyFilterBuilderOperatorLabel`, but I assume these are not expected to be used by anyone.

saskliutas: Yes, that one I don't expect to be used.

PR #1003, opened 2024-09-06
Fix `Toolbar` to avoid rendering of borders when there are no items

### LINKED ISSUE #1002: Two "dots" showing up in non-viewer frontstage
### Describe the bug

Seeing two dots that shouldn't be there in the non-viewer frontstage where the upper left tools usually appear.

### To Reproduce

_No response_

### Expected Behavior

No dots

### Screenshots

image

I have tracked it down to these two divs:
```html
<div class="_iui3134-surface uifw-toolbar-group-toolGroup uifw-vertical" style=""></div>
<div class="_iui3134-surface uifw-toolbar-group-toolGroup uifw-horizontal" style=""></div>
```

Dev tools hierarchy:
image

Running in Electron, AppUI 4.15.5.

### Additional context

Reported in internal Teams channel

GerardasB:
## Changes

This PR fixes #1002 to avoid rendering a dot artifact when the toolbar is empty (with `newToolbars` preview feature enabled).

## Testing

Added additional story.

### DIFF
--- a/common/changes/@itwin/appui-react/fix-1002_2024-09-06-13-04.json
+++ b/common/changes/@itwin/appui-react/fix-1002_2024-09-06-13-04.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "Fix `Toolbar` to avoid rendering of borders artifact",
+      "type": "none"
+    }
+  ],
+  "packageName": "@itwin/appui-react"
+}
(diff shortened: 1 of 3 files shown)


GerardasB:

mayank99 (approved):
This works, but I'm also curious why you didn't just render `null` when `items` is empty?

GerardasB:
> This works, but I'm also curious why you didn't just render `null` when `items` is empty?

I thought this was a safer option due to `ToolGroup` iterating the `children` React node and code to handle the overflow.

PR #1005, opened 2024-09-06
Fix `copyStyles` to ignore errors on external resource links

GerardasB:
## Changes

This PR fixes a potential issue where a popout widget is never rendered if there's an error with one of the external resource links (stylesheets).

This was reported by one of the AppUI consumers, who managed to get around this issue by fixing the resource.

## Testing

Tested manually via the `test-app`.

### DIFF
--- a/ui/appui-react/src/appui-react/childwindow/CopyStyles.ts
+++ b/ui/appui-react/src/appui-react/childwindow/CopyStyles.ts
@@ -2,6 +2,8 @@
  * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
+import { Logger } from "@itwin/core-bentley";
+import { UiFramework } from "../UiFramework";
 
 /** Copies the CSS style sheets from source document into the target document.
  * @internal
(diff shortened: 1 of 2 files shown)


GerardasB:

ui/appui-react/src/appui-react/childwindow/CopyStyles.ts
```diff
@@ -22,12 +24,19 @@ export async function copyStyles(
...
       promises.push(
-        new Promise((resolve, reject) => {
+        new Promise((resolve) => {
           clonedNode.onload = () => {
             resolve();
           };
-          clonedNode.onerror = (error) => {
-            reject(error);
+          clonedNode.onerror = () => {
+            Logger.logWarning(
```
mayank99, line 32: I think this should at least show a `console.error`

mayank99 (approved):

ui/appui-react/src/appui-react/childwindow/CopyStyles.ts
GerardasB: It will, depending on how the logger is configured. We could increase `LogLevel` to `Error` (instead of a `Warning`) though. We try to avoid direct `console` calls.

GerardasB: Changed to `logError()`

PR #1034, opened 2024-09-16
Add condition to PropertyFilterBuilderRenderer to allow last rule deletion

morganadrales:
## Changes

Added a prop called `allowLastRuleDelete` that will enable the deletion of the last remaining rule in a group.

## Testing

Built and tested locally with the Pineapple groups/query-builder component.


saskliutas (changes requested):
Looks good. Can you add unit test for `PropertyFilterBuilderActions.removeItem` in `FilterBuilderState.test.ts` to test the new behavior?

morganadrales:
> Looks good. Can you add unit test for `PropertyFilterBuilderActions.removeItem` in `FilterBuilderState.test.ts` to test the new behavior?

Added and also fixed a test that was failing in `FilterBuilderRule.test.tsx` due to not expecting that `allowLastRuleDelete` prop to be passed in the `removeRule` function

GerardasB (approved):
LGTM, describe the change in `NextVersion.md` as well

saskliutas (approved):

PR #1042, opened 2024-09-23
Updated `MessageManager.addMessage` to only add unique messages

GerardasB:
## Changes

This PR changes `MessageManager.addMessage()` to ignore already active messages displayed to the user (for all message types, like `Toast` or `Sticky`). This API is used by various tools indirectly via `IModelApp.notifications.outputMessage` when `AppNotificationManager` is set up. This change should prevent the same message from being displayed multiple times unnecessarily.

Ideally, tools and other consumers should throttle the messages manually i.e. in mouse wheel scroll handlers.
In the future, an option argument can be added if there is a use case where multiple equal messages need to be added.

## Testing

Additional unit tests were added and tested via the test app.

### DIFF
--- a/common/changes/@itwin/appui-react/add-message-once_2024-09-23-12-30.json
+++ b/common/changes/@itwin/appui-react/add-message-once_2024-09-23-12-30.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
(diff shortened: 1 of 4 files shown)


saskliutas (approved):

GerardasB:

PR #1048, opened 2024-09-25
Update `package.json` to use Node.js field definitions

GerardasB:
## Changes

This PR adds an [`exports`](https://nodejs.org/api/packages.html#exports) field to `package.json` to prevent consumers from importing `@internal` APIs from unsupported submodules. Supported import paths: 
- main entry point (i.e. `@itwin/appui-react`)
- `package.json` subpath (i.e. `@itwin/appui-react/package.json`). 

Additional import paths that are available until the next major version to facilitate the **AppUI 5.0** adoption: 
- ESM submodule (i.e. `@itwin/appui-react/esm`)
- CJS submodule (i.e. `@itwin/appui-react/cjs`)
- All SCSS files are exported via using `sass` [custom condition](https://sass-lang.com/blog/rfc-package-importer/#details) (i.e. `@itwin/core-react/lib/cjs/core-react/typography`).

Additionally this PR:
- Renames the `typings` field to `types` in `package.json`
- Bumps `eslint` version
- Adds [`type`](https://nodejs.org/api/packages.html#type) field to use `ESM` by default
- Adds a script to create a dummy `package.json` with `"type": "commonjs"` to `./lib/cjs` directory of each package

[`moduleResolution`](https://www.typescriptlang.org/tsconfig/#moduleResolution) used by AppUI packages will be updated with a separate PR, for now, it should be enough to build the test-app and storybook to validate the correctness of imports.

## Testing

Build and run `storybook` and `test-app`.


mayank99 (commented):
The changes in `package.json` files look ok, but there is a slight issue: both ESM and CJS files use the `.js` file extension. CJS files should use the `.cjs` extension or have a dummy `package.json` with `"type": "commonjs"` (see [example](https://unpkg.com/browse/@itwin/itwinui-react@3.15.0/cjs/package.json)).

GerardasB:
> The changes in `package.json` files look ok, but there is a slight issue: both ESM and CJS files use the `.js` file extension. CJS files should use the `.cjs` extension or have a dummy `package.json` with `"type": "commonjs"` (see [example](https://unpkg.com/browse/@itwin/itwinui-react@3.15.0/cjs/package.json)).

There's a script that adds one, see https://github.com/iTwin/appui/pull/1048/commits/1c7005abeba842369af9474027b9e70aadfb1cbd

mayank99:
> There's a script that adds the dummy package.json, see [1c7005a](https://github.com/iTwin/appui/commit/1c7005abeba842369af9474027b9e70aadfb1cbd)

Ah ok, this would be good to mention in the PR description.

mayank99 (approved):
LGTM!

non-blocking: One more thing I want to point out is that for proper ESM support, [file extensions are mandatory](https://nodejs.org/api/esm.html#mandatory-file-extensions) in import specifiers. This might not be an issue when using bundlers, but it could become an issue for some test runners.

GerardasB:
> LGTM!
> 
> non-blocking: One more thing I want to point out is that for proper ESM support, [file extensions are mandatory](https://nodejs.org/api/esm.html#mandatory-file-extensions) in import specifiers. This might not be an issue when using bundlers, but it could become an issue for some test runners or pure SSR.

Thanks, I've noticed that while changing `moduleResolution` and decided to update that in a separate PR due to the amount of changes required.

PR #1050, opened 2024-09-25
Drop support for iTwin.js 3.x

### LINKED ISSUE #826: Drop support for `iTwin.js 3.x`

Drop support for `iTwin.js 3.x` which is scheduled to enter end-of-life state in 2024-05-22 https://www.itwinjs.org/learning/api-support-policies/#version-support-status

GerardasB:
## Changes

This PR fixes #826 by dropping support for iTwin.js 3.x which is in [end-of-life phase](https://www.itwinjs.org/learning/api-support-policies/#version-support-status).

## Testing

N/A


saskliutas (approved):

aruniverse:
should we also update min core 4.x peer deps to be the latest?

GerardasB:
> This PR fixes #826 by dropping support for iTwin.js 3.x which is in [end-of-life phase](https://www.itwinjs.org/learning/api-support-policies/#version-support-status).

Let's stick to major version increments unless there's a good reason (i.e. AppUI needs a specific feature from one of the minor versions). This gives our consumers who are keeping up with dependency upgrades more time to upgrade gradually (on their schedule), instead of updating everything at once.

PR #1056, opened 2024-10-01
Update module resolution and provide mandatory file extensions

GerardasB:
## Changes

This PR changes [`moduleResolution`](https://www.typescriptlang.org/tsconfig/#moduleResolution) used in AppUI packages from `Node` to `NodeNext`. This change required mandatory file extensions to be provided in import declarations https://nodejs.org/api/esm.html#mandatory-file-extensions
Additionally, the [`module`](https://www.typescriptlang.org/tsconfig/#module) is changed as well to be compatible with the new `moduleResolution`.

`classnames` and `immer` package versions are updated to solve some of the export problems caused by new module resolution.

CJS modules are dropped from a private  `@itwin/appui-test-providers` package.

## Testing

N/A


ui/appui-react/tsconfig.json
```diff
@@ -3,6 +3,8 @@
   "compilerOptions": {
     "jsx": "react",
     "resolveJsonModule": true,
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
     "types": ["vitest/globals"]
   },
   "include": ["./src/**/*.ts*"]
```
mayank99, line 10: While the new additions to this file look good, it is generally advised to have separate tsconfig files for build vs test.

See https://kettanaito.com/blog/one-thing-nobody-explained-to-you-about-typescript

mayank99: There are also other changes you might want to make to the tsconfig, such as using the new JSX transform or adding `verbatimModuleSyntax`.

mayank99 (commented):
I haven't looked deeply through all the file changes (I'm assuming that build passes means everything is generally ok), but I did have one comment on the `tsconfig`

mayank99 (approved):

PR #1064, opened 2024-10-03
Update to TypeScript 5.6

wgoehrig:
It's TypeScript update time again!

Since our last update, there have been 3 major TypeScript versions released (remember that TypeScript does _not_ follow semver).  As always, the release notes are the best way to catch up on what's changed since 5.3:
- [TS 5.4 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-4.html)
- [TS 5.5 release notes](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html)
- [TS 5.6 release notes](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/)

Even more than usual, these new versions bring better compiler performance and (especially) better type inferencing.  Other highlights include:

- [Regex Syntax Checking](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#regular-expression-syntax-checking) - TypeScript can now tell you when you have syntax errors in your regular expressions.
- [Isolated Declarations](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-5-5.html#isolated-declarations) - We're certainly not ready to opt-in to this new, stricter enforcement for type annotations, but someday this may allow us to _fully_ parallelize the _entire_ monorepo's build!
- [Disallowed Nullish and Truthy Checks](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#disallowed-nullish-and-truthy-checks) - Some expressions that always evaluate either truthy or nullish will now be flagged as errors.  Very handy to catch these at compile time since these can be easy mistakes to make.
- [Iterator Helper Methods](https://devblogs.microsoft.com/typescript/announcing-typescript-5-6/#iterator-helper-methods) - This implements support for a very cool new ECMAScript proposal that will make it _much_ easier to work with generators and iterators.

Most of the breaking compiler changes were due to the newly-disallowed nullish and truthy checks affecting some patterns in testing/debugging code.  However, better type inferencing also means some new lint errors, and while I'm separately updating ESLint in #1063 there are of course a few lint failures caused just by updating our TypeScript version.


ui/appui-react/src/appui-react/stagepanels/StagePanelLocation.ts
```diff
@@ -11,7 +11,7 @@
  */
 export enum StagePanelLocation {
   Top = 101,
-  Left,
-  Right,
-  Bottom,
+  Left = 103,
```
wgoehrig, line 14: Not sure if this is the right way to handle this, since it's probably a breaking change, but TypeScript is now correctly pointing out that the definition for `StagePanelLocation` in appui-react is out of sync with [the definition in appui-abstract](https://github.com/iTwin/itwinjs-core/blob/master/ui/appui-abstract/src/appui-abstract/widget/StagePanel.ts#L23)!

Was originally flagged by TypeScript here: https://github.com/iTwin/appui/blob/20eba19c8092955e3e86d9afd4f282e571e6c308/ui/appui-react/src/test/ui-items-provider/UiItemsManager.test.tsx#L776-L780

GerardasB (approved):

ui/appui-react/src/appui-react/stagepanels/StagePanelLocation.ts
GerardasB: **Ouch!** That's a good catch by typescript! I think we can simply go with this for the major release (would only break if someone were to use numeric values directly)

GerardasB: After a second thought:
I think we should revert this change since there's no need to keep the two enums in sync. We are mapping enum values correctly: https://github.com/iTwin/appui/blob/0eeaa10787b0612d6364d5427cd2afef835b88cc/ui/appui-react/src/appui-react/ui-items-provider/AbstractUiItemsManager.ts#L661-L697

GerardasB: In the test file, simply use enum from `@itwin/appui-abstract` with `AbstractUiItemsManager`

wgoehrig: Made that change in https://github.com/iTwin/appui/pull/1064/commits/36eeed35f08a38bc086e4285dcd086fb670b3c04

PR #1067, opened 2024-10-08
Fix user-select issue when resizing a stage panel

### LINKED ISSUE #1057: Widget content gets selected or picked up in a drag motion on panel resize
### Describe the bug

- All widget content gets highlighted when panel is resized.
- When panel is resized by dragging, widget content is picked up in that drag motion rather than the actual resize of a panel.

### To Reproduce

1. Make sure you use Chromium based browser.
2. Try to resize any of the panels a few times.

### Expected Behavior

_No response_

### Screenshots

https://github.com/user-attachments/assets/96115f7e-ce28-48ea-a2bc-fc19d3bd181e

- iTwin.js 4.9.0
- AppUI 4.17.0
- iTwinUI 3.14.2
- Chrome 129.0.6668.71

### Additional context

_No response_

GerardasB:
## Changes

This PR fixes #1057 by setting the [`user-select`](https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) property on a stage panel resize handle to avoid content selection when resizing the panel.

Additionally, `-webkit-user-select` property is added with each `user-select` to fix widget drag/resize issues in Safari.

## Testing

Tested manually in test-app.

### DIFF
--- a/apps/test-providers/src/ui/content/SampleContentControl.scss
+++ b/apps/test-providers/src/ui/content/SampleContentControl.scss
@@ -9,4 +9,5 @@
   width: 100%;
   justify-content: center;
   user-select: none;
+  -webkit-user-select: none;
 }
--- a/ui/appui-react/src/appui-react/layout/widget-panels/Grip.scss
+++ b/ui/appui-react/src/appui-react/layout/widget-panels/Grip.scss
@@ -18,6 +18,8 @@ $line-grip-width: calc(#{$nz-grip-width} - 2px);
     position: relative;
     display: flex;
     transition: 0.2s ease-out;
+    user-select: none;
+    -webkit-user-select: none;
 
     > .nz-handle {
       position: absolute;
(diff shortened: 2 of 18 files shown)


ui/appui-react/src/appui-react/layout/widget-panels/Grip.scss
```diff
@@ -18,6 +18,8 @@ $line-grip-width: calc(#{$nz-grip-width} - 2px);
     position: relative;
     display: flex;
     transition: 0.2s ease-out;
+    user-select: none;
```
mayank99, line 21: Instead of adding `user-select: none` in all places, another solution would be to temporarily add it only on `<body>` (using JS) for the duration of dragging. This would also allow the user to select the text when they're not dragging.

apps/test-providers/src/ui/content/SampleContentControl.scss
```diff
@@ -9,4 +9,5 @@
   width: 100%;
   justify-content: center;
   user-select: none;
+  -webkit-user-select: none;
```
mayank99, line 12: You might want to add a build step that automatically adds vendor prefixes (in the future).

mayank99 (approved):

GerardasB:

PR #1071, opened 2024-10-09
Remove `activateDroppedTab` preview feature

GerardasB:
## Changes

This PR enables the [`activateDroppedTab`](https://github.com/iTwin/appui/discussions/679) behavior by default and removes the preview feature.
Additionally, a widget tab's drag-and-drop logic is updated to activate the dropped tab only if it was active when the drag interaction was initiated.

## Testing

Additional unit tests were added and tested manually via the test app.

### DIFF
--- a/ui/appui-react/src/appui-react/layout/state/internal/TabStateHelpers.ts
+++ b/ui/appui-react/src/appui-react/layout/state/internal/TabStateHelpers.ts
@@ -10,7 +10,7 @@ import type { Draft } from "immer";
 import { castDraft, produce } from "immer";
 import { UiError } from "@itwin/appui-abstract";
 import type { NineZoneState } from "../NineZoneState.js";
-import type { DraggedTabState, TabState } from "../TabState.js";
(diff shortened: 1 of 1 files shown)


ui/appui-react/src/appui-react/layout/state/internal/TabStateHelpers.ts
```diff
@@ -53,6 +53,7 @@ export function createDraggedTabState(
       widgetIndex: 0,
     },
     position: { x: 0, y: 0 },
+    active: true,
```
saskliutas, line 56: Is this necessary? Looks like `createDraggedTabState` is always called with `active` in `args`.

saskliutas (approved):

ui/appui-react/src/appui-react/layout/state/internal/TabStateHelpers.ts
GerardasB: Good point, I moved it to test utils.

PR #1081, opened 2024-10-21
Drop support for CommonJS modules

GerardasB:
## Changes

This PR removes CommonJS modules from all AppUI packages:
- Removed all `esm` and `cjs` specific npm scripts
- Removed `build-cjs` script
- Moved `package-cjs` script to test-app/scripts since it's still needed for the backend

Additionally, the [`exports`](https://nodejs.org/api/packages.html#exports) field is updated to support legacy `cjs/esm` import paths in SCSS files. I.e. `@import "@itwin/core-react/lib/cjs/core-react/base/base";` will continue to resolve correctly. However, consumers should avoid importing [deprecated](https://github.com/iTwin/appui/releases/tag/release%2F4.17.0) SCSS partial files since these will not be accessible from AppUI 6.0.

## Testing

N/A


GerardasB:
Blocker: https://github.com/iTwin/itwinjs-core/pull/7288

ui/components-react/package.json
```diff
@@ -3,24 +3,19 @@
...
-  "module": "lib/esm/components-react.js",
-  "types": "lib/cjs/components-react",
   "exports": {
-    ".": {
-      "import": "./lib/esm/components-react.js",
-      "require": "./lib/cjs/components-react.js"
+    ".": "./lib/components-react.js",
+    "./internal": "./lib/internal.js",
+    "./package.json": "./package.json",
+    "./lib/cjs/*": {
```
saskliutas, line 10: Why `/lib/cjs` entry is left for sass files?

GerardasB: Previously consumers would import SCSS files via these paths and I didn't want to force everyone to upgrade at once (I will leave a recommendation in NextVersion.md to get rid of the SASS imports instead)

GerardasB: I.e. `@import "@itwin/core-react/lib/cjs/core-react/style/themecolors.scss"` continues to work and resolves to `@import "@itwin/core-react/lib/core-react/style/themecolors.scss"`

saskliutas (approved):

PR #1090, opened 2024-10-30
[Property-grid]: add callback to show only editor

MartynasStrazdas:
Closes https://github.com/iTwin/presentation/issues/738

## Changes

- Added callback to `VirtualizedPropertyGrid` to determine which property rows should show only the editor.

## Testing

- Added tests.


common/api/components-react.api.md
```diff
@@ -260,6 +260,7 @@ export interface CommonPropertyGridProps extends CommonProps {
     onPropertyUpdated?: (args: PropertyUpdatedArgs, category: PropertyCategory) => Promise<boolean>;
     orientation?: Orientation;
     propertyValueRendererManager?: PropertyValueRendererManager;
+    showOnlyEditor?: (editorName?: string) => boolean;
```
saskliutas, line 263: I think `showOnlyEditor` should get whole `PropertyRecord` instead of only editor name.

MartynasStrazdas: Adjusted.

ui/components-react/src/components-react/propertygrid/internal/flat-properties/FlatPropertyRenderer.tsx
```diff
@@ -179,34 +158,41 @@ const DisplayValue: React.FC<DisplayValueProps> = (props) => {
     props.onHeightChanged
   );
 
-  if (props.isEditing) {
+  if (
```
saskliutas, line 161: This `if` condition is very confusing. Is it possible that `isEditing` is `true` when `isPropertyEditingEnabled` is `false`? If no than we could check if `isPropertyEditingEnabled` is `false` we don't need to check any other stuff. 

Also what does it mean when `isEditing` is `true` and `showOnlyEditor` also returns `true`? Is it a valid case?

saskliutas (changes requested):
Can you add a story using the new `showOnlyEditor`?

ui/components-react/src/components-react/propertygrid/internal/flat-properties/FlatPropertyRenderer.tsx
MartynasStrazdas: `IsEditing` is not going to be set if `isPropertyEditingEnabled` is false and will also be false, but `isEditing` is going to be set to true only for one row at a time so we still need to check it if the specific row should have editing, and we need `isPropertyEditingEnabled` in case it is false so the `showOnlyEditor` would not show editors.

For the second part I check if `showOnlyEditor` function exists and not that it returns true, which a case where it exists and returns false we would still want to focus if editing was manually enabled. I guess we could save the result of showOnlyEditor and use it instead of checking if function exists.

saskliutas: So we could simplify this doing something like `props.isPropertyEditingEnabled && (props.isEditing || (props.showOnlyEditor && props.showOnlyEditor(props.propertyRecord)))`. I think this is much more readable. Also I think we should reuse `props.showOnlyEditor` result when determining if focus should be applied. I guess we could invoke `props.showOnlyEditor` again but I think we could further simplify everything by computing editor "visibility" before with something like this:

```ts
const computeEditorVisibility = () => {
  if (props.isPropertyEditingEnabled) {
    return "hidden";
  }
  if (props.showOnlyEditor && props.showOnlyEditor(props.propertyRecord)) {
    return "alwaysVisible";
  }
  return props.isEditing ? "visible" : "hidden";
};

const editorVisibility = computeEditorVisibility();
```

This would make following check very simple:
`if (editorVisibiliy !== "hidden")`
`const setFocus = editorVisibility !== "alwaysVisibile"`

MartynasStrazdas: We only want to focus on item when user is editing, in this case if `isEditing` is set to true we focus on the input if the editor is not shown at all times it would render the editor and focus it aditionally, if the editor is shown at all times the focus does not matter as clicking the editor will focus it so we would focus on it again. Changed to `setFocus={props.isEditing}`

common/api/components-react.api.md
```diff
@@ -260,6 +260,7 @@ export interface CommonPropertyGridProps extends CommonProps {
     onPropertyUpdated?: (args: PropertyUpdatedArgs, category: PropertyCategory) => Promise<boolean>;
     orientation?: Orientation;
     propertyValueRendererManager?: PropertyValueRendererManager;
+    showOnlyEditor?: (property: PropertyRecord) => boolean;
```
saskliutas, line 263: What do you think about naming this: `alwaysShowEditor`? `showOnlyEditor` sounds a little bit misleading as property label is also shown. 

Also can you add a story with some editors always visible? At least for default checkbox and toggle editors.

saskliutas (changes requested):

common/api/components-react.api.md
MartynasStrazdas: added a story, which revealed a bug, added test to cover it.

saskliutas (approved):

GerardasB (approved):
Entry in `NextVersion.md` would be appreciated.

GerardasB:

PR #1096, opened 2024-11-05
Fix restore layout without an active frontstage

### LINKED ISSUE #1088: FronstageDef.restoreLayout does not restore the layout when there is no active frontstage
### Describe the bug

`FronstageDef.restoreLayout()` does not restore the layout when there is no active frontstage. From what I understand, this happens because the `InternalFrontstageManager.onFrontstageRestoreLayoutEvent` listener is only active when there is an active frontstage.

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Screenshots

_No response_

- iTwin.js 4.9.3
- AppUI 4.17.1
- iTwinUI 3.11.3

### Additional context

Some additional context why this is an issue for us: 

In our use case we restore the layout from the settings route - meaning there is no active frontstage at the time. 
Currently to achieve that we use `internal` functions, which we are trying to remove, the only public alternative that I found was `FronstageDef.restoreLayout`.

GerardasB:
## Changes

This PR fixes #1088 by correctly clearing the state storage even when no active frontstage exists.

## Testing

Added a `settings` route to the test-app.


GerardasB:

DaumantasJankauskas (approved):

GerardasB:

PR #1101, opened 2024-11-06
cSpell maintenance: removed cSpell ignores from files and fixed spelling mistakes

MBudreviciusBentley:
## Changes

cSpell maintenance:
- removed cSpell ignore comments from separate files
- Fixed a lot of spelling mistakes through the project
- Updates cspell.json file

## Testing

To test the cSpell warnings used `cspell lint "**" --gitignore` command of [cSpell package](https://www.npmjs.com/package/cspell)


MBudreviciusBentley:
Need to discuss if cSpell should be added to CI builds.

.vscode/cSpell.json
```diff
@@ -1,121 +1,315 @@
...
+    "dockings",
+    "dont",
     "downvote",
+    "dragdrop",
+    "ecef",
     "ecschema",
     "ecsql",
-    "focusable",
-    "frontend",
+    "élêçtèd",
```
GerardasB, line 83: Should we simply disable checking for some of these? I.e. things like these translation strings, hex values or other one-offs? `// cspell:disable-next-line`

GerardasB (approved):
Nice, I think we should enable this in CI (I'd start from adding a few commands to `command-line.json`. Git hook would be helpful as well to catch any spelling issues early. We can do this in other/separate PRs.

.vscode/cSpell.json
MBudreviciusBentley: I think so. At least for the v1 of the pr removed all the `// cspell:disable-next-line` as the item requested, but it does seem as a way better approach in some places.

MBudreviciusBentley: Removed not real words from `cSpell.json` and used `// cspell:disable-next-line` instead

PR #1125, opened 2024-11-19
Copy SCSS files to `cjs` and `esm` directories for compatibility

GerardasB:
## Changes

Copy SCSS files to `cjs` and `esm` directories for backwards compatibility. 
Previously assumed https://sass-lang.com/documentation/js-api/classes/nodepackageimporter/ is not a viable solution, since it requires an explicit `pkg:` URL scheme https://sass-lang.com/documentation/at-rules/use/#pkg-ur-ls

Some bundlers/loaders do additional work to ensure that `exports` field is respected when resolving the imports (verified in `vite` and `sass-loader` for `webpack`). `esbuild-sass-plugin` does not respect the `exports` field and only built with custom `importMapper`. 

## Testing

N/A

### DIFF
--- a/common/changes/@itwin/appui-react/css-compat_2024-11-19-16-11.json
+++ b/common/changes/@itwin/appui-react/css-compat_2024-11-19-16-11.json
@@ -1,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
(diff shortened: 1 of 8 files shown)


GerardasB:

saskliutas (approved):

PR #1130, opened 2024-11-25
Cesium viewer sample

GerardasB:
## Changes

A sample application with the [cesium viewer](https://cesium.com/learn/cesiumjs/ref-doc/Viewer.html) rendered in a frontstage which consists a few other UI items: widget, toolbar button, status bar item.
In this application `IModelApp.startup` and `UiFramework.initialize` are not called.

PR will not be merged as it serves as an POC around using the widget system w/o  the `IModelApp` initialization.

Related Three.js viewer sample: https://github.com/iTwin/appui/pull/1119

## Testing

N/A

### DIFF
--- a/apps/cesium-app/index.html
+++ b/apps/cesium-app/index.html
@@ -1,0 +1,17 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta
+      name="viewport"
+      content="initial-scale=1,viewport-fit=cover,shrink-to-fit=no"
+    />
+    <title>AppUI Cesium App</title>
+  </head>
+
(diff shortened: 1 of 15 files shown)


calebmshafer:
@GerardasB what do you think about making a true PR to add the capability to do this? Are there other gaps that we'd need to work through?

GerardasB:
> @GerardasB what do you think about making a true PR to add the capability to do this? Are there other gaps that we'd need to work through?

Sure, I'll need to polish this PR a bit and we can merge it in.

GerardasB:
Closing as stale, this served as a POC. Instead of creating a separate app we could simply add a separate page/frontstage to the existing `test-app`.

PR #1147, opened 2024-12-09
Remove `@itwin/core-geometry` from `@itwin/components-react`

saskliutas:
## Changes

Inlined single usage of `@itwin/core-geometry` in `@itwin/components-react` to avoid having it as a peer dependency.

## Testing

N/A


saskliutas:
> Should we simply inline `clamp` instead?

I guess that also works. Made a change.

GerardasB (approved):

GerardasB:

PR #1156, opened 2024-12-13
Add `ModalFrontstageButton` component

GerardasB:
## Changes

This PR adds `backButton` property to `ModalFrontstageInfo` interface to allow specifying of a custom back button for a modal frontstage. Additionally `ModalFrontstageButton` component is added to maintain visual consistency between modal frontstages.

`ModalFrontstageButton` uses iTwinUI `IconButton` instead of outdated custom implementation.
Additionally, `ModalFrontstageButton` is used as a default back button if no `backButton` is provided via `ModalFrontstageInfo`.

## Testing

Added additional story.


GerardasB:

BeAliAslam:
Instead/in addition to this, is it possible to add a 'onBeforeExit' kind of event on the button click which users can handle via their provided function, and are able to cancel the back button operation if required.

GerardasB:
> Instead/in addition to this, is it possible to add a 'onBeforeExit' kind of event on the button click which users can handle via their provided function, and are able to cancel the back button operation if required.

You can do that in a custom `onClick` handler of your "custom back buttton", here's an example (added with this PR): https://itwin.github.io/appui/1156/?path=/story/frontstage-modalfrontstage--back-button

BeAliAslam:
[This story](https://itwin.github.io/appui/1156/?path=/story/frontstage-modalfrontstage--back-button) looks exactly like my use case.

PR #1157, opened 2024-12-13
Accudraw update fixes

mathieu-fournier:
# Changes

This PR goes hand in hand with some fixes that have been made in Core-frontend : [itwinjs-core/pull/7475](https://github.com/iTwin/itwinjs-core/pull/7475). 
This stabilizes AccuDraw behaviours in general. 

### Fixes : 
1- When moving the mouse, the focus only change between X or Y in rectangular mode only. 

2- *Revereted that fix, it will need to be done in another PR. See the comments.* Only updating input fields value when their state is dynamic. Reason : `onMotion` is supposed to be called way more often, like on every key typed. This prevents the input field from updating at the wrong moment. (Works with the core-frontend fix)

3- Removed the delay after typing in the input field, the visual update should be immediate. (Works with the core-frontend fix)

4- Input fields are automatically focused when the AccuDrawFieldContainer appears and the the compass mode changes. 
image

### Improvements :
5- Rectangular inputs no longer have colors.

6- We can enter letters in the Bearing angle input field. `N, S, E, W` for North, South, East and West.

7- Bearing angle input field automatically adds special characters `° ' "` to facilitate entering bearing angles. 

8- The Focus is now trapped in Accudraw input fields. To focus out of these fields, the user must press Escape. 

Thanks (:
Mathf


ui/appui-react/src/appui-react/accudraw/AccuDrawFieldContainer.tsx
```diff
@@ -206,7 +213,30 @@ export function AccuDrawFieldContainer(props: AccuDrawFieldContainerProps) {
...
+        IModelApp.accuDraw.setFocusItem(ItemField.DIST_Item);
+        break;
+      case ItemField.DIST_Item:
+        IModelApp.accuDraw.setFocusItem(ItemField.ANGLE_Item);
+        break;
+    }
+  }, [showZ]);
+
   const handleEscPressed = React.useCallback(() => {
+    IModelApp.accuDraw.unlockAllFields();
```
bbastings, line 239: I understand the focus out on ESC. Wondering the reason for unlocking all the fields too?

ui/appui-react/src/appui-react/accudraw/FrameworkAccuDraw.ts
```diff
@@ -307,11 +308,12 @@ export class FrameworkAccuDraw
   }
 
   private fieldValuesChanged(): void {
-    this.onFieldValueChange(ItemField.X_Item);
-    this.onFieldValueChange(ItemField.Y_Item);
-    this.onFieldValueChange(ItemField.Z_Item);
-    this.onFieldValueChange(ItemField.ANGLE_Item);
-    this.onFieldValueChange(ItemField.DIST_Item);
+    // Only change the value when in Dynamic mode. Other mode are "DontUpdate" when the input is locked and "Partial" when the user is typing.
+    (this.getKeyinStatus(ItemField.X_Item) === KeyinStatus.Dynamic) && this.onFieldValueChange(ItemField.X_Item);
```
bbastings, line 312: Checking Dynamic isn't doing anything here right? Because we determined it's always Dynamic currentlym, yes? KeyinStatus is also internal and this code will replaced by calling the new "processMotion" method once I open a new core PR...

mathieu-fournier: Oh, I thought that we would still need to do the KeyinStatus check. 
Should I remove it then ?

ui/appui-react/src/appui-react/accudraw/AccuDrawFieldContainer.tsx
mathieu-fournier: Usually, `Escape` is used to cancel the tool. So we thought that using `Escape` to cancel AccuDraw inputs might be the natural reflex.

ui/appui-react/src/appui-react/accudraw/FrameworkAccuDraw.ts
bbastings: Yeah, I'd revert this, it's not doing anything now and onMotion is going to end up like this:

public override onMotion(_ev: BeButtonEvent): void {
  if (UiFramework.isContextMenuOpen) // I don't know what this is, assume we still want to check it. :)
    return;
  this.processMotion();
}

ui/appui-react/src/appui-react/accudraw/AccuDrawFieldContainer.tsx
bbastings: If ESC cancels the tool, then unlocking the fields isn't necessary. If ESC is not canceling the tool, then I think it should leave the fields locked. If the user has a field locked but pressed ESC to set focus to tool settings or something, I think unlock the field would be unexpected...

mathieu-fournier: Alright, I'll remove it, anyway we can add it back at some point, if needed.

ui/appui-react/src/appui-react/accudraw/FrameworkAccuDraw.ts
mathieu-fournier: Oh, I just checked the implementation of the `processMotion` function you proposed and it does check the KeyinStatus. 
I'll do an update soon with that code.

mathieu-fournier: I removed the Dynamic check, this fix will need to be done in another PR with the new core-frontend API.

common/changes/@itwin/appui-react/accudraw_update_fixes_2024-12-13-20-55.json
```diff
@@ -0,0 +1,10 @@
+{
+  "changes": [
+    {
+      "packageName": "@itwin/appui-react",
+      "comment": "Accudraw input fields are now focused when the accudraw input field container becomes visible and when the compass mode changes. Fix : The focus does not change between X or Y when in polar mode. Now only updating input fields value when their state is dynamic. Made rectangular inputs color standard to iTwinUI. We can enter letters in the Bearing angle input field. Bearing angle input field automatically adds special characters to facilitate entering bearing angles. Removed the delay after typing in the input field, the visual update should be immediate. The Focus is now trapped in Accudraw input fields, when doing Tab. To focus out of these fields, the user must press Escape. Escape now unlocks all fields.     ",
```
GerardasB, line 5: This can contain multiple change objects.
Alternatively just briefly mention the changes and add extensive description to `NextVersion.md`.

ui/appui-react/src/appui-react/accudraw/AccuDrawFieldContainer.tsx
```diff
@@ -206,6 +213,28 @@ export function AccuDrawFieldContainer(props: AccuDrawFieldContainerProps) {
     []
   );
 
+  const focusNextField = React.useCallback((field: ItemField) => {
```
GerardasB, line 216: I think we can get rid of `ItemField` param in `onTabPressed` to align with other handlers. This switch can be removed (code blocks can simply be inlined in their respective `AccuDrawInputField` elements.

ui/appui-react/src/appui-react/accudraw/AccuDrawInputField.tsx
```diff
@@ -53,12 +55,12 @@ export interface AccuDrawInputFieldProps extends CommonProps {
...
   /** Triggered when the content is changed */
   onValueChanged: (stringValue: string) => void;
-  /** Frequency to poll for changes in value, in milliseconds */
-  valueChangedDelay?: number;
   /** Listens for <Enter> keypress */
   onEnterPressed?: () => void;
   /** Listens for <Esc> keypress */
   onEscPressed?: () => void;
+  /** Listens for <Tab> keypress */
+  onTabPressed?: (field: ItemField) => void;
```
GerardasB, line 63:
```suggestion
  onTabPressed?: () => void;
```

common/changes/@itwin/appui-react/accudraw_update_fixes_2024-12-13-20-55.json
mathieu-fournier: But it wont be in the next version (5.x)  it will be a patch from a previous version 4.x, right ?

GerardasB: This will have to be a new minor version, ideally a `5.1.0`.

ui/appui-react/src/appui-react/accudraw/AccuDrawInputField.tsx
```diff
@@ -106,69 +107,74 @@ const ForwardRefAccuDrawInput = React.forwardRef<
 
   const handleChange = React.useCallback(
     (event: React.ChangeEvent<HTMLInputElement>): void => {
-      const value = event.currentTarget.value;
+      let value = event.currentTarget.value;
 
       if (value === undefined) return;
 
       if (stringValue !== value) {
+        if (isBearingAngle) {
```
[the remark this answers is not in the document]
mathieu-fournier: Yes, It's only active when the unit of the input field is Bearing Angle.

mathieu-fournier:
> Additionally, I'd appreciate if we could update (in separate PR) `AccuDraw.md` to actually describe what accu draw is and how it is expected to work.

I'll do it in another PR, AccuDraw doc is probably already in Core. I'll link this doc and make a fresh doc for FrameworkAccuDraw

bbastings:
> AccuDraw doc is probably already in Core.

@mathieu-fournier There's an AccuDraw section in the learning docs (PrimitiveTool.md) but since there's no UI in core it's pretty bare bones.

GerardasB (approved):

mathieu-fournier:
> > AccuDraw doc is probably already in Core.
> 
> @mathieu-fournier There's an AccuDraw section in the learning docs (PrimitiveTool.md) but since there's no UI in core it's pretty bare bones.

Should we put the doc in core or appui ?

bbastings:
@mathieu-fournier Document in appui so you can reference/link ui classes, etc.

PR #1166, opened 2024-12-20
Proposal for new editors system

saskliutas:
# Proposal for new editor system

Proposal for the new editors system. This would replace existing editors system in order to simplify requirements for custom editors. Also it removes static EditorManager in favor of editors registry controlled through React context.

API overview:

- EditorsRegistry - Application should use it to register their on custom editors or custom editors provided by some dependencies so that they would be accessible to all components in the application that uses editors system:

  - `EditorsRegistryProvider` - adds supplied editors to the registry hold in `React` context. It supports nesting multiple `EditorsRegistryProvider` to allow registering custom editors specific for some component that have higher priority than the ones registered near the root of the application.
  - `useEditor` - hook to get the editor that should be used to edit the supplied value. First it looks for applicable editor in EditorsRegistry and if none was found it fallbacks to the default editors.

- `EditorRenderer` - wrapper around EditorRegistry that provides a convenient way to render editors for specific value.
- `useCommittableValue` - custom React hooks that provides a convenient way to add `cancel/commit` actions on `Esc`/`Enter` key click to the editor.
- `Value` - type for all values that are supported by editors.
- `ValueMetadata` - type for additional metadata that can be supplied to editors alongside value itself. It can be extended when implementing custom editors. (E.g. passing available choices and icons to the enum editor that is rendered as button group)
- `createEditorSpec` - an utility function to that provides a convenient way to defined editor spec for typed editors.
- `PropertyRecordEditor` - React component that allows to use existing `PropertyRecord` type with the new editors system.

## Usage

Some examples of the new editors system usage.

### Rendering editor and registering custom editors

Defining component that renders value editor:

```tsx
function EditingComponent(props: { value: Value; metadata: ValueMetadata }) {
  const onChange = (newValue: Value) => {
    // handle value change
  };
  return (
    <EditorRenderer
      value={value}
      metadata={metadata}
      onChange={onChange}
      size="small"
    />
  );
}
```

Defining component that renders value editor with `Commit`/`Cancel` actions:

```tsx
function EditingComponent(props: {
  initialValue: Value;
  metadata: ValueMetadata;
  onCommit: (committedValue: Value) => void;
  onCancel: () => void;
}) {
  // `onKeydown` callback returned by `useCommittableValue` will invoke `onCommit` with current value when `ENTER` is pressed or
  // `onCancel` when `ESC` key is pressed.
  // Additionally `commit` or `cancel` callback can be invoked to do commit or cancel manually.
  const { value, onChange, onKeydown, commit, cancel } = useCommittableValue({
    initialValue,
    onCommit,
    onCancel,
  });

  return (
    // cancel edit when editor is blurred
    <span onKeydown={onKeydown} onBlur={cancel}>
      <EditorRenderer
        value={value}
        metadata={metadata}
        onChange={onChange}
        commit={commit}
        cancel={cancel}
      />
    </span>
  );
}
```

Registering custom editors to be available through out all application:

```tsx
import {
  CustomBooleanEditorSpec,
  CustomNumericEditorSpec,
} from "./customEditors";

const rootEditors: EditorSpec[] = [
  CustomBooleanEditorSpec,
  CustomNumericEditorSpec,
];

function App() {
  return (
    <EditorsRegistryProvider editors={rootEditors}>
      {/* Render whole application components tree. Components down the tree will be able to use custom editors */}
    </EditorsRegistryProvider>
  );
}
```

Registering custom editors that should be available only for specific component:

```tsx
// setup custom editors for whole application
import {
  CustomBooleanEditorSpec,
  CustomNumericEditorSpec,
} from "./customEditors";

const rootEditors: EditorSpec[] = [
  CustomBooleanEditorSpec,
  CustomNumericEditorSpec,
];

function App() {
  return (
    <EditorsRegistryProvider editors={rootEditors}>
      <EditingComponent />
    </EditorsRegistryProvider>
  );
}
```

```tsx
// setup custom editors for specific component
import { SpecialNumericEditorSpec } from "./specialEditors";

const customEditors: EditorSpec[] = [SpecialNumericEditorSpec];

function EditingComponent(props: EditingComponentProps) {
  return (
    <EditorsRegistryProvider editors={customEditors}>
      {/* SpecialNumericEditorSpec has higher priority than CustomBooleanEditorSpec and CustomNumericEditorSpec registered at the root of application */}
      <EditorRenderer {...props} />
    </EditorsRegistryProvider>
  );
}
```

### Defining custom editors

The goal of the new editors system is to remove the need for static editor registration and provide more convenient API for implementing custom editors. Current API has quite a lot optional properties that do not make sense (`propertyRecord` is optional but if it is `undefined` there is no way to figure out what to render):

#### Custom editor using old editor system and react class components:

```tsx
interface CustomBooleanEditorState {
  currentValue: boolean;
}

class CustomBooleanEditor
  extends React.PureComponent<PropertyEditorProps, CustomBooleanEditorState>
  implements TypeEditor
{
  private _inputElement = React.createRef<HTMLInputElement>();
  public override readonly state: Readonly<CustomBooleanEditorState> = {
    currentValue: false,
  };

  public async getPropertyValue(): Promise<PropertyValue | undefined> {
    // this is an optional prop for some reason.
    const record = this.props.propertyRecord;
    let propertyValue: PropertyValue | undefined;

    if (record && record.value.valueFormat === PropertyValueFormat.Primitive) {
      propertyValue = {
        valueFormat: PropertyValueFormat.Primitive,
        value: this.state.currentValue,
        displayValue: "",
      };
    }

    return propertyValue;
  }

  public get htmlElement(): HTMLElement | null {
    return this._inputElement.current;
  }

  public get hasFocus(): boolean {
    return document.activeElement === this._inputElement.current;
  }

  public override componentDidUpdate() {
    const { propertyRecord } = this.props;
    if (
      propertyRecord &&
      propertyRecord.value.valueFormat === PropertyValueFormat.Primitive
    ) {
      this.setState({ currentValue: propertyRecord.value.value as boolean });
    }
    this.setState({ currentValue: false });
  }

  public override render() {
    return (
      <ToggleSwitch
        ref={this._inputElement}
        checked={this.state.currentValue}
        onChange={(e) => {
          const newValue = e.currentTarget.checked;
          this.setState({ currentValue: newValue }, () => {
            if (!this.props.propertyRecord || !this.props.onCommit) return;
            this.props.onCommit({
              propertyRecord: this.props.propertyRecord,
              newValue: {
                valueFormat: PropertyValueFormat.Primitive,
                value: newValue,
                displayValue: "",
              },
            });
          });
        }}
      />
    );
  }
}

class CustomBooleanPropertyEditor extends PropertyEditorBase {
  public get reactNode(): React.ReactNode {
    return <CustomBooleanEditor />;
  }
}
```

#### Custom editor using old system and react functional components:

```tsx
const CustomBooleanEditor = React.forwardRef<TypeEditor, PropertyEditorProps>(
  (props, ref) => {
    const inputRef = React.useRef<HTMLInputElement>(null);
    const getCurrentValue = () => {
      if (
        props.propertyRecord &&
        props.propertyRecord.value.valueFormat === PropertyValueFormat.Primitive
      ) {
        return props.propertyRecord.value.value as boolean;
      }
      return false;
    };
    const currentValue = getCurrentValue();

    React.useImperativeHandle(
      ref,
      () => ({
        getPropertyValue: async () => {
          let propertyValue: PropertyValue | undefined;
          if (
            props.propertyRecord &&
            props.propertyRecord.value.valueFormat ===
              PropertyValueFormat.Primitive
          ) {
            propertyValue = {
              valueFormat: PropertyValueFormat.Primitive,
              value: currentValue,
              displayValue: "",
            };
          }
          return propertyValue;
        },
        htmlElement: inputRef.current,
        hasFocus: document.activeElement === inputRef.current,
      }),
      [currentValue, props.propertyRecord]
    );

    return (
      <ToggleSwitch
        ref={inputRef}
        checked={currentValue}
        onChange={(e) => {
          if (!props.propertyRecord || !props.onCommit) return;
          props.onCommit({
            propertyRecord: props.propertyRecord,
            newValue: {
              valueFormat: PropertyValueFormat.Primitive,
              value: e.target.checked,
              displayValue: "",
            },
          });
        }}
      />
    );
  }
);

export class CustomBooleanPropertyEditor extends PropertyEditorBase {
  public get reactNode(): React.ReactNode {
    return <CustomBooleanEditor />;
  }
}
```

#### Custom boolean editor using new system:

```tsx
export const CustomBoolEditorSpec: EditorSpec = createEditorSpec({
  isMetadataSupported: (metadata): metadata is ValueMetadata =>
    metadata.type === "bool",
  isValueSupported: Value.isBoolean,
  Editor: CustomBooleanEditor,
});

export function CustomBooleanEditor(
  props: EditorProps<ValueMetadata, BooleanValue>
) {
  const currentValue = props.value?.value ?? false;
  const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
    const newValue = { value: e.target.checked };
    props.onChange(newValue);
    props.onFinish();
  };

  return <ToggleSwitch checked={currentValue} onChange={handleChange} />;
}
```

The new system removes all the code that was associated with class components and accessing values through editor `ref`. It is not clear if that was used/useful so the chosen approach is to add something similar later if that is still needed. Majority of that was used by `EditorContainer` that is replaced by `useCommittableValue` hook in the new system.

## Units supports

New editors system was aimed to provide better support for units. There is base component that should help with that `FormattedNumericInput`. It should be easy to write an editor on top of it that would know how to find `Parser`/`Formatter` for specific unit.

## TODO

- [x] Rewrite existing editors that support `PropertyEditorParams` from `PropertyRecord`. Need to find a way how to sunset those `PropertyEditorParams` in the future but in mean time if should be possible to maintain what is already there in the old system.
- [x] Need more work on webfont icons referenced by Tools in `PropertyEditorParams`. The initial approach is to maintain internal registry (iconName: string) => ReactNode that would hold currently used icons. Open for suggestions on this one.
- [x] Investigate more if current approach can be easily integrated with unit format overrides.
- [x] Do we need `CommitingEditor` as general solutions for committing entered values only on Enter/Blur or each components should have it's own logic to handle such workflows?
- [x] Add visual tests.
- [x] Add unit tests.
- [x] Deprecate old editors. Created separate issue: https://github.com/iTwin/appui/issues/1224
- [x] Remove Presentation from test-app (used for debugging and testing editors feature parity)


ui/components-react/src/components-react/newEditors/editorsRegistry/EditorsRegistryContext.ts
```diff
@@ -0,0 +1,19 @@
...
+  editors: EditorSpec[];
+}
+
+/**
+ * Context for storing registered editors.
+ * @internal
+ */
+export const editorsRegistryContext = React.createContext<EditorsRegistry>({
+  editors: [],
+});
```
GerardasB, line 19:
```suggestion
export const EditorsRegistryContext = React.createContext<EditorsRegistry>({
  editors: [],
});
EditorsRegistryContext.displayName = "uifw:EditorsRegistryContext";
```

ui/components-react/src/components-react/newEditors/CommittingEditor.tsx
```diff
@@ -0,0 +1,67 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
+import { useRef, useState } from "react";
+import { Editor } from "./Editor.js";
+import type { Value } from "./values/Values.js";
+import type { EditorProps } from "./Types.js";
+
+type CommittingEditorProps = Omit<EditorProps, "onFinish" | "onChange"> & {
```
GerardasB, line 11: This means, that `CommitingEditor` is not a "real editor" and can not be registered via `EditorsRegistryProvider`?

ui/components-react/src/components-react/newEditors/Types.ts
```diff
@@ -0,0 +1,38 @@
...
+  onFinish: () => void;
+  disabled?: boolean;
+  size?: "small" | "large";
+}
+
+/**
+ * A type that makes a specific properties required in a type.
+ * @beta
+ */
+export type RequiredProps<TProps, TKey extends keyof TProps> = Omit<
```
GerardasB, line 34:
```suggestion
/**
 * A type that makes a specific properties required in a type.
 * @internal
 */
export type RequiredProps<TProps, TKey extends keyof TProps> = Omit<
```

GerardasB (commented):
Looks good for the most part, however I'd really like to reduce the API surface to a minimum before merging this in (instead of thinking about all the what-if use-cases).
One more major thing to address is the addition of `core-quantity` peer dep.

ui/components-react/src/components-react/newEditors/CommittingEditor.tsx
GerardasB: Nvm, I see this is only adding some keyboard handlers for the regulard `Editor`.

ui/components-react/src/components-react/newEditors/Types.ts
```diff
@@ -0,0 +1,38 @@
...
+ * Generic editor props that are supplied to the editor for rendering.
+ * @beta
+ */
+export interface EditorProps<TValue = Value> {
+  metadata: ValueMetadata;
+  value?: TValue;
+  onChange: (value: TValue) => void;
+  onFinish: () => void;
+  disabled?: boolean;
+  size?: "small" | "large";
```
GerardasB, line 27: `size` should probably not be controllable per editor?

ui/components-react/src/components-react/newEditors/values/Values.ts
```diff
@@ -0,0 +1,126 @@
...
+  displayValue: string;
+  roundingError?: number;
+}
+
+/**
+ * Type definition for instance key value that can be handled by editor.
+ * @beta
+ */
+export interface InstanceKeyValue {
+  key: { id: string; className: string };
```
grigasp, line 21: I think using `Id64String` over `string` makes it more clear what we expect there.

```suggestion
  key: { id: Id64String; className: string };
```

ui/components-react/src/components-react/newEditors/values/Values.ts
```diff
@@ -0,0 +1,126 @@
...
+  | TextValue
+  | BooleanValue
+  | DateValue
+  | EnumValue;
+
+/**
+ * Type guard for text value.
+ * @beta
+ */
+export function isTextValue(
```
grigasp, line 73: maybe it makes sense to put these type guards under `Value` namespace?

ui/components-react/src/components-react/newEditors/values/Metadata.ts
```diff
@@ -0,0 +1,41 @@
...
+export type ValueType =
+  | "string"
+  | "number"
+  | "bool"
+  | "date"
+  | "enum"
+  | "instanceKey";
+
+/**
+ * Additional metadata that is used along side value to determine applicable editor.
```
grigasp, line 19: do we want to mention "editor" in docs of these types? I could see them being used in non-editing contexts, e.g. read-only property rendering use cases

ui/components-react/src/components-react/newEditors/values/Values.ts
```diff
@@ -0,0 +1,126 @@
...
+  value: Date;
+}
+
+/**
+ * Type definition for enum value that can be handled by editor.
+ * @beta
+ */
+export interface EnumValue {
+  choice: number | string;
+  label: string;
```
grigasp, line 55: does enum value need to have a label? I think it could be retrieved from metadata based on `choice`. Also, maybe `choice` should be renamed to `value`?

ui/components-react/src/components-react/newEditors/values/Metadata.ts
```diff
@@ -0,0 +1,41 @@
...
+ */
+export interface EnumChoice {
+  value: number | string;
+  label: string;
+}
+
+/**
+ * Additional metadata that is used along side enum value to determine applicable editor.
+ */
+export interface EnumValueMetadata extends ValueMetadata {
```
grigasp, line 38: I think this type should also have `isStrict: boolean`. See `isStrict` here: https://www.itwinjs.org/reference/ecschema-metadata/metadata/enumeration/

ui/components-react/src/components-react/newEditors/editors/boolean/UseBooleanEditorProps.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { EditorProps, RequiredProps } from "../../Types.js";
+import type { BooleanValue, Value } from "../../values/Values.js";
+import { isBooleanValue } from "../../values/Values.js";
+
+/**
+ * Hooks that converts generic `EditorProps` into editor props with boolean value. If value is not boolean, it will be converted into `false`.
```
grigasp, line 11: > If value is not boolean, it will be converted into `false`.

Maybe we should use truthiness instead? E.g. `0` would convert to `false`, but any other number would convert to `true`.

ui/components-react/src/components-react/newEditors/editors/boolean/BooleanEditor.tsx
```diff
@@ -0,0 +1,32 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
+import { ToggleSwitch } from "@itwin/itwinui-react";
+import type { EditorProps } from "../../Types.js";
+import { useBooleanEditorProps } from "./UseBooleanEditorProps.js";
+
+/**
+ * Simple editor for editing boolean values.
```
grigasp, line 11: Maybe instead of "simple" it should say what underlying component it uses to render the editor (toggle switch)? I think that may be important to whoever uses it, as some consumers may want something different, e.g. a checkbox.

ui/components-react/src/components-react/newEditors/editors/enum/UseEnumEditorProps.ts
```diff
@@ -0,0 +1,44 @@
...
+  choices: EnumChoice[];
+} {
+  const choices =
+    metadata.type === "enum" ? (metadata as EnumValueMetadata).choices : [];
+
+  return {
+    ...rest,
+    metadata,
+    value: getEnumValue(value, choices),
+    onChange,
```
grigasp, line 30: this could be just left in `rest`

ui/components-react/src/components-react/newEditors/values/Metadata.ts
```diff
@@ -0,0 +1,41 @@
...
+  | "bool"
+  | "date"
+  | "enum"
+  | "instanceKey";
+
+/**
+ * Additional metadata that is used along side value to determine applicable editor.
+ * @beta
+ */
+export interface ValueMetadata {
```
grigasp, line 22: maybe this should also have information if the value is nullable? in our case most properties will be nullable, but some won't (e.g. some navigation properties like `BisCore:Element.Model`)

ui/components-react/src/components-react/newEditors/editors/enum/UseEnumEditorProps.ts
```diff
@@ -0,0 +1,44 @@
...
+}
+
+function getEnumValue(
+  value: Value | undefined,
+  choices: EnumChoice[]
+): EnumValue {
+  const defaultValue =
+    choices.length > 0
+      ? { choice: choices[0].value, label: choices[0].label }
+      : { choice: 0, label: "" };
```
grigasp, line 42: maybe in this case the `choice` should be empty string instead of `0`?

ui/components-react/src/components-react/newEditors/editors/text/UseTextEditorProps.ts
```diff
@@ -0,0 +1,28 @@
...
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+
+import type { EditorProps, RequiredProps } from "../../Types.js";
+import type { TextValue, Value } from "../../values/Values.js";
+import { isTextValue } from "../../values/Values.js";
+
+/**
+ * Hooks that converts generic `EditorProps` into editor props with text value. If value is not text, it will be converted into empty text value.
```
grigasp, line 11: > If value is not text, it will be converted into empty text value.

Maybe we should attempt to convert the value to its text representation instead?

ui/components-react/src/components-react/newEditors/editors/numeric/FormattedNumericInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+  }
+
+  const [state, setState] = React.useState<State>(() => {
+    return {
+      value: {
+        rawValue: initialValue,
+        displayValue:
+          initialValue !== undefined ? formatValue(initialValue) : "",
+      },
+      placeholder: formatValue(123.45),
```
grigasp, line 76: maybe use `initialValue` as the placeholder, if it's provided?

ui/imodel-components-react/src/imodel-components-react/editors/NewColorEditor.tsx
```diff
@@ -0,0 +1,105 @@
...
+    (metadata as ColorValueMetadata).params !== undefined,
+  Editor: ColorEditor,
+};
+
+/**
+ * Metadata for color editor.
+ * @beta
+ */
+export interface ColorValueMetadata extends ValueMetadata {
+  params: ColorEditorParams[];
```
grigasp, line 44: why is this an array?

grigasp (commented):
Reviewed most of the stuff in `components-react` and `imodel-components-react`.

apps/test-app/src/routes/iTwin_.$iTwinId.tsx
```diff
@@ -2,42 +2,42 @@
...
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
-import React from "react";
-import { createFileRoute, useNavigate } from "@tanstack/react-router";
-import { IModelGrid } from "@itwin/imodel-browser-react";
-import { config } from "../frontend/config";
-import { useAuth } from "../frontend/Auth";
-import { PageLayout } from "@itwin/itwinui-layouts-react";
-import { SignInPage } from "../frontend/SignInPage";
+import React from 'react'
```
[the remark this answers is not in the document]
saskliutas: Reverted

apps/test-app/src/routes/local_.$fileName.tsx
```diff
@@ -2,55 +2,55 @@
...
-import React from "react";
-import { createFileRoute } from "@tanstack/react-router";
-import { PageLayout } from "@itwin/itwinui-layouts-react";
-import { config } from "../frontend/config";
-import { SnapshotConnection } from "@itwin/core-frontend";
-import { createViewState } from "../frontend/createViewState";
-import { appInitializer } from "../frontend/AppInitializer";
-import { App } from "../frontend/App";
-import { UiFramework } from "@itwin/appui-react";
+import React from 'react'
```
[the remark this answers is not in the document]
saskliutas: Reverted

apps/test-providers/src/tools/SampleTool.ts
```diff
@@ -628,20 +630,36 @@ export class SampleTool extends PrimitiveTool {
...
+      property: {
+        ...this._surveyLengthDescription,
+        quantityType: getQuantityTypeKey(
+          this._surveyLengthDescription.formatterQuantityType
+        ),
+      },
       editorPosition: { rowPriority: 21, columnIndex: 2 },
       isDisabled: readonly,
     });
+    const angleDescription = new AngleDescription();
```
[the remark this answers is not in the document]
saskliutas: Reverted. It was used for testing.

apps/test-app/src/frontend/appui/frontstages/EditorFrontstage.tsx
```diff
@@ -53,24 +49,6 @@ export function createEditorFrontstageProvider(): UiItemsProvider {
         icon: <SvgEdit />,
       }),
     ],
-    getToolbarItems: () => {
```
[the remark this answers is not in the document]
saskliutas: `CreateArcTool`, `CreateLineStringTool` are not exported from  "@itwin/editor-frontend" after version bump

ui/components-react/package.json
```diff
@@ -57,6 +57,7 @@
     "@itwin/appui-abstract": "^4.0.0",
     "@itwin/core-bentley": "^4.0.0",
     "@itwin/core-react": "workspace:^5.1.0-dev.0",
+    "@itwin/core-quantity": "^4.0.0",
```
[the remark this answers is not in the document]
saskliutas: Removed. It was a leftover from initial attempt.

ui/components-react/src/components-react/newEditors/interop/EditorInterop.ts
```diff
@@ -0,0 +1,173 @@
...
+import {
+  isBooleanValue,
+  isDateValue,
+  isEnumValue,
+  isNumericValue,
+  isTextValue,
+} from "../values/Values.js";
+import type { OldEditorMetadata } from "./Metadata.js";
+
+export namespace EditorInterop {
```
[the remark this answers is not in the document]
saskliutas: Made internal

ui/components-react/src/components-react.ts
```diff
@@ -92,6 +92,8 @@ export {
 } from "./components-react/datepicker/DatePickerPopupButton.js";
 export { IntlFormatter } from "./components-react/datepicker/IntlFormatter.js";
 
+export { EditorInterop } from "./components-react/newEditors/interop/EditorInterop.js";
```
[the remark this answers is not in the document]
saskliutas: Removed

ui/components-react/src/components-react.ts
```diff
@@ -92,6 +92,8 @@ export {
...
+export * from "./components-react/newEditors/index.js";
```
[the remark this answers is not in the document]
saskliutas: Cleaned up exports. Removed default editor implementations and accompanying code from barrel files.

ui/components-react/src/components-react/newEditors/values/Values.ts
saskliutas: Updated type.

ui/components-react/src/components-react/newEditors/values/Values.ts
saskliutas: Added namespace

ui/components-react/src/components-react/newEditors/values/Metadata.ts
saskliutas: Removed `editor` from comment

ui/components-react/src/components-react/newEditors/values/Metadata.ts
saskliutas: Added `isStrict`.

ui/components-react/src/components-react/newEditors/editors/boolean/UseBooleanEditorProps.ts
saskliutas: Removed conversion itself.

ui/components-react/src/components-react/newEditors/editors/enum/UseEnumEditorProps.ts
saskliutas: Removed this hook.

ui/components-react/src/components-react/newEditors/editors/boolean/BooleanEditor.tsx
saskliutas: Updated comment on default editors.

ui/components-react/src/components-react/newEditors/values/Metadata.ts
saskliutas: Added `isNullable`

ui/components-react/src/components-react/newEditors/editors/numeric/FormattedNumericInput.tsx
saskliutas: Switched to `initialValue`

ui/components-react/src/components-react/newEditors/editors/text/UseTextEditorProps.ts
saskliutas: Removed conversion.

ui/components-react/src/components-react/newEditors/editors/enum/UseEnumEditorProps.ts
saskliutas: Switched default to empty string.

ui/imodel-components-react/src/imodel-components-react/editors/NewColorEditor.tsx
saskliutas: Switched to `params: PropertyEditorParams[];`. This is what is passed to old editors through `PropertyRecord`

ui/components-react/src/components-react/newEditors/values/Values.ts
grigasp: Do you think the "Value" suffix in method names are needed? IMO, `Value.isDate` > `Value.isDateValue`

ui/components-react/src/components-react/newEditors/values/Values.ts
grigasp: Do we have cases where we render a property value without having its metadata? I think `InstanceKeyValue` is different, because getting its display value requires querying an iModel - looking at the metadata is not enough.

grigasp (commented):
I think the `CustomBooleanEditor` example in the PR description should be updated and should include `BoolEditorSpec` (the old system examples have value validation, which is part of the spec now).

ui/components-react/src/components-react/newEditors/Types.ts
```diff
@@ -0,0 +1,56 @@
...
+}: {
+  isMetadataSupported: (metadata: ValueMetadata) => metadata is TMetadata;
+  isValueSupported: (value: Value) => value is TValue;
+  applies?: (metadata: TMetadata, value?: TValue) => boolean;
+  Editor: React.ComponentType<EditorProps<TMetadata, TValue>>;
+}): EditorSpec {
+  return {
+    applies: (metadata: ValueMetadata, value?: Value) =>
+      isMetadataSupported(metadata) &&
+      (value === undefined || isValueSupported(value)) &&
```
grigasp, line 51: `applies` will return `true` when `metadata.isNullable === false` and `value === undefined`. Is that intended?

ui/components-react/src/components-react/newEditors/CommittingEditor.tsx
```diff
@@ -0,0 +1,67 @@
...
+      handleCommit();
+    } else if (e.key === "Escape") {
+      onCancel?.();
+    }
+  };
+
+  return (
+    // eslint-disable-next-line jsx-a11y/no-static-element-interactions
+    <div onKeyDown={handleKeyDown}>
+      <Editor
```
grigasp, line 57: Is `Editor` component expected to ever be used outside of here? Maybe it should be moved into this source file and not exported?

ui/components-react/src/components-react/newEditors/editors/NumericEditor.tsx
```diff
@@ -0,0 +1,41 @@
...
+}: EditorProps<ValueMetadata, NumericValue>) {
+  const currentValue = getNumericValue(value);
+  return (
+    <Input
+      value={currentValue.displayValue}
+      onChange={(e) =>
+        onChange({
+          rawValue: parseFloat(e.target.value),
+          displayValue: e.target.value,
+        })
```
grigasp, line 30: Should this editor have some default rounding? Maybe rounding precision could be specified through metadata?

ui/components-react/src/components-react/newEditors/editorsRegistry/DefaultEditors.tsx
```diff
@@ -0,0 +1,97 @@
...
+    metadata.type === "enum",
+  isValueSupported: Value.isEnumValue,
+  Editor: EnumEditor,
+});
+
+/**
+ * List of default editors that are used as fallback if EditorRegistry does not provide a custom editor.
+ * @internal
+ */
+export const defaultEditors: EditorSpec[] = [
```
grigasp, line 77: I suggest renaming this to `defaultEditorSpecs` to avoid confunsion between editors and editor specs. Same with `interopEditors`.

ui/components-react/src/components-react/newEditors/editorsRegistry/UseEditor.ts
```diff
@@ -0,0 +1,47 @@
...
+import * as React from "react";
+import type { EditorSpec } from "../Types.js";
+import type { ValueMetadata } from "../values/Metadata.js";
+import type { Value } from "../values/Values.js";
+import { defaultEditors, interopEditors } from "./DefaultEditors.js";
+import { EditorsRegistryContext } from "./EditorsRegistryContext.js";
+import { FallbackEditor } from "../editors/FallbackEditor.js";
+
+/**
+ * Custom hooks that returns editor for specified metadata and value. It uses `EditorsRegistry` context to find registered editors.
```
grigasp, line 14:
```suggestion
 * Custom React hook that returns editor for specified metadata and value. It uses `EditorsRegistry` context to find registered editors.
```

ui/components-react/src/components-react/newEditors/interop/old-editors/enum/Enum.tsx
```diff
@@ -0,0 +1,29 @@
...
+import { EnumEditor as NewEnumEditor } from "../../../editors/EnumEditor.js";
+import type { OldEditorMetadata } from "../../Metadata.js";
+import { isOldEditorMetadata } from "../../Metadata.js";
+import type { EnumValue } from "../../../values/Values.js";
+import { Value } from "../../../values/Values.js";
+
+export const EnumEditorSpec: EditorSpec = createEditorSpec({
+  isMetadataSupported: isOldEditorMetadata,
+  isValueSupported: Value.isEnumValue,
+  applies: (metadata) => metadata.type === "enum",
```
grigasp, line 22: why is this check not in `isMetadataSupported`?

ui/components-react/src/components-react/newEditors/editors/EnumEditor.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+  const handleChange = (newChoice: number | string) => {
+    const choice = choices.find((c) => c.value === newChoice);
+    const newValue = { choice: newChoice, label: choice?.label ?? "" };
+    onChange(newValue);
+    onFinish();
+  };
+
+  return (
+    <Select
```
grigasp, line 35: Does this component allow entering free-text values? If the enum is not strict, we should accept anything user enters.

ui/components-react/src/components-react/newEditors/interop/EditorInterop.ts
```diff
@@ -0,0 +1,170 @@
...
+          value: {
+            value: (primitiveValue.value as string) ?? "",
+          } satisfies TextValue,
+        };
+      case "dateTime":
+      case "shortdate":
+        return {
+          metadata: {
+            ...baseMetadata,
+            type: "date",
```
grigasp, line 60: I believe the difference between `dateTime` and `shortDate` in the old system was that the former also accepted time, while the latter only accepted date. Do we have that distinction in the new system?

ui/components-react/src/components-react/newEditors/interop/EditorInterop.ts
```diff
@@ -0,0 +1,170 @@
...
+          metadata: {
+            ...baseMetadata,
+            type: "number",
+          },
+          value: {
+            rawValue: primitiveValue.value as number,
+            displayValue:
+              primitiveValue.value !== undefined
+                ? `${primitiveValue.value as number}`
+                : primitiveValue.displayValue ?? "",
```
grigasp, line 92: shouldn't we prefer `displayValue` over `value`?

ui/components-react/src/components-react/newEditors/CommittingEditor.tsx
```diff
@@ -0,0 +1,67 @@
...
+    setCurrentValue(newValue);
+  };
+
+  const handleCommit = () => {
+    if (currentValueRef.current !== undefined) {
+      onCommit(currentValueRef.current);
+      return;
+    }
+
+    onCancel?.();
```
grigasp, line 41: shouldn't we allow to commit an `undefined` value?

ui/components-react/src/components-react/newEditors/Types.ts
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Generic editor props that are supplied to the editor for rendering.
+ * @beta
+ */
+export interface EditorProps<TValue = Value> {
+  metadata: ValueMetadata;
+  value?: TValue;
+  onChange: (value: TValue) => void;
+  onFinish: () => void;
```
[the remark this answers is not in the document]
grigasp: Just a thought... IMO, it would make sense to rename `onFinish` to `commit`. For consistency, I'd also rename `onChange` to `changeValue`. What do you think?

apps/test-providers/src/ui/ViewportContent.tsx
```diff
@@ -6,9 +6,12 @@ import * as React from "react";
 import { DefaultViewOverlay, UiFramework } from "@itwin/appui-react";
 import { IModelApp, ScreenViewport, ViewState } from "@itwin/core-frontend";
 import { ViewportComponent } from "@itwin/imodel-components-react";
+import { viewWithUnifiedSelection } from "@itwin/presentation-components";
 
 type ViewportComponentProps = React.ComponentProps<typeof ViewportComponent>;
 
+const UnifiedViewport = viewWithUnifiedSelection(ViewportComponent);
```
grigasp, line 13: I'd rather we use the `@itwin/unified-selection` package than the deprecated `viewWithUnifiedSelection`

saskliutas: This is just for local debugging and testing. All presentation related APIs will be removed from appui test-app.

ui/components-react/src/components-react/newEditors/CommittingEditor.tsx
saskliutas: Yes. I expect to use it in `FilterBuilder` component where we treat editor as a controlled input and do not wait for committing action.

ui/components-react/src/components-react/propertygrid/component/VirtualizedPropertyGridWithDataProvider.tsx
```diff
@@ -38,6 +38,7 @@ export interface VirtualizedPropertyGridWithDataProviderProps
   width: number;
   /** Height of the property grid component. */
   height: number;
+  usedEditor?: "old" | "new";
```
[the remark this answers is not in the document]
grigasp: I'd also consider renaming "old" to "legacy".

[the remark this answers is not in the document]
grigasp: I'd also consider renaming "old" to "legacy".

ui/components-react/src/components-react/newEditors/editors/EnumEditor.tsx
saskliutas: No. This works the same as an old editor. I think it would be better to add support for free-text values in a separate PR if we need it.

ui/components-react/src/components-react/newEditors/Types.ts
saskliutas: Yes, this is intended for use cases where editor is rendered without initial value.

ui/components-react/src/components-react/newEditors/values/Values.ts
saskliutas: I don't think we have such cases. I guess I chose to add label to the value to simplify value rendering when it's edited based on `PropertyRecord` that has `choices` as `Promise` https://github.com/iTwin/itwinjs-core/blob/1bc47d38777b51c4a3146c10207629defb38ee08/ui/appui-abstract/src/appui-abstract/properties/Description.ts#L32

However, I guess it will have to be awaited when creating `ValueMetadata` so it should not cause problems. Other use case I am now thinking is entering free form text when enum is not strict. Maybe in that case it would be useful to have label on value? For example if enum value is based on number and allows to enter label for that new choice.

ui/components-react/src/components-react/newEditors/Types.ts
grigasp: It means the editor is responsible for making it clear to users that it requires a value. Do we have a plan for how editors will expose validation errors?

ui/components-react/src/components-react/newEditors/values/Values.ts
grigasp: For non-strict enums, label always matches raw value. Of course, label is always a string, and value could be a number, but they easily convert to/from each other.

ui/components-react/src/components-react/newEditors/Types.ts
saskliutas: Created a new issue for investigating and implementing the ability to show validation errors. https://github.com/iTwin/appui/issues/1219

ui/components-react/src/components-react/new-editors/Types.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @beta
+ */
+export function createEditorSpec<
+  TMetadata extends ValueMetadata,
+  TValue extends Value
+>({
+  Editor,
+  isMetadataSupported,
+  isValueSupported,
+  applies,
```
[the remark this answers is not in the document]
saskliutas: My idea was that `isMetadataSupported` and `isValueSupported` works only as a typeguards, simple check that supplied object are what editor expects to receive. Additional logic that is needed to determine if editor is applicable for current property would be in `applies` function. I thought this would be useful if multiple editors works with the same type of `ValueMetadata` so they could use the same typeguard and make additional check in apply.

However, after implementing existing editors using new system I feel that `applies` function might not be needed.

ui/components-react/src/components-react/propertygrid/internal/flat-properties/FlatPropertyRenderer.tsx
```diff
@@ -208,3 +239,34 @@ function useResetHeightOnEdit(
...
+  onCancel?: () => void;
+}) {
+  const { value, onChange, commit, cancel, onKeydown } = useCommittableValue({
+    initialValue,
+    onCancel,
+    onCommit,
+  });
+
+  return (
+    <div onKeyDown={onKeydown} onBlur={commit} role="presentation">
```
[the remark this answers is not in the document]
saskliutas: Updated to use the same `PropertyRecordEditor` in both places.

ui/components-react/src/components-react/new-editors/UseCommittableValue.tsx
```diff
@@ -0,0 +1,101 @@
...
+
+  const handleCommit = () => {
+    if (currentValueRef.current.state === "changed") {
+      onCommit(currentValueRef.current.value);
+      return;
+    }
+    if (currentValueRef.current.state === "cancelled") {
+      return;
+    }
+    onCancel?.();
```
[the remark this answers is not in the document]
saskliutas: Added additional state `initial` so the `onCancel` would be called when trying to commit unchanged value. 

I feel this might be improved a little bit to handle cases when value is changed and later reverted back to initial value: https://github.com/iTwin/appui/issues/1220

ui/components-react/src/components-react/newEditors/interop/EditorInterop.ts
saskliutas: Switched to `value`

ui/components-react/src/components-react/propertygrid/component/VirtualizedPropertyGrid.tsx
```diff
@@ -69,6 +69,7 @@ export interface VirtualizedPropertyGridProps extends CommonPropertyGridProps {
   width: number;
   /** Height of the property grid component. */
   height: number;
+  usedEditor?: "old" | "new";
```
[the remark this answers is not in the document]
saskliutas: Added comment

ui/components-react/src/components-react/propertygrid/component/VirtualizedPropertyGridWithDataProvider.tsx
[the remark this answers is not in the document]
saskliutas: Added comment and renamed to `legacy`

ui/components-react/src/components-react/new-editors/Types.ts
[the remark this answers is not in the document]
grigasp: I like it - for me it's much easier now to understand what check goes where.

ui/components-react/src/components-react/newEditors/editors/NumericEditor.tsx
grigasp: NVM, I guess formatting in the `onChange` callback doesn't make much sense.

ui/components-react/src/components-react/propertygrid/component/VirtualizedPropertyGridWithDataProvider.tsx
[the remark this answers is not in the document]
grigasp: one last nitpick... what do you think about renaming `usedEditor` prop in all components to `editorSystem`?

ui/components-react/src/components-react/new-editors/interop/IconsRegistry.tsx
```diff
@@ -0,0 +1,71 @@
...
+    </svg>
```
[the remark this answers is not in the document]
saskliutas: In the test-app and e2e-tests

grigasp (approved):

ui/components-react/src/components-react/new-editors/editors/FormattedNumericInput.tsx
```diff
@@ -0,0 +1,120 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
```
JonasDov, line 5:
```suggestion
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
import * as React from "react";
```

ui/components-react/src/components-react/new-editors/editors/FormattedNumericInput.tsx
```diff
@@ -0,0 +1,120 @@
...
+  parseValue: (value: string) => number | undefined;
+  formatValue: (num: number) => string;
+  disabled?: boolean;
+  size?: "small" | "large";
+}
+
+/**
+ * A numeric input that allows to pass custom parsing/formatting logic to handle values with units etc.
+ * @beta
+ */
```
JonasDov, line 27: Why is `DateInput` marked as `@internal`, but `FormattedNumericInput` marked as `@beta`

ui/components-react/src/components-react/new-editors/editors/NumericEditor.tsx
```diff
@@ -0,0 +1,43 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
```
JonasDov, line 5:
```suggestion
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
import * as React from "react";
```

ui/components-react/src/components-react/new-editors/editors/TextEditor.tsx
```diff
@@ -0,0 +1,35 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
```
JonasDov, line 5:
```suggestion
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
import * as React from "react";
```

ui/components-react/src/components-react/new-editors/interop/EditorInterop.ts
```diff
@@ -0,0 +1,173 @@
...
+
+/**
+ * Interop utilities for converting between old and new editor values.
+ * @internal
+ */
+export namespace EditorInterop {
+  /**
+   * @internal
+   */
+  export function getMetadataAndValue(propertyRecord: PropertyRecord): {
```
JonasDov, line 31: Do we need to mark this function as `@internal` if namespace is marked as `@internal`?

ui/components-react/src/components-react/new-editors/interop/EditorInterop.ts
```diff
@@ -0,0 +1,173 @@
...
+    return {
+      metadata: undefined,
+      value: undefined,
+    };
+  }
+
+  /**
+   * @internal
+   */
+  export function convertToPrimitiveValue(
```
JonasDov, line 132: Do we need to mark this function as `@internal` if namespace is marked as `@internal`?

ui/imodel-components-react/src/imodel-components-react/editors/NewWeightEditor.tsx
```diff
@@ -0,0 +1,50 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
```
JonasDov, line 5:
```suggestion
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
import * as React from "react";
```

ui/imodel-components-react/src/imodel-components-react/inputs/newEditors/QuantityEditor.tsx
```diff
@@ -0,0 +1,66 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
```
JonasDov, line 5:
```suggestion
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
import * as React from "react";
```

ui/imodel-components-react/src/imodel-components-react/inputs/newEditors/QuantityInput.tsx
```diff
@@ -0,0 +1,66 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import * as React from "react";
```
JonasDov, line 5:
```suggestion
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
import * as React from "react";
```

ui/imodel-components-react/src/imodel-components-react/inputs/newEditors/UseQuantityInfo.ts
```diff
@@ -0,0 +1,71 @@
+/*---------------------------------------------------------------------------------------------
+ * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
+ * See LICENSE.md in the project root for license terms and full copyright notice.
+ *--------------------------------------------------------------------------------------------*/
+import type { QuantityTypeArg } from "@itwin/core-frontend";
```
JonasDov, line 5:
```suggestion
/*---------------------------------------------------------------------------------------------
 * Copyright (c) Bentley Systems, Incorporated. All rights reserved.
 * See LICENSE.md in the project root for license terms and full copyright notice.
 *--------------------------------------------------------------------------------------------*/
 
import type { QuantityTypeArg } from "@itwin/core-frontend";
```

ui/components-react/src/components-react/new-editors/editors-registry/DefaultEditors.tsx
```diff
@@ -0,0 +1,140 @@
...
+/**
+ * Specification for default date editor. It applies for values whose type is "date".
+ * @internal
+ */
+export const DateTimeEditorSpec: EditorSpec = createEditorSpec({
+  isMetadataSupported: (metadata): metadata is ValueMetadata =>
+    metadata.type === "dateTime",
+  isValueSupported: Value.isDate,
+  Editor: DateTimeEditor,
+});
```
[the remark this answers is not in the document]
saskliutas: Updated comment

ui/components-react/src/components-react/new-editors/editors/FormattedNumericInput.tsx
saskliutas: I expect `FormattedNumericInput` to be used when implementing custom editors using quantities or something similar. `DateInput` is just a simple component shared between `Date` and `DateTime` editors and I don't think it provides much value for consumers. Also it could be made public in a future if necessary.

grigasp (commented):
Should the legacy APIs be deprecated in this PR? Or are you planning to do that separately?

saskliutas:
> Should the legacy APIs be deprecated in this PR? Or are you planning to do that separately?

I want to do that in separate PR. Have an issue for that https://github.com/iTwin/appui/issues/1224.

common/api/components-react.api.md
```diff
@@ -3249,6 +3367,32 @@ export function useVirtualizedPropertyGridLayoutStorage<T extends Element>(): {
     restore: () => void;
 };
 
+// @beta
+export type Value = NumericValue | InstanceKeyValue | TextValue | BooleanValue | DateValue | EnumValue;
```
GerardasB, line 3371: Can we add new value types later if needed? I.e. for string literal type extensibility we do this: https://github.com/iTwin/appui/blob/89d6bf22b8110d74d01b55f6cbfc497559c8bcf1/ui/appui-react/src/appui-react/theme/ThemeId.ts#L38

common/api/components-react.api.md
```diff
@@ -3249,6 +3367,32 @@ export function useVirtualizedPropertyGridLayoutStorage<T extends Element>(): {
...
+    // (undocumented)
+    isNullable?: boolean;
+    // (undocumented)
+    preferredEditor?: string;
+    // (undocumented)
+    type: ValueType;
+}
+
+// @beta
+export type ValueType = "string" | "number" | "bool" | "date" | "dateTime" | "enum" | "instanceKey";
```
GerardasB, line 3394: Similar to `Value`, should we make this extensible?

common/api/components-react.api.md
```diff
@@ -3249,6 +3367,32 @@ export function useVirtualizedPropertyGridLayoutStorage<T extends Element>(): {
     restore: () => void;
 };
 
+// @beta
+export type Value = NumericValue | InstanceKeyValue | TextValue | BooleanValue | DateValue | EnumValue;
+
+// @beta (undocumented)
+export namespace Value {
```
GerardasB, line 3374: Should we avoid adding new namespaces? https://typescript-eslint.io/rules/no-namespace/
Add a new module `ValueUtilities.ts` instead and export it in the barrel file `export * as ValueUtilities from "ValueUtilities.js"`

GerardasB (approved):

calebmshafer:
@saskliutas I have a more meta question about this PR more than anything specific to the implementation.

Should this new set of editors live inside of the AppUI components package or be split out of AppUI into its own separate package? Is there anything technically that would prevent this today, such as existing dependencies in other components within the package?

I have the same question generally for a lot of the things in components react given the way we are not splitting major components into their own individual packages with their own release version and cadence.

cc: @GerardasB @aruniverse

saskliutas:
> @saskliutas I have a more meta question about this PR more than anything specific to the implementation.
> 
> Should this new set of editors live inside of the AppUI components package or be split out of AppUI into its own separate package? Is there anything technically that would prevent this today, such as existing dependencies in other components within the package?
> 
> I have the same question generally for a lot of the things in components react given the way we are not splitting major components into their own individual packages with their own release version and cadence.
> 
> cc: @GerardasB @aruniverse

I think there is nothing preventing from moving editors into a separate package. Only uncomfortable part is that I am using React context to allow supplying custom editors so that new editors package would have to be a peerDependency in other packages delivering components using editors. As majority of packages delivering widgets already has peerDependecy on `components-react` we decided that leaving editor in `components-react` would be the least disturbing change for consumers. This also means that appui-react would need to have a new peerDependency on the new editors package in order to use them for tool settings.

e2e-tests/tests/editors/editors.test.ts-snapshots/Editor-Primitive-boolean-image-check-box-UiAbstract-CheckBoxImages-default-visual-in-New-system-1-chromium-linux.png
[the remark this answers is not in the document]
saskliutas: I haven't implemented ImageCheckbox editor using new system. Now sure if it's used. As this new editors system is opt-in I think we can ship without it and add it if there someone requesting.

Also it gets Icons as `string` so in my opinion it would be better to introduce new ImageCheckboxEditor that works with icons as `ReactNode` instead of `string`.

common/api/components-react.api.md
saskliutas: I guess we could make it extensible. One thing I don't like about this approach is that there will be no errors for typos. E.g. 

```ts
function doSomething() {
  const stringValue: ValueMetadata = {
    type: "stirng", // no error for misspelled type
  };

  // no error for misspelled type
  if (stringValue.type === "strng") {
    // handle string
  }
}
```

The intelisense does suggest available string literals so I guess this is not the much of an issue? What do you think?

GerardasB: I like this approach over simply adding a comment that new types can be added. What about https://github.com/iTwin/appui/pull/1166#discussion_r1991033472?

common/api/components-react.api.md
GerardasB: Using the module aggregation method described above will reduce the clutter without using the namespace:

```tsx
// ValueUtilities.ts
export const x = 0;
export const y = 0;

// barrel.ts
export * as ValueUtilities from "ValueUtilities.js";

// Consumer usage
import { ValueUtilities } from "components-react";
ValueUtilities.x;
```

PR #1179, opened 2025-01-16
Fix popout widget issue when `useDefaultPopoutUrl` is not used

### LINKED ISSUE #1178: Popout does not close properly when useDefaultPopoutUrl = false
### Describe the bug

When trying to reproduce the issue with disappearing widget after closing (see #1170), we ended up discovering that it works in the test case because useDefaultPopUrl is set to `true`, while in our app it is always `false`. 

If we were to set it to `true`, then we could end up with undesirable behavior as it goes through the app's routing policies and may add other UI components to the popout page. For our test-app, it added the itwin project/imodel header.

### To Reproduce

_No response_

### Expected Behavior

I would expect the popout to work even when useDefaultPopUrl is set to `false`.

### Screenshots

_No response_

_No response_

### Additional context

_No response_

GerardasB:
## Changes

This PR fixes #1178 by replacing deprecated [`document.write()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/write) with DOM manipulation methods.

Popout widget would not close correctly, due to [`pagehide`](https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event) listener not being called. See https://developer.mozilla.org/en-US/docs/Web/API/Document/open#notes
I.e. using [`document.close()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/close) (after writing to the document) and then adding the `pagehide` listener seems to work correctly as well.

## Testing

N/A

### DIFF
--- a/apps/test-app/.env.template
+++ b/apps/test-app/.env.template
@@ -27,5 +27,3 @@ IMJS_IMODEL_VIEWID=
 IMJS_ELECTRON_NO_DEV_TOOLS=
 # Used to toggle the environments.
 IMJS_URL_PREFIX=
-# If defined, uses the default popout URL.
-IMJS_USE_DEFAULT_POPOUT_URL=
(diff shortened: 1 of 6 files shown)


saskliutas (approved):

GerardasB:

PR #1190, opened 2025-01-27
Use iTwinUI Tab instead of our current one

MBudreviciusBentley:
## Changes

Rewrite our current TabBar/Tabs implementation to use iTwin elements.

Some of the visual changes:
image

## Testing

Tests + manually playing around with tabs.


MBudreviciusBentley:
Investigate: why Buttons doesn't take up the whole width. 
image
(Right now we use div with buttons className which is used in tests to find components, if using empty react fragment <>...</> then problem is fixed)

MBudreviciusBentley:
Investigate: Tabs height is a bit bigger than the div itself. 
image

GerardasB (commented):
Some issues while playing around with the storybook:
- [] Widget tabs are bigger in height than the title bar area
- Double borders between tabs (and on first tab)
- Tab labels (and borders) are wrapping outside of tab area
- Dragging a single tab should not undock a whole stacked widget w/ multiple tabs
- No longer possible to drop widget on title bar tab target

GerardasB:
Some teams are adding buttons (i.e. to close a widget) to the widget tabs. This is blocked until iTwinUI adds support for tab actions, since currently tabs are implemented as buttons. @mayank99, @r100-stack, @FlyersPh9

PR #1210, opened 2025-02-06
Add `useSavedState` flag to `Widget` config

### LINKED ISSUE #1201: Widgets that default to "Hidden" don't save their state between application restarts
### Describe the bug

When widgets are created with `defaultState: WidgetState.Hidden`, they will not save their state between the application reopening. The widget will always start hidden.

The state will be restored correctly after switching frontstages. Including opening a different iModel.

### To Reproduce

1. Create a UIProvider that
  - provides a widget with the default state to hidden.
  - has some way to open the widget.
2. Open the widget
3. Restart the application and open an iModel

The widget will start hidden.

### Expected Behavior

I expect the widget would be restored to its former visibility, like all the other default widget states.

### Screenshots

_No response_

Desktop app
AppUI: 4.17.6
iTwinUI: 3.16.0

### Additional context

_No response_

GerardasB:
## Changes

This PR resolves #1201 by adding `useSavedState` flag to `Widget` configuration. When enabled the widget will **NOT** be hidden after the layout is restored (i.e. page reload) when defaultState is set to Hidden (as described [in the docs](https://www.itwinjs.org/reference/appui-react/widget/widget/defaultstate/#:~:text=If%20Hidden%20the%20widget%20will%20be%20hidden%20after%20the%20layout%20is%20restored%20independently%20of%20saved%20layout%20state.)).

## Testing

Added additional e2e tests

### DIFF
--- a/ui/appui-react/src/appui-react/widgets/Widget.tsx
+++ b/ui/appui-react/src/appui-react/widgets/Widget.tsx
@@ -61,6 +61,7 @@ export interface Widget {
   readonly defaultState?: WidgetState;
   /** When enabled, the widget will always use the saved layout state. `defaultState` will only be used for the initial layout setup.
    * This is useful for scenarios where the widget should only be hidden/unloaded initially, but is always restored to the last state.
+   * @note This property is only useful to disable the special handling of `defaultState` when it is set to `Hidden` or `Unloaded`.
    */
   readonly useSavedState?: boolean;
   /** Content of the Widget. */
--- a/ui/appui-react/src/appui-react/widgets/Widget.tsx
+++ b/ui/appui-react/src/appui-react/widgets/Widget.tsx
@@ -51,10 +51,18 @@ export interface Widget {
    * It is not possible to disable the floating of a widget if `allowedPanels` is an empty array.
    */
   readonly canFloat?: boolean | CanFloatWidgetOptions;
-  /** Defaults to `Floating` if widget is not allowed to dock to any panels. Otherwise defaults to `Closed`.
-   * @note If `Hidden` the widget will be hidden after the layout is restored independently of saved layout state.
+  /** The default state of the widget.
+   * - Defaults to `Floating` if the widget is not allowed to dock to any panels.
+   * - Otherwise, defaults to `Closed`.
+   *
+   * @note If set to `Hidden`, the widget will be hidden after the layout is restored, independently of the saved layout state. Set `useSavedState` to `true` to disable this behavior.
+   * @note If set to `Unloaded`, the widget will be unloaded after the layout is restored, independently of the saved layout state. Set `useSavedState` to `true` to disable this behavior.
    */
   readonly defaultState?: WidgetState;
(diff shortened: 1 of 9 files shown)


ui/appui-react/src/appui-react/widgets/Widget.tsx
```diff
@@ -51,10 +51,18 @@ export interface Widget {
...
+   * - Otherwise, defaults to `Closed`.
+   *
+   * @note If set to `Hidden`, the widget will be hidden after the layout is restored, independently of the saved layout state. Set `useSavedState` to `true` to disable this behavior.
+   * @note If set to `Unloaded`, the widget will be unloaded after the layout is restored, independently of the saved layout state. Set `useSavedState` to `true` to disable this behavior.
    */
   readonly defaultState?: WidgetState;
+  /** When enabled, the widget will always use the saved layout state. `defaultState` will only be used for the initial layout setup.
+   * This is useful for scenarios where the widget should only be hidden/unloaded initially, but is always restored to the last state.
+   */
+  readonly useSavedState?: boolean;
```
simihartstein, line 65: This will satisfy our requirements as-is, so thanks for getting this done so quickly.

It looks like this property will only change the functionality for widgets with default state Hidden. What if the default state is something visible, let's say Floating, but `useSavedState` is set to false? Then it wouldn't honor that property and use its saved state anyway.

GerardasB: That's correct I believe the confusion comes from the naming aspect, so alternative flag names are welcome..

GerardasB: Ideal API would probably look similar to this:
```
defaultState?: WidgetState;
state?: WidgetState.Hidden | WidgetState.Unloaded; // used to Hide/Unload widgets between re-loads
```

GerardasB: Alternatively, instead of a `boolean` we could change the flag to `undefined` or `true`?

simihartstein: Well, I actually prefer the naming as you have it now, just that the functionality is a bit inconsistent. Ideally, I would say the state should _always_ be restored (even hidden) AND it always honors your new `useSavedState` property.

But I get that you don't want to alter pre-existing behavior, so maybe it could function like I mentioned, but default `useSavedState` to false in the hidden/unloaded `defaultState` scenario, and true otherwise. Then, it just needs to be documented in a comment and it should be ok.

GerardasB: One more suggestion is to add `initialState` that would be used for initial setup only (which is what you need, although there would be a little bit of ambiguity w/ `defaultState`).

GerardasB: Thanks, added additional docs to clarify [f87f0c5](https://github.com/iTwin/appui/pull/1210/commits/f87f0c5d8c49d2e917ecf72dd0cf661bb3a1da7d)

PR #1211, opened 2025-02-10
Add `promptAtContent` to `ToolAssistanceField` components

### LINKED ISSUE #1167: Tool Assistance Popup Bugs
### Describe the bug

We want the tool assistance popup to be displayed by default and be persitent (visible without fading out after some milliseconds).

## Cannot hide when Displayed by default 
By setting `defaultPromptAtCursor` to true, we get the expected behaviour of having the cursor popup by default. 
Ex: ` content: <ToolAssistanceField defaultPromptAtCursor={true}/>`

The bug we are facing is that the tool assistance toggle button does not hide the popup. We must reopen the application for the changes to apply and get a hidden popup.

## Persistence
By setting `cursorPromptTimeout` to a very high value, the popup stays visible for a very long time, simulating a persistent poup. 
Ex : ` content: <ToolAssistanceField cursorPromptTimeout={900000000} />`

A bug became apparent with this. 
Popups are visible outside from the viewport. They are also visible when changing frontstages. 
In the image below we can see the popup visible on top of widgets. 
Image

Related to #1135 

### To Reproduce

_No response_

### Expected Behavior

_No response_

### Screenshots

_No response_

_No response_

### Additional context

_No response_

GerardasB:
## Changes

This PR fixes #1167 by adding `promptAtContent` prop to `ToolAssistanceField` which when enabled will hide the tool assistance prompt when non-content element is hovered (i.e. when hovering something outside from the view).
Additionally, `cursorPromptTimeout` now accepts `Number.POSITIVE_INFINITY` which will keep the prompt open indefinitely.
Cleanup of `ToolAssistanceField` is fixed as well: the prompt will be closed when the toggle is switched off or when the component is unmounted.

## Testing

Added additional stories.


common/api/appui-react.api.md
```diff
@@ -4674,6 +4677,7 @@ export interface ToolAssistanceFieldProps extends CommonProps {
     defaultPromptAtCursor: boolean;
     fadeOutCursorPrompt: boolean;
     includePromptAtCursor: boolean;
+    promptAtContent?: boolean;
```
mathieu-fournier, line 4680: Do we have any cases when we want a popup to be visible outside from the view ? 
Seems to me that this might be the default behaviour, I'm not sure we need to let the consumer control that.

GerardasB: Merging as is for better configurability

PR #1274, opened 2025-04-08
Skip flaky test

GerardasB:
## Changes

This PR skips flaky test which is causing issues in CI pipelines.

## Testing

N/A


saskliutas (approved):

GerardasB:
> I think the real fix to this would be to drop usage of faker. In this specific case, I bet the problem is in setting up `testItems` array.

Merging as is, since `faker` is only used by Presentation owned packages.

PR #1291, opened 2025-05-05
Replace `rush` with `pnpm`

GerardasB:
## Changes

This PR replaces usage of [`rush`](https://rushjs.io/) monorepo manager with `pnpm`.

`rush change`, custom changelog publishing and package publishing scripts are replaced by the usage of [`changesets`](https://github.com/changesets/changesets). New changeset is generated instead of `NextVersion.md`

New workflows and structure is based on https://github.com/iTwin/presentation

Unit tests will run on `macos` for now, since `ubuntu` and `windows` fail due to some flakiness.

## Testing

N/A


ui/components-react/src/test/editors/TextEditor.test.tsx
```diff
@@ -68,7 +68,11 @@ describe("<TextEditor />", () => {
         propertyDescription: PropertyDescription,
         value?: Primitives.Value | undefined
       ): string | Promise<string> {
-        return super.convertPropertyToString(propertyDescription, `-${value}`);
+        return super.convertPropertyToString(
+          propertyDescription,
+          // eslint-disable-next-line @typescript-eslint/no-base-to-string
```
saskliutas, line 73: I guess this ignore is not necessary after change to `string(value)`?

apps/test-app/src/frontend/appui/frontstages/EditorFrontstage.tsx
```diff
@@ -12,12 +12,12 @@ import {
...
-  ToolbarItemUtilities,
-  ToolbarOrientation,
-  ToolbarUsage,
+  // ToolbarItemUtilities,
+  // ToolbarOrientation,
+  // ToolbarUsage,
   UiItemsProvider,
 } from "@itwin/appui-react";
-import { CreateArcTool, CreateLineStringTool } from "@itwin/editor-frontend";
+// import { CreateArcTool, CreateLineStringTool } from "@itwin/editor-frontend";
```
saskliutas, line 20: Should this be removed?

.github/workflows/extract-api.yaml
```diff
@@ -17,30 +17,22 @@ jobs:
...
 
-      - name: Rush extract-api
-        run: node common/scripts/install-run-rush.js extract-api
+      - run: pnpm install
+      - run: pnpm build
+      - run: pnpm extract-api
 
-      - name: Publish extracted api
-        uses: actions/upload-artifact@v4
+      - uses: actions/upload-artifact@v4
```
saskliutas, line 35: What is happening with uploaded artifact? Maybe this job could be moved to `CI.yaml` so it could reuse cached `node_modules`?

Dockerfile
```diff
@@ -4,17 +4,18 @@
 #----------------------------------------------------------------------------------------------
 FROM mcr.microsoft.com/playwright:v1.46.1-jammy
 
-# Install rush
-RUN npm install -g @microsoft/rush
+# Install pnpm
+RUN --mount=type=cache,target=/root/.npm \
+    npm install -g pnpm@10
```
saskliutas, line 9: Is this necessary? Shouldn't node setup this through corepack now that package.json contains packageManager field?

ui/components-react/src/test/editors/TextEditor.test.tsx
GerardasB: Linter still complains, due to potential `[object Object]` when stringified.

apps/test-app/src/frontend/appui/frontstages/EditorFrontstage.tsx
GerardasB: [Remove unused editor tools](https://github.com/iTwin/appui/pull/1291/commits/1f871cd55405658ef6340f73e54b408b95139516)

Dockerfile
GerardasB: corepack complains about integrity key mismatches, I think playwright needs to be updated to resolve that (to use newer node version). Let's do that in a separate PR since it's a big change.

saskliutas (approved):

PR #1294, opened 2025-05-06
Update peer dependencies to support iTwin.js core v5

GerardasB:
## Changes

This PR updates the iTwin.js core `peerDependencies` of AppUI packages to include the v5 in the version range.
- Dev dependencies are bumped to latest `rc` which is `5.0.0-dev.111`.
- All iTwin.js core dependencies are specified using [catalogs](https://pnpm.io/catalogs).

API changes:
- `IDisposable` deprecation is handled by using `Symbol.dispose` and introducing a polyfill (for FF and Safari). Based on:
  - https://github.com/iTwin/presentation/pull/802
  - https://github.com/iTwin/presentation/pull/828

Misc changes:
- Fixed the unspecified types tsc issue in `AbstractUiItemsManager`.
- Fixed snapshot connection capability in `test-app` by using the internal `LocalhostIpcApp`.
  - `IMJS_TESTS` env variable can be set to continue using `IModelApp.startup` which is used to run tests w/o backend
- Deprecation linter errors are ignored when used in deprecated APIs.
- [Remove property.autoExpand delete workaround](https://github.com/iTwin/appui/pull/1294/commits/57c34aafbe39e48db5f76d65966790c257a6c2ca) after https://github.com/iTwin/itwinjs-core/pull/7563

## Testing

N/A


pnpm-workspace.yaml
```diff
@@ -4,3 +4,26 @@ packages:
   - e2e-tests
   - tools/codemod
   - docs/storybook
+
+catalog:
+  # iTwin.js core packages
```
saskliutas, line 9: Have you considered using named catalogs instead of a comment? E.g:
```yaml
catalogs:
  itwinjs-core:
    '@itwin/appui-abstract': rc
    ...
```

GerardasB: I did, but there's no use-case for named catalogs yet.

saskliutas (approved):

PR #1324, opened 2025-05-21
Fix toolbar font icon color when theme bridge is enabled

### LINKED ISSUE #1305: Toolbar icon fill color difference between SVG and font-based icons
### Describe the bug

After upgrading our Mango application to one of the latest versions (which has enabled the Kiwi theme bridge), we noticed that there are differences in fill color between SVGs and font-based icons. The SVGs look noticeably lighter than the font-based icons.

We use the helper method `ToolbarItemUtilities.createActionItem()` to create the `ToolbarActionItem`.

### To Reproduce

_No response_

### Expected Behavior

Toolbar icons should have the same fill color and look uniform.

### Screenshots

Image

 - iTwin.js AppUI: 5.4.0
 - iTwin.js: 4.11.0
 - Mango: 1.0.159
 - iTwinUI: 3.18.0

### Additional context

_No response_

GerardasB:
## Changes

~This PR fixes #1305 by setting the `color` property to match the `fill` property used by SVG icons to fix the icon color when font icons are used with iTwinUI theme bridge enabled.~ EDIT: this PR bumps iTwinUI version and adds additional stories to verify toolbar button icon colors when font icons are used. 

| Before | After |
| --- | --- |
| Screenshot 2025-05-21 at 12 10 36 | Screenshot 2025-05-21 at 12 08 04 |

Misc changes:
- Updated StrataKit version.

## Testing

Added additional stories.

### DIFF
--- a/apps/test-app/package.json
+++ b/apps/test-app/package.json
@@ -97,7 +97,7 @@
     "@itwin/imodels-access-frontend": "^5.2.3",
     "@itwin/imodels-client-authoring": "^5.9.0",
     "@itwin/imodels-client-management": "^5.9.0",
-    "@itwin/itwinui-react": "^3.17.0",
+    "@itwin/itwinui-react": "catalog:",
     "@itwin/itwinui-react-v2": "npm:@itwin/itwinui-react@^2.12.26",
     "@itwin/itwinui-icons-react": "^2.8.0",
     "@itwin/itwinui-layouts-react": "~0.4.1",
--- a/apps/test-providers/package.json
+++ b/apps/test-providers/package.json
@@ -57,7 +57,7 @@
     "@itwin/core-react": "workspace:*",
     "@itwin/appui-react": "workspace:*",
     "@itwin/imodel-components-react": "workspace:*",
-    "@itwin/itwinui-react": "^3.17.0",
+    "@itwin/itwinui-react": "catalog:",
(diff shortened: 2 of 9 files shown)


GerardasB:
@mayank99 this is a bit length, since I can't use internal iTwinUI CSS vars, but need to set `color` property to match the `fill`. Maybe iTwinUI could set `color` when a wrapper HTML element exists for the icon (i.e. on `_iui-button-icon`)?

saskliutas (approved):

mayank99 (commented):
> Maybe iTwinUI could set `color` when a wrapper HTML element exists for the icon (i.e. on `_iui-button-icon`)?

This is redundant but I think it should be fine? And it matches what we do in StrataKit.

Do you want to look into it and submit a PR? It may get a little tricky because iTwinUI icon code is written using [Sass mixins](https://github.com/iTwin/iTwinUI/blob/04619737aac75e0a07d4c53569ab7406f42d0b48/packages/itwinui-css/src/mixins.scss#L98-L129) which are used in _lots_ of different places.

Also left an unrelated comment on this PR.

GerardasB:
This will be fixed in iTwinUI instead https://github.com/iTwin/iTwinUI/pull/2554
I will still merge this in after updated the iTwinUI version to keep the stories (there should be no changes in any of the packages).

PR #1348, opened 2025-06-16
Highlight group items of `<ToolbarComposer />`

### LINKED ISSUE #1338: Highlight Group button when action button is selected in toolbar

When selecting an action item under a group button in the ToolbarItemUtilities, the item is selected but when the dropdown is closed, there's no way of knowing which one of the toolbar groups the active item is located under.
i.e. "Add Vertex" here is selected, but when the dropdown is closed, the Pencil icon is not highlighted to help us identify which menu item is currently active. 
Image

I would like the GroupButton to also be highlighted when one of its children is active.

### Describe alternatives you've considered

None for now

### Additional context

_No response_

GerardasB:
## Changes

This PR closes #1338 by adding group item highlighting to `ToolbarComposer`. With this change the group items that contain an active toolbar item will be rendered as active as well.

I.e. in this toolbar the `Item 3` is active (note: `Item 3` corresponds to an active tool, since `ToolbarComposer` determines which items are active based on the active tool id).

Screenshot 2025-06-16 at 19 04 58

## Testing

Updated unit test to assert for new behavior.
Added new stories: https://itwin.github.io/appui/storybook/?path=/story/components-toolbarcomposer--active-items

### DIFF
--- a/.changeset/sixty-cities-retire.md
+++ b/.changeset/sixty-cities-retire.md
@@ -1,0 +1,5 @@
+---
+"@itwin/appui-react": minor
+---
+
+Updated `ToolbarComposer` component to highlight group items that contain an active toolbar item.


arome:
thats what we were looking for, lgtm!

FlyersPh9:
At first I was thinking the active button would include the stripe:
Screenshot 2025-06-16 at 2 49 10 PM

But I think that button group as a parent is required for the stripe.

GerardasB:
> At first I was thinking the active button would include the stripe: 

The stripe is not displayed for active toolbar items. Group item is using the same styling as action toolbar item.

PR #1404, opened 2025-09-04
style: status bar items cleanup

leo-belanger:
## Changes

- SnapMode Icon is now at the start
- Fix tool assistance width
- Add padding for view selector

Screenshot 2025-09-04 at 13 09 35
Screenshot 2025-09-22 at 15 50 00

Solves part of https://dev.azure.com/bentleycs/iModelTechnologies/_workitems/edit/1737736

## Testing

Tested using test-app


FlyersPh9:
From the backlog item:

> Remove excessive spacing after "identify element"

If I recall correctly, the empty space is there because the tool assistance will display various different commands.  Those commands may be short or long in character length.  That empty space was intended to try to prevent the tool assistance's width changing resulting in shifting content.

GerardasB:
> That empty space was intended to try to prevent the tool assistance's width changing resulting in shifting content.

That's correct, @leo-belanger we'd need UX input to merge the tool assistance `width` change. Note that the tool assistance will display the main instruction of a tool, which usually changes frequently: both when another tool is activated AND when interacting with the same tool to indicate current user-action.

leo-belanger:
> Hello @leo-belanger ,
> 
> here's my two cents on this from a strata/design perspective. This whole status bar needs to be rethought ideally. Things are not ordered logically and more rules should be created to understand what should be displayed at all times vs what should just default in an overflow. The space is at a premium and every item creates cognitive load.
> 
> The change you're suggesting makes sense:
> 
>     * some labels are actually currently truncated due to the width of the toolAssistance button
> 
>     * to avoid layout shift as much as possible the toolAssistance should be placed in last position on the left
> 
>     * even if layout shift occurs the user's eyes are either on the toolbar or on the canvas so hopefully not too jarring
> 
>     * i think it would be wise to do rigorous testing to see how acceptable the layout shift would be before implementing this
> 
> 
> thanks!

Hey @ahilhorst ! In that case, I think the re-organization of the status bar could be done in another ticket that should first be refine by UI/UX. Because if we think that the status bar need to be rethought, there is certainly more stuff to do than just moving controls 😅

That way I can merge the current changes and the other improvements could be tackled later on. Wdyt?

GerardasB (approved):
LGTM, although we should probably deprecate `ListPicker`

GerardasB (changes requested):
You'll want to sign CLA, additionally generate 3 changeset files using `pnpm changeset`)

leo-belanger:
> You'll want to sign CLA, additionally generate 3 changeset files (using `pnpm changeset`)

@GerardasB done!

PR #1408, opened 2025-09-09
Enable accent on status bar elements

LukaszKokot:
## Changes
We apply colour accents to the status bar item container, and slightly adjust the button style. The idea is for status bar elements that have popovers to get the accent colour when the popover shows up.

## Testing
Tested using the test-app, both web and electron, and with and without the theme bridge.

Here is how it looks like with the theme bridge:
https://github.com/user-attachments/assets/5748cb1a-0e14-425b-812d-60b7448bce3d

and without:
https://github.com/user-attachments/assets/c74bb163-c457-4bad-b3c8-5ff518e73185

### DIFF
--- a/ui/appui-react/src/appui-react/statusbar/StatusBar.scss
+++ b/ui/appui-react/src/appui-react/statusbar/StatusBar.scss
@@ -71,6 +71,11 @@
         padding-left: 0;
       }
     }
+
+    & > [aria-haspopup] {
+      border-radius: 0;
+      border-width: 2px 0 0 0;
+    }
   }
 
   .uifw-statusBar-left .uifw-statusBar-item-container {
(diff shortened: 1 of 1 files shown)


GerardasB (commented):
In iTwinUI button trigger with an open popover is displayed as "hovered". Ideally we should avoid customizations like these +@FlyersPh9

LukaszKokot:
> In iTwinUI button trigger with an open popover is displayed as "hovered". Ideally we should avoid customizations like these +@FlyersPh9

Thanks @GerardasB for the quick review and explanation. You are absolutely right (to paraphrase GPTs 🙃), I also noticed hover was used as the trigger for this and I wasn't sure what to think of it. Also, I'll make sure to not use iTwinUI internals next time.

LukaszKokot:
Thanks again for the review @GerardasB. Closing this, as I got more information about the requirements, and the change is a bit larger than this (and the colour changes not on the bar buttons)

PR #1416, opened 2025-09-18
Add `ResizeObserver` to `ViewportComponent` as a workaround for parent zero dimension uncaught exception

jason-crow:
Adds a  `ResizeObserver` handler in the `ViewportComponent` implementation as a workaround for an uncaught error that is thrown when creating a viewport with a div that has a zero dimension

## Changes

- workaround for an issue where the ViewportComponent crashes the application if the parent div it is attempting to create a viewport inside has a width or height of 0

- the solution is to prevent creating a viewport when the parent div has a zero dimension and setup a ResizeObserver to wait until it has non zero dimensions before proceeding to create the viewport

## Testing
- if you assign display: none or width /height 0 to the application root, then you will see an error such as the one in the screenshot below if while those conditions apply to the root the ViewportComponent mounts and attempts to create a viewport
- image
- Note: the underlying error is thrown from https://github.com/iTwin/itwinjs-core/blob/532a0eddb0c98fc11fa48ecca4c7e668140f709d/core/frontend/src/Viewport.ts#L3149

### DIFF
--- a/.changeset/major-zoos-battle.md
+++ b/.changeset/major-zoos-battle.md
@@ -1,0 +1,5 @@
+---
+"@itwin/imodel-components-react": patch
+---
+
+Fixed `Cube` component to correctly hover the center face cell when cube is rotated.
--- a/ui/imodel-components-react/src/imodel-components-react/navigationaids/Cube.scss
+++ b/ui/imodel-components-react/src/imodel-components-react/navigationaids/Cube.scss
@@ -10,10 +10,13 @@ $cube-side-width: 12px;
   height: $cube-width;
   perspective: 300px;
   transform-style: preserve-3d;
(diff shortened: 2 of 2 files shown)


.changeset/funny-ads-write.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/imodel-components-react": patch
+---
+
+add a resize observer to the ViewportComponent as a workaround to throwing error when parent div has a zero dimension
```
GerardasB, line 5:
```suggestion
Fixed zero dimension error in `ViewportComponent` by using a resize observer.
```

GerardasB (approved):
This seems to work, although there's a lot of spaghetti in the `ViewportComponent` already.

Issue #1380, opened 2025-08-07
Use `StrataKit` design system in `AppUI`


GerardasB:

mayank99:
A few questions:
1. How many **icons** does AppUI use? Does StrataKit have equivalents for all of them?
2. What features is StrataKit **Toolbar** currently missing? We should prioritize those.
   - I think **vertical** orientation and **grouping** are the main ones. We might also need support for ["corner decorations"](https://github.com/user-attachments/assets/4fdb2a21-6263-4873-988e-1a25f85ae2a5).
3. What does "**Widgets**" entail? Which iTwinUI components is it currently using and do we have StrataKit equivalents for all of them? (e.g. tabs)
4. Are there any other prominent elements to consider (e.g. navigation aid)? Theme bridge may suffice for a lot of these, but it would still be nice to identify them.
   - The goal should be to move everything over to StrataKit eventually. Of course, this may not happen as part of the initial goal, or might happen with StrataKit's own layout system.

mayank99:
I've noticed a visual issue when the StrataKit theme bridge is enabled. When the left or right panel are collapsed (by resizing to 0 width), a "line grip" appears which is barely visible because of low color contrast, opacity, and thin lines.

<img width="140" height="378" alt="Screenshot showing a barely visible gripper at the edge of the viewport" src="https://github.com/user-attachments/assets/5b099ab3-cbc0-4e90-a500-f08f6fd4fc52" />

For now this could be addressed by changing some colors. I'm not sure if this issue is noticeable without theme bridge.

In the future, I imagine we want to replace this with a more prominent "collapse/expand" IconButton near the corner of the viewport.

<img width="324" height="217" alt="Screenshot showing what looks like a window collapse/expand button with solid styling, next to the navigation aid" src="https://github.com/user-attachments/assets/341f09b5-1a56-4e72-af57-dafc2dd444f1" />

mayank99:
Trying to identify blockers for the Toolbar adoption. From the list [above](https://github.com/iTwin/appui/issues/1380#issuecomment-3174980428):

> * `Toolbar` orientation
> * Corner decorations for custom and group items to visually indicate that particular items are expandable
> * `Popover` to display custom content in a panel for [`ToolbarCustomItem`](https://www.itwinjs.org/reference/appui-react/toolbar/toolbarcustomitem/). See: [custom item story](https://itwin.github.io/appui/storybook/?path=/story/components-toolbarcomposer--custom-item)
[...]

PR #1421, opened 2025-09-30
Fix: Bottom Status Bar - Update select state

jfcorr:
## Changes

This PR contains UI changes for the statusbar field MessageCenter and ToolAssistance to fit the UX recommendations
(screenshot above)
  - color accent in the popover
  - button with background when popover is opened (same as pressed and hover)
  - popover now aligned with the button locations (resize events are handled)

The following screenshots contain the actual changes that were made and the result:
Screenshot 2025-09-29 at 2 30 22 PM
Screenshot 2025-09-29 at 2 30 29 PM
Screenshot 2025-09-30 at 9 21 06 AM
Screenshot 2025-09-30 at 9 21 21 AM

### Task
https://dev.azure.com/bentleycs/iModelTechnologies/_workitems/edit/1737742

### Question
I planned to use @stratakit at first to easily edit the UI to fit the given requirements. Given that I don't know for sure who is consuming appui, I'd like to know if it is safe to assume that they will have the <Root> from @stratakit/foundations in its scope? Otherwise it will not wok properly (I may be wrong). That is why I stick to appui "things" only.

4080306f-4e28-4c5d-bf4e-bf92507873ad

I got another branch that contains the changes with stratakit/bricks, @stratakit/structure and @stratakit/icons
(plus foundation here for rendering the icons)

## Testing

Unit tests were added for the positional hook

### DIFF
--- a/ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
+++ b/ui/appui-react/src/appui-react/statusfields/message-center/MessageCenterField.tsx
@@ -28,7 +28,6 @@ import { StatusBarPopover } from "../../statusbar/popup/StatusBarPopover.js";
 import { usePosition } from "../usePosition.js";
 import classNames from "classnames";
 
-
 /** Type for Status state to satisfy NotificationMarker type checking. */
 type NotificationMarkerStatus = Required<
   React.ComponentProps<typeof NotificationMarker>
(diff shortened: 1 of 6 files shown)


GerardasB (changes requested):
I'd like to avoid these excessive styling overrides in iTwinUI components to match SK design language. AppUI needs to support applications running both iTwinUI and SK. This work could fall under https://github.com/iTwin/appui/issues/1380 where `MessageCenter` could be implemented from scratch using SK components (and would replace existing field when `useStrataKit` preview feature is enabled). @mayank99

Where is the design coming from? Do we really need style overrides for all these SK components? @iTwin/design-system-designers

PR #1443, opened 2025-10-17
Add support for active widget label visibility in ConfigurableUiContent

arome:
## Changes

Add an option in ConfigurableUiContent to always display the label of the active tab when only the icons are displayed.

## Testing

Added stories to cover some options in ConfigurableUiContent.

### DIFF
--- a/common/api/appui-react.api.md
+++ b/common/api/appui-react.api.md
@@ -3928,7 +3928,7 @@ export const SessionStateActions: {
     setNumItemsSelected: (numSelected: number) => ActionWithPayload_2<SessionStateActionId.SetNumItemsSelected, number>;
     setIModelConnection: (iModelConnection: any) => ActionWithPayload_2<SessionStateActionId.SetIModelConnection, any>;
     setSelectionScope: (activeSelectionScope: string) => ActionWithPayload_2<SessionStateActionId.SetSelectionScope, string>;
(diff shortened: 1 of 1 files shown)


ui/appui-react/src/appui-react/configurableui/ConfigurableUiContent.tsx
```diff
@@ -73,6 +73,8 @@ export interface ConfigurableUiContentProps extends CommonProps {
   widgetOpacity?: number;
   /** Controls if the widget icons should be displayed. Uses redux store as a fallback. Defaults to `true`. */
   widgetIcon?: boolean;
+  /** Controls if the active widget's label should always be displayed. Defaults to `false`. */
+  activeWidgetLabel?: boolean;
```
GerardasB: Let's rename to `displayActiveWidgetLabel`/`showActiveWidgetLabel` to indicate a `boolean` (even if this is consistent with poorly named `widgetIcon` above)

docs/storybook/src/frontstage/ConfigurableUiContent.stories.tsx
```diff
@@ -0,0 +1,187 @@
...
+  args: {
+    widgetOpacity: 0.7,
+    toolbarOpacity: 0.6,
+    activeWidgetLabel: true,
+    widgetIcon: true,
+  },
+};
+
+// Example showing how props could be changed at runtime (controls re-render via state).
+export const DynamicOpacity: Story = {
```
GerardasB: This seems to re-initialize the whole story (rather than dynamically changing the transparency). Missing toolbars as well. (can be removed for now)

PR #1480, opened 2025-12-15
Add `getPanels` to `UiItemsProvider`

GerardasB:
## Changes

This PR adds `getPanels` property to `UiItemsProvider` interface.
Standard layout now supports dynamic panels with `"left"` and `"right"` placements.
Additionally, imperative `panels` API is exposed via `FrontstageDef` class.

## Preview

https://itwin.github.io/appui/1480/?path=/story/frontstage-panels--dynamic-panel


grigasp:
🕐

ui/appui-react/src/appui-react/panel/Panel.ts
```diff
@@ -0,0 +1,40 @@
...
+
+interface DynamicPanel extends CommonPanel {
+  readonly type: "dynamic";
+  readonly placement?: "left" | "right" | (string & {});
+}
+
+/** Describes the data needed to provide a panel.
+ * @public
+ */
+export type Panel = CommonPanel | InformationPanel | DynamicPanel;
```
grigasp, line 35: Is there documentation somewhere that I missed, describing the differences between these panel types?

grigasp (commented):
Not competent to review the implementation, but looked at the API and the story - AFAICT the dynamic panel meets our needs. Would ask @saskliutas for a deeper look.

ui/appui-react/src/appui-react/panel/Panel.ts
GerardasB: I believe `InformationPanel` is just a placeholder for specifically https://itwinui.bentley.com/docs/informationpanel (one more panel type, that overlaps other stage panels/dynamic panels).
The idea was to use the same set of APIs, with slightly different definitions.
We should remove `InformationPanel` for now.

saskliutas (approved):

PR #1490, opened 2026-01-12
Remove noisy CursorMenu logging

alexdunae:
## Changes

Removes noisy logging

This log line can be emitted hundreds of times when starting an iTwin.js app

Screenshot 2026-01-12 at 10 07 45 AM

## Testing

`pnpm test`


GerardasB (approved):

GerardasB:
This "solves" the consequence, but it might be worth looking into underlying reason of why is this called/rendered so many times.

PR #1506, opened 2026-02-04
Fix spacing in WidgetContentLayout when some options are enabled

arome:
## Changes

| What | Before | After |
| ---- | ---- | ---- | 
| Fix spacing when hideDivider is enabled |  image |  image |
| Fix extra top spacing when only custom header is provided | image |  image |


docs/storybook/src/widget/WidgetContentLayout/WidgetContentLayout.Header.stories.tsx
GerardasB: Is this change intentional?

GerardasB (approved):
Need `pnpm changeset`.

PR #1508, opened 2026-02-13
fix(a11y): add id={property.name} to all property editors for label association

arome:
## Summary

All property editor components (both legacy and new editor system) now set `id={propertyRecord.property.name}` on their primary interactive element. This enables proper `<label htmlFor>` association and makes editors uniquely targetable in tests via `getByLabelText`.

## Problem

The legacy editors relied on a shared, hardcoded `data-testid` (e.g. `"components-select-editor"`). When multiple editors of the same type appeared on the same page — e.g. two enum selects in a ToolSettings panel — they were indistinguishable. The new editor system had no `id` at all.

### New editor system (`editorSystem="new"`)
- Added `id?: string` to `EditorProps` in `Types.ts`
- Threaded `id={propertyRecord.property.name}` from `PropertyRecordEditor` → `CommittingEditor` → `EditorRenderer` → each concrete editor
- Applied to: `TextEditor`, `NumericEditor`, `EnumEditor`, `BooleanEditor`, `ToggleEditor`, `FallbackEditor`, `DateEditor`, `DateTimeEditor` (via `DateInput`)
- Applied to all old-editor interop wrappers: `TextArea`, `Color`, `CustomNumber`, `EnumButtonGroup`, `NumericInput`, `Slider`

### Legacy editor system
- `EnumEditor` → `id` on `<Select>`
- `BooleanEditor` → `id` on `<Checkbox>`
- `ToggleEditor` → `id` on `<ToggleSwitch>`
- `ImageCheckBoxEditor` → `id` on `<input type="checkbox">` (required adding `id` prop to `ImageCheckBoxProps` in core-react)
- `NumericInputEditor` → `id` on `<input>` via `NumericEditor`
- `TextareaEditor`, `DateTimeEditor`, `SliderEditor` → `id` on outer container div (popup-based; interactive content is not always in DOM)
- `EnumButtonGroupEditor` → `id` on container `<div>`
- `IconEditor` → `id` on `<IconPickerButton>`
- `TextEditor` and `CustomNumberEditor` already had `id` set

## Testing
- 4 integration tests added to `PropertyRecordEditor.test.tsx` covering label association via `getByLabelText`
- 1 unit test added to `EditorRenderer.test.tsx` verifying `id` forwarding


GerardasB (commented):
Property editor elements are (usually) created by `EditorContainer`, how do you plan to pass down the `itemId` prop? 
Additionally, `itemId` prop documentation states, that it adds `data-item-id`, not `data-testid`: https://github.com/iTwin/appui/blob/17e28f44bae899f9277e82814975ec61e3be4eb3/ui/core-react/src/core-react/utils/Props.tsx#L28 (might be safer to add new attribute, rather than breaking existing?)

arome:
> Property editor elements are (usually) created by `EditorContainer`, how do you plan to pass down the `itemId` prop? Additionally, `itemId` prop documentation states, that it adds `data-item-id`, not `data-testid`:
> 
> https://github.com/iTwin/appui/blob/17e28f44bae899f9277e82814975ec61e3be4eb3/ui/core-react/src/core-react/utils/Props.tsx#L28
> 
> (might be safer to add new attribute, rather than breaking existing?)

indeed i misread, i thought it added data-testid not data-item-id, will add a new attribute instead like you suggested

arome:
> How about usage? Property editor elements are (usually) created by `EditorContainer`, how do you plan to pass down the `itemId` prop?

Ya I looked more into it and the change actually needs to be made in [itwinjs-core](https://github.com/iTwin/itwinjs-core/pull/9011) first. I think you saw the discussion there, I'll wait to see from our QA team if they're fine with using something other than the data-testid and will update this PR accordingly

ui/components-react/src/components-react/editors/BooleanEditor.tsx
```diff
@@ -139,7 +139,8 @@ export class BooleanEditor
           this.props.propertyRecord?.isDisabled ||
           this.props.propertyRecord?.isReadonly
         }
-        data-testid="components-checkbox-editor"
+        data-testid={this.props.itemId ?? "components-checkbox-editor"}
```
saskliutas: I guess `this.props.itemId ?? "components-checkbox-editor"` change can be reverted after `id` was introduced

ui/components-react/src/components-react/editors/BooleanEditor.tsx
```diff
@@ -139,7 +139,8 @@ export class BooleanEditor
...
+        id={this.props.propertyRecord?.property.name}
```
GerardasB, line 143: The [`id`](https://developer.mozilla.org/en-US/docs/Web/API/Element/id) properties must be unique in a document.
Looks like the same property record (with the same property name, which results to duplicate ids) might be rendered multiple times e.g. in a property grid. @saskliutas

saskliutas: Double checked `PropertyRecord`s we created for iModel element property grid and it looks like we should create unique names inside that single property grid. But there might be custom property grids or I guess multiple iModel property grids present that might result in duplicating ids. 

So I think we should think about how to ensure unique ids in document scope.

ui/components-react/src/components-react/editors/BooleanEditor.tsx
arome: reverted the previous changes to data-testid

ui/components-react/src/components-react/editors/BooleanEditor.tsx
arome: Any idea how we can tackle this? The only solution i can think of is allowing the user to pass in the unique id, but this will take us back to the initial problem I was trying to solve in this [PR](https://github.com/iTwin/itwinjs-core/pull/9011). It doesnt have to be data-testid but we should have a way to pass some unique id from these generic property builder.

GerardasB: To avoid introducing new context based APIs and to keep this explicit we could:
- Add `id` prop to components that are used by consumers to render the editors: `EditorContainer`  and new ones - `PropertyRecordEditor`, `EditorRenderer`. Editors should use the passed down `id` and set it individually. This would be available in custom editors as well.
- Update `ComponentGenerator` to correctly associate labels and editors using the new prop. This might be the complicated part, because `ComponentGenerator` is just a class for some reason and not a component, so `React.useId()` is not available. `Guid` could be used as a workaround, but ideally `ComponentGenerator` should be **deprecated** in favor of React components (in the end it just generates some JSX).

Additionally, we will not be able to label all editors using the [`for` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label#attributes), in some cases label id (or even the label) will need to be known by the editor.

---

To unblock this PR we can leave all of the above for future improvements (ideal case), property names are currently used in the labels, so current solution should still be an improvement.

.changeset/sweet-islands-dance.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/components-react": minor
+---
+
+Provide option to define data-testids for PropertyEditor components
```
GerardasB: This is outdated

arome: I updated the changeset wording

ui/components-react/src/components-react/new-editors/editors/TextEditor.tsx
GerardasB, line 1: Some editors were missed (the `id` is not passed through), see https://github.com/iTwin/appui/blob/master/ui/appui-react/src/appui-react/new-editors/TextEditor.tsx#L36 (this applies to all editors in https://github.com/iTwin/appui/tree/master/ui/appui-react/src/appui-react/new-editors)

ui/components-react/src/components-react/editors/EnumEditor.tsx
```diff
@@ -181,6 +181,7 @@ export class EnumEditor
           value={selectValue}
           onChange={this._updateSelectValue}
           data-testid="components-select-editor"
+          id={this.props.propertyRecord?.property.name}
```
GerardasB: Should this be set on `triggetProps` so that the `id` attribute would be set on element with [`role="combobox"`](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/combobox_role)?
Either way, this will not work with the `for` attribute of the [`<label>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/label), since the element is not https://developer.mozilla.org/en-US/docs/Web/HTML/Guides/Content_categories#labelable.

GerardasB (approved):

ui/components-react/src/components-react/new-editors/editors/TextEditor.tsx
arome: Addressed in 2439baa17 — TextEditor and NumericEditor now forward the `id` prop to `InputWithDecorations.Input`. CustomNumberEditor already spreads all props through.

ui/components-react/src/components-react/editors/EnumEditor.tsx
arome: Addressed in 67d68d5a0 — moved the id to `triggerProps` so it lands on the `role="combobox"` element.

ui/components-react/src/components-react/editors/DateTimeEditor.tsx
```diff
@@ -291,7 +291,11 @@ export class DateTimeEditor
     );
 
     return (
-      <div className={className} ref={this._divElement}>
+      <div
+        className={className}
+        ref={this._divElement}
+        id={this.props.propertyRecord?.property.name}
+      >
         <PopupButton
```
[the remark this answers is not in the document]
arome: Addressed in 4d8c1afc1 — added an optional `id` prop to `PopupButton` and DateTimeEditor now passes `id={property.name}` to the PopupButton instead of the wrapper div.

ui/components-react/src/components-react/editors/TextareaEditor.tsx
```diff
@@ -208,7 +208,11 @@ export class TextareaEditor
     textareaProps["aria-label"] = UiComponents.translate("editor.textarea");
 
     return (
-      <div className={className} ref={this._divElement}>
+      <div
+        className={className}
+        ref={this._divElement}
+        id={this.props.propertyRecord?.property.name}
+      >
```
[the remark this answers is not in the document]
arome: Addressed in 4d8c1afc1 — TextareaEditor now passes `id={property.name}` to the focusable `PopupButton` trigger instead of the wrapper div.

ui/components-react/src/components-react/editors/SliderEditor.tsx
```diff
@@ -317,7 +317,11 @@ export class SliderEditor
     );
 
     return (
-      <div className={className} ref={this._divElement}>
+      <div
+        className={className}
+        ref={this._divElement}
+        id={this.props.propertyRecord?.property.name}
+      >
         <PopupButton
```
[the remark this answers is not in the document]
arome: Addressed in 4d8c1afc1 — SliderEditor now passes `id={property.name}` to the `PopupButton` trigger instead of the wrapper div.

ui/components-react/src/test/new-editors/interop/PropertyRecordEditor.test.tsx
```diff
@@ -42,4 +42,89 @@ describe("PropertyRecordEditor", () => {
...
+
+      await waitFor(() =>
+        expect(container.querySelector("input")).not.toBeNull()
+      );
+      // Legacy EditorContainer uses property.name as id too, but through a different path.
+      // This test just guards that the new-editor id is not present on the container element.
+      expect(
+        container.querySelector(".components-editor-container")
+      ).not.toBeNull();
+    });
```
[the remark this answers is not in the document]
arome: Addressed in cf5fe7555 — renamed the test and now asserts that the legacy editor renders `<input id="arcLength">` (the legacy primitive editors set `id={property.name}` on their input directly). Removed the misleading comment.

saskliutas (approved):

PR #1531, opened 2026-04-09
[Property grid]: update merged value renderer to render `-- unit` instead of `*** varies ***`

### LINKED ISSUE #1189: Components: Change the way property editors represent merged values
> [@fmikai](https://github.com/fmikai) there is another issue with displaying placeholders, ux input would be greatly appreciated: Currently, when displaying multiple values for a property in read-only mode, we show **varies** text. The issue arises when we have editors always displayed — in this case, we do not display any value, which can be misleading for users, because there are two different values for the property instead of no value displayed.
> 
> Example: Node 1 has length set to 5m. Node 2 has length set to 3m. When both nodes are selected:
> 
> * Read-only mode: Property grid displays **varies**
> * Always show editors mode: Property grid displays nothing (only a greyed-out placeholder)
> 
> @MartynasStrazdas In the property grid, whenever we have a multiple selection with different values, we display two dashes, like --.
> 
> Could we do something like this:
> - Disabled field in read only and active filed in show-editor. Both using the two dashes: --
> Image
>  

 _Originally posted by @fmikai in [#1591](https://github.com/iTwin/itwinjs-backlog/issues/1591#issuecomment-3784886310)_

MartynasStrazdas:
## Changes

Updated `MergedPropertyValueRenderer` to allow rendering `-- unit` style values for merged properties, when provided value does not start with `--` it would override value to `--` instead of `*** varies ***`

image


MartynasStrazdas:
@copilot resolve the merge conflicts in this pull request

saskliutas (approved):
Missing change file

GerardasB (approved):

PR #1535, opened 2026-04-14
Added `maxSize` property to `InputEditorSizeParams` interface of `@itwin/appui-abstract`

ryanlitzke:
## Changes

While creating tool settings, I was running into the issue where the `InputEditorSizeParam.size` was not having any effect on the size of the dialog property I was trying to change. Upon evaluation of the rendering logic, I came to the conclusion that  `InputEditorSizeParam.size` was being set to `minSize` what doesn't necessarily define the width of the dialog item. 

I am proposing changing the value from reading `minSize `to `inlineSize `so the user can control the width of the dialog.

## Testing

Tested in web application


GerardasB (commented):
There are multiple editors that use the same logic and every or none should be updated.
However, I think this change should not be applied - size is mainly used to ensure that enough space is allocated to the editor w/o collapsing (i.e. tool settings have an overflow logic, that allows inputs to shrink in size first and then move into the overflow menu).

You could implement a custom editor if you'd like custom styling (i.e. to apply max-width). @saskliutas for other ideas 

Additionally, I'd recommend to enable new editors: https://github.com/iTwin/appui/blob/3585ac3e7a97e7b75885a1c8a7e86096d5e0e4be/ui/appui-react/src/appui-react/preview/PreviewFeatures.tsx#L75

ryanlitzke:
> There are multiple editors that use the same logic and every or none should be updated. However, I think this change should not be applied - size is mainly used to ensure that enough space is allocated to the editor w/o collapsing (i.e. tool settings have an overflow logic, that allows inputs to shrink in size first and then move into the overflow menu).
> 
> You could implement a custom editor if you'd like custom styling (i.e. to apply max-width). @saskliutas for other ideas
> 
> Additionally, I'd recommend to enable new editors:
> 
> https://github.com/iTwin/appui/blob/3585ac3e7a97e7b75885a1c8a7e86096d5e0e4be/ui/appui-react/src/appui-react/preview/PreviewFeatures.tsx#L75

@GerardasB I see what you mean by the overflow logic control. If size is used to ensure that enough space is allocated to the editor w/o collapsing, then how do you suggest I control the width of the editor? The overflow space is more than I need / require. This seems like too common of a design control to create a custom editor, no?

See ex here:
image

ryanlitzke:
It appears that [here](https://github.com/iTwin/appui/blob/master/ui/components-react/src/components-react/new-editors/editors/FormattedNumericInput.tsx#L29) the formatted numeric input only accepts small or large and cannot reflect the size that I am trying to pass in. It is also always set to small.

GerardasB:
> It appears that [here](https://github.com/iTwin/appui/blob/master/ui/components-react/src/components-react/new-editors/editors/FormattedNumericInput.tsx#L29) the formatted numeric input only accepts small or large and cannot reflect the size that I am trying to pass in. It is also always set to small.

For this use-case, we could look into introducing an editor param for styling (would allow specifying style/className), which could be passed in when creating a property record https://github.com/iTwin/appui/blob/3585ac3e7a97e7b75885a1c8a7e86096d5e0e4be/apps/test-providers/src/tools/SampleTool.tsx#L233-L239

This needs more investigation, since some editors might have multiple elements w/o a shared ancestor/container in an editor (i.e. button with a popover).

Could you confirm if you have `toolSettingsNewEditors` preview feature enabled or not?

GerardasB:
As discussed internally, we can look into augmenting the `InputEditorSizeParams` interface with a `maxSize` property.

ryanlitzke:
> As discussed internally, we can look into augmenting the `InputEditorSizeParams` interface with a `maxSize` property.

Commit [fbf4d7f](https://github.com/ryanlitzke/appui/commit/fbf4d7fa2cbd749512bc13d42fa4d4ba5cfb4d4a) adds `maxSize` which sets `maxWidth` styling

ui/appui-react/src/appui-react/core-icons.ts
```diff
@@ -11,3 +11,10 @@ declare module "@itwin/core-frontend" {
     iconElement?: ReactElement;
   }
 }
+
+declare module "@itwin/appui-abstract" {
+  interface InputEditorSizeParams {
+    /** The max-width for an input editor. */
+    maxWidth?: number;
```
GerardasB: Let's keep consistent with current definitions https://github.com/iTwin/itwinjs-core/blob/30301a637afd5680e27fc67f09cf7d0a60c12baa/ui/appui-abstract/src/appui-abstract/properties/EditorParams.ts#L92-L93

```suggestion
    /** Optionally define the max width in characters. */
    maxSize?: number;
```

ui/components-react/src/components-react/editors/CustomNumberEditor.tsx
GerardasB, line 1: https://github.com/iTwin/appui/blob/3585ac3e7a97e7b75885a1c8a7e86096d5e0e4be/ui/components-react/src/components-react/new-editors/interop/old-editors/CustomNumber.tsx#L56 needs to be updated as well?

ui/components-react/src/components-react/editors/NumericInputEditor.tsx
GerardasB, line 1: https://github.com/iTwin/appui/blob/3585ac3e7a97e7b75885a1c8a7e86096d5e0e4be/ui/components-react/src/components-react/new-editors/interop/old-editors/NumericInput.tsx#L39 needs to be updated as well?

ui/appui-react/src/appui-react/core-icons.ts
GerardasB, line 1: ~let's rename this to `core-types.ts`~

EDIT: seems like build is failing due to `TS2339: Property 'maxWidth' does not exist on type 'InputEditorSizeParams'.`. This makes sense, because `@itwin/components-react` is a dependency of `@itwin/appui-react` (where the module augmentation is happening). We should introduce `abstract-types.ts` to components-react package.

ui/appui-react/src/appui-react/core-icons.ts
```diff
@@ -11,3 +11,10 @@ declare module "@itwin/core-frontend" {
     iconElement?: ReactElement;
   }
 }
+
+declare module "@itwin/appui-abstract" {
```
GerardasB: You'll want to generate a changeset file as well, using `pnpm changeset`, see https://github.com/iTwin/appui/pull/1439/changes#diff-6832a8fa66e5767503ca7489efbb021f19611ea441d8d77ce308a755d8478355 for an example
(make sure `components-react` has the associated change, not `appui-react` https://github.com/iTwin/appui/pull/1535#discussion_r3099305328)

ui/appui-react/src/appui-react/core-icons.ts
ryanlitzke: resolved: https://github.com/ryanlitzke/appui/commit/294bafab5a5f9a24a21a26fbc9975a41db6f6c5d

ui/components-react/src/components-react/editors/CustomNumberEditor.tsx
ryanlitzke: resolved: https://github.com/ryanlitzke/appui/commit/294bafab5a5f9a24a21a26fbc9975a41db6f6c5d

ui/components-react/src/components-react/editors/NumericInputEditor.tsx
ryanlitzke: resolved: https://github.com/ryanlitzke/appui/commit/1945a74450c33b496aa5c2969267a8d5162baad9

ui/appui-react/src/appui-react/core-icons.ts
ryanlitzke: resolved: https://github.com/ryanlitzke/appui/commit/aa1d5933fe4fde7f4944be4aaf6bd94a892faef7

ui/appui-react/src/appui-react/core-icons.ts
ryanlitzke: resolved: https://github.com/ryanlitzke/appui/commit/33b0a567bfb4866dfce906489c1effe651046452

ui/appui-react/src/appui-react/core-icons.ts
GerardasB: We should clarify in the comment that the unit/value is in characters.

.changeset/metal-shrimps-glow.md
```diff
@@ -0,0 +1,5 @@
+---
+"@itwin/components-react": minor
+---
+
+Added `maxSize` property to `InputEditorSizeParams` interface of `@itwin/components-react` package (using TypeScript declaration merging) to allow max sizing for tool settings input properties.
```
GerardasB: Fixed package name and nit on wording.

```suggestion
Added `maxSize` property to `InputEditorSizeParams` interface of `@itwin/appui-abstract` package (using TypeScript declaration merging) to allow setting the max size for input editors.
```

ui/appui-react/src/appui-react/core-icons.ts
ryanlitzke: added unit: https://github.com/ryanlitzke/appui/commit/467f0c464a2bc2908c5d2f8788ab1921c50b944f

GerardasB (approved):

PR #1544, opened 2026-04-30
[Editors]: update Koq, numeric and text editors to support constraints

### LINKED ISSUE #779: Support ECProperty value range
ECProperty supports adding constraints to values. 
https://www.itwinjs.org/bis/ec/ec-property/#additional-optional-attributes

These are not enforced by the ECDb.
They should be enforced by the inputs in the property grid.

To query those, you can use the [meta.ECPropertyDef](https://www.itwinjs.org/learning/ecdbmeta.ecschema/#ecpropertydef) table

JonasDov:
Part of https://github.com/iTwin/presentation/issues/779

### DIFF
--- a/ui/components-react/src/components-react/new-editors/editors/NumericEditor.tsx
+++ b/ui/components-react/src/components-react/new-editors/editors/NumericEditor.tsx
@@ -32,29 +32,15 @@ export function NumericEditor({
       value={currentValue.displayValue}
       onChange={(e) => {
         const parsedValue = parseFloat(e.target.value);
-        onChange(
-          {
-            rawValue: Number.isNaN(parsedValue) ? undefined : parsedValue,
-            displayValue: e.target.value,
-          },
-          () => {
-            const { min, max } = getNumericConstraints(metadata.constraints);
-            if (
-              !Number.isNaN(parsedValue) &&
-              (min !== undefined || max !== undefined)
-            ) {
-              const clamped = applyNumericConstraints({
+        onChange({
+          rawValue: Number.isNaN(parsedValue)
+            ? undefined
+            : applyNumericConstraints({
+                ...getNumericConstraints(metadata.constraints),
                 value: parsedValue,
-                min,
-                max,
-              });
-              if (clamped !== parsedValue) {
-                return { rawValue: clamped, displayValue: `${clamped}` };
-              }
-            }
-            return { rawValue: parsedValue, displayValue: e.target.value };
-          }
-        );
+              }),
+          displayValue: e.target.value,
+        });
       }}
       size={size}
       disabled={disabled}
--- a/ui/components-react/src/components-react/new-editors/editors/NumericEditor.tsx
+++ b/ui/components-react/src/components-react/new-editors/editors/NumericEditor.tsx
@@ -6,8 +6,12 @@
 import * as React from "react";
 import { Input } from "@itwin/itwinui-react";
 import type { EditorProps } from "../Types.js";
-import type { ValueMetadata } from "../values/Metadata.js";
+import type { NumericValueMetadata } from "../values/Metadata.js";
 import type { NumericValue } from "../values/Values.js";
+import {
+  applyNumericConstraints,
+  getNumericConstraints,
+} from "../ConstraintUtils.js";
 
 /* v8 ignore start */
 
@@ -16,23 +20,41 @@ import type { NumericValue } from "../values/Values.js";
  * @internal
  */
 export function NumericEditor({
+  metadata,
   value,
   onChange,
   size,
   disabled,
-}: EditorProps<ValueMetadata, NumericValue>) {
+}: EditorProps<NumericValueMetadata, NumericValue>) {
   const currentValue = getNumericValue(value);
   return (
     <Input
       value={currentValue.displayValue}
       onChange={(e) => {
         const parsedValue = parseFloat(e.target.value);
-        onChange({
-          rawValue: Number.isNaN(parsedValue)
-            ? undefined
-            : parseFloat(e.target.value),
-          displayValue: e.target.value,
-        });
+        onChange(
+          {
+            rawValue: Number.isNaN(parsedValue) ? undefined : parsedValue,
+            displayValue: e.target.value,
+          },
+          () => {
+            const { min, max } = getNumericConstraints(metadata.constraints);
+            if (
+              !Number.isNaN(parsedValue) &&
+              (min !== undefined || max !== undefined)
+            ) {
+              const clamped = applyNumericConstraints({
+                value: parsedValue,
+                min,
+                max,
+              });
+              if (clamped !== parsedValue) {
+                return { rawValue: clamped, displayValue: `${clamped}` };
+              }
+            }
+            return { rawValue: parsedValue, displayValue: e.target.value };
+          }
+        );
       }}
       size={size}
       disabled={disabled}
--- a/ui/components-react/src/test/new-editors/CommittingEditor.test.tsx
+++ b/ui/components-react/src/test/new-editors/CommittingEditor.test.tsx
@@ -3,23 +3,21 @@
  * See LICENSE.md in the project root for license terms and full copyright notice.
  *--------------------------------------------------------------------------------------------*/
 
-import { fireEvent, render } from "@testing-library/react";
+import { render } from "@testing-library/react";
 import { userEvent } from "@testing-library/user-event";
 import * as React from "react";
 import { describe, expect, it, vi } from "vitest";
 import { CommittingEditor } from "../../components-react/new-editors/interop/PropertyRecordEditor.js";
 import type { TextValueMetadata } from "../../components-react/new-editors/values/Metadata.js";
-import type { TextValue } from "../../components-react/new-editors/values/Values.js";
 
 describe("CommittingEditor", () => {
   const metadata: TextValueMetadata = { type: "string" };
-  const initialValue: TextValue = { value: "hello" };
 
   it("renders editor with initial value", () => {
     const { getByDisplayValue } = render(
       <CommittingEditor
         metadata={metadata}
-        initialValue={initialValue}
+        initialValue={{ value: "hello" }}
         onCommit={() => {}}
         onCancel={() => {}}
       />
@@ -36,14 +34,13 @@ describe("CommittingEditor", () => {
     const { getByRole } = render(
       <CommittingEditor
         metadata={metadata}
-        initialValue={initialValue}
+        initialValue={undefined}
         onCommit={onCommit}
         onCancel={onCancel}
       />
     );
 
     const input = getByRole("textbox");
-    await user.clear(input);
     await user.type(input, "world");
     await user.keyboard("{Enter}");
 
@@ -59,14 +56,13 @@ describe("CommittingEditor", () => {
     const { getByRole } = render(
       <CommittingEditor
         metadata={metadata}
-        initialValue={initialValue}
+        initialValue={undefined}
         onCommit={onCommit}
         onCancel={onCancel}
       />
     );
 
     const input = getByRole("textbox");
-    await user.clear(input);
     await user.type(input, "world");
     await user.keyboard("{Tab}");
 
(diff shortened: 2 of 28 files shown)


ui/components-react/src/components-react/new-editors/editors/NumericEditor.tsx
```diff
@@ -16,23 +20,41 @@ import type { NumericValue } from "../values/Values.js";
...
-            ? undefined
-            : parseFloat(e.target.value),
-          displayValue: e.target.value,
-        });
+        onChange(
+          {
+            rawValue: Number.isNaN(parsedValue) ? undefined : parsedValue,
+            displayValue: e.target.value,
+          },
+          () => {
```
saskliutas: This looks wrong to me. From this API usage it looks like code in this callback should not live inside this component. It is only needed when value is committed however this component does not now anything about committing it only allows to enter values.

I think in this component we can use constraints to set max/min properties on underlying components but validation should have in parent component that does "know" when commit is happening and can perform validation/clamping right before it.

This function can be delivered as some utility function if needed to avoid code duplication but not through `onChange` callback prop.

JonasDov: The reason for the implemenation was that I wanted to commit formatted display value. Since formatter exists at the type specific editor, needed some way to be able to format the value before it was commited. 

Decided to change the implementation: now `rawValue` always has constraints appplied, while `displayValue` is the same that is seen on the input. Added comment explaining this behavior for quantity editor.

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+  it("calls onChange when user types a value", () => {
+    const onChange = vi.fn();
+    renderQuantityEditor(
+      metadata,
+      { rawValue: 0, displayValue: "0 m" },
+      onChange
+    );
+    onChange.mockClear();
+
+    fireEvent.change(screen.getByRole("textbox"), {
```
saskliutas: These are new tests. You should use `@testing-library/user-event` instead of `fireEvent`. `@testing-library/user-event` better matches browser behavior

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+  };
+
+  it("renders input with formatted value", () => {
+    renderQuantityEditor(
+      metadata,
+      { rawValue: 5, displayValue: "5 m" },
+      () => {}
+    );
+
+    expect(screen.getByRole("textbox")).toHaveProperty("value", "5 m");
```
saskliutas: Can we use `getByDisplayValue("5 m")`?

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+}
+
+describe("QuantityEditor", () => {
+  const metadata: QuantityValueMetadata = {
+    type: "number",
+    quantityType: 1, // any value, hook is mocked
+  };
+
+  it("renders input with formatted value", () => {
+    renderQuantityEditor(
```
saskliutas: I think we should use `get*` selectors returned by `render` instead of directly using `screen`.

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
```diff
@@ -0,0 +1,171 @@
...
+  });
+
+  it("calls onChange when user types a value", () => {
+    const onChange = vi.fn();
+    renderQuantityEditor(
+      metadata,
+      { rawValue: 0, displayValue: "0 m" },
+      onChange
+    );
+    onChange.mockClear();
```
saskliutas: Why do we need mockClear on newly created mock?

ui/imodel-components-react/src/imodel-components-react/inputs/new-editors/QuantityEditor.tsx
```diff
@@ -71,3 +92,18 @@ function QuantityEditor({
 }
 
 /* v8 ignore stop */
+
+function applyConstraints(
```
saskliutas: I think you can export internal utilities from `components-react` through internal barrel file and use in this package. cc @GerardasB

ui/components-react/src/test/new-editors/CommittingEditor.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+    render(
+      <CommittingEditor
+        metadata={metadata}
+        initialValue={initialValue}
+        onCommit={() => {}}
+        onCancel={() => {}}
+      />
+    );
+
+    expect(screen.getByRole("textbox")).toHaveProperty("value", "hello");
```
saskliutas: You should be using `get*` selectors returned by `render`. Also probably should use `getByDisplayValue` here.

ui/components-react/src/test/new-editors/CommittingEditor.test.tsx
```diff
@@ -0,0 +1,206 @@
...
+      <CommittingEditor
+        metadata={metadata}
+        initialValue={initialValue}
+        onCommit={onCommit}
+        onCancel={onCancel}
+      />
+    );
+
+    const input = screen.getByRole("textbox");
+    fireEvent.change(input, { target: { value: "world" } });
```
saskliutas: Use `@testing-library/user-event` instead of `fireEvent`

saskliutas (changes requested):

ui/components-react/src/test/new-editors/CommittingEditor.test.tsx
saskliutas: Same in other test files

ui/components-react/src/test/new-editors/CommittingEditor.test.tsx
saskliutas: Same in other test files

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
JonasDov: Changed

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
JonasDov: Changed

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
JonasDov: Changed

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
JonasDov: Removed

ui/imodel-components-react/src/imodel-components-react/inputs/new-editors/QuantityEditor.tsx
JonasDov: Changed

ui/components-react/src/test/new-editors/CommittingEditor.test.tsx
JonasDov: Changed

ui/components-react/src/test/new-editors/CommittingEditor.test.tsx
JonasDov: Changed

ui/components-react/src/test/new-editors/NumericInputEditor.test.tsx
```diff
@@ -0,0 +1,117 @@
...
+  };
+}
+
+function renderNumericInput(
+  metadata: WithConstraints<OldEditorMetadata>,
+  value: { rawValue: number | undefined; displayValue: string },
+  onChange: (...args: any[]) => void
+) {
+  return render(
+    <EditorsRegistryProvider editors={[NumericInputEditorSpec]}>
```
saskliutas: Can we export NumericInputEditor and use it directly instead throug editors system like in other tests?

ui/components-react/src/test/new-editors/NumericEditor.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+    const { getByRole } = render(
+      <NumericEditor
+        metadata={metadata}
+        value={{ rawValue: undefined, displayValue: "" }}
+        onChange={onChange}
+      />
+    );
+
+    await user.click(getByRole("textbox"));
+    await user.paste("-5");
```
saskliutas: Why does these tests use `paste` instead of `type`?

ui/components-react/src/test/new-editors/CustomNumberEditor.test.tsx
```diff
@@ -0,0 +1,187 @@
...
+    const { getByRole } = render(
+      <CustomNumberEditor
+        metadata={createMetadata()}
+        value={{ rawValue: 0, displayValue: "0" }}
+        onChange={onChange}
+      />
+    );
+
+    const input = getByRole("textbox");
+    await user.clear(input);
```
saskliutas: Can you render editor with `undefined` value instead of `0` to not need `clear`?

ui/components-react/src/test/new-editors/CommittingEditor.test.tsx
```diff
@@ -0,0 +1,221 @@
...
+        onCommit={onCommit}
+        onCancel={onCancel}
+      />
+    );
+
+    const input = getByRole("textbox");
+    await user.clear(input);
+    await user.type(input, "changed");
+    await user.keyboard("{Escape}");
+    fireEvent.blur(input);
```
saskliutas: Can we use `keyboard("{Enter}")` here? Or maybe we don't need this test at all?

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+  "../../imodel-components-react/inputs/new-editors/QuantityEditor.js"
+);
+
+function renderQuantityEditor(
+  metadata: QuantityValueMetadata,
+  value: { rawValue: number | undefined; displayValue: string },
+  onChange: (...args: any[]) => void
+) {
+  return render(
+    <EditorsRegistryProvider editors={[QuantityEditorSpec]}>
```
saskliutas: Maybe export `QuantityEditor` and use directly?

ui/components-react/src/test/new-editors/CustomNumberEditor.test.tsx
saskliutas: Same in other test where `clear` is needed before `type`

ui/components-react/src/test/new-editors/NumericInputEditor.test.tsx
JonasDov: Changed

ui/components-react/src/test/new-editors/NumericEditor.test.tsx
JonasDov: Changed

ui/components-react/src/test/new-editors/CustomNumberEditor.test.tsx
JonasDov: Changed

ui/components-react/src/test/new-editors/CommittingEditor.test.tsx
JonasDov: Changed

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
JonasDov: Changed

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
```diff
@@ -0,0 +1,304 @@
...
+      const user = userEvent.setup();
+      const onChange = vi.fn();
+      const { getByRole } = renderQuantityEditor(
+        metadata,
+        { rawValue: 0, displayValue: "0 m" },
+        onChange
+      );
+
+      const input = getByRole("textbox");
+      await user.clear(input);
```
saskliutas: You can remove initial value in these tests to avoid `user.clear` calls in each test

saskliutas (approved):

ui/imodel-components-react/src/test/inputs/QuantityEditor.test.tsx
JonasDov: Removed `clear` completely

GerardasB (approved):

PR #1553, opened 2026-05-15
Update `AccuDrawStore` outside of `FrameworkAccuDraw`

GerardasB:
## Changes

This PR fixes an issue where 2d vs 3d mode is not respected by AccuDraw keyboard shortcuts or AccuDraw widget when the `FrameworkAccuDraw` class is not used to initialize the `IModelApp`.

The `is3d` state update logic is moved from `FrameworkAccuDraw` class to `ConfigurableUiContent` component.
Currently the logic uses native browser "mousemove" events to identify last hovered viewport, which can be changed in the future, if a more convenient API is added to the core packages.

https://github.com/user-attachments/assets/0489352f-4bba-491c-9189-2c6b6496281f

## Testing

Additionally added `frameworkAccuDraw=0` URL param to `test-app` to facilitate testing of a scenario where `FrameworkAccuDraw` is not used in the `IModelApp` initializer.


saskliutas (approved):

ryanlitzke (approved):
thanks for tackling this! looks good to me.

PR #1562, opened 2026-05-26
Remove internal `Backstage` implementation

GerardasB:
## Changes

This PR replaces underlying internal components that are used to implement the `BackstageComposer` component.

| Before | After |
| ------ | ----- |
| before |  Screenshot 2026-05-28 at 13 44 36 |

Other than a slight visual change, the backstage has greatly improved accessibility. Main changes:

- The backstage is displayed as a modal dialog which prevents from interacting with the rest of the UI while the backstage is open
- The backstage items are now interactive elements (buttons) which can be navigated using the keyboard

## Testing

Tested in `test-app` and storybook:
- [New Badges story](https://itwin.github.io/appui/1562/?path=/story/components-backstagecomposer--badges)
- [Icons story](https://itwin.github.io/appui/1562/?path=/story/components-backstagecomposer--icons)
- [Subtitle story](https://itwin.github.io/appui/1562/?path=/story/components-backstagecomposer--icons)
- [No overlay and custom header](https://itwin.github.io/appui/1562/?path=/story/components-backstagecomposer--icons&args=header:Custom%20header;showOverlay:!false)


ui/appui-react/src/appui-react/backstage/Backstage.scss
```diff
@@ -0,0 +1,63 @@
...
+  }
+}
+
+.uifw-backstage-backstage_content {
+  padding-inline: var(--iui-size-xs);
+}
+
+.uifw-backstage-backstage_divider {
+  margin-block: var(--iui-size-2xs);
+  margin-inline: calc(-1 * var(--iui-size-xs));
```
FlyersPh9: Do we need to worry about that Safari bug here?

If yes, then we need to use long form here.
If no, then we can use short hand on lines 13-14.

ui/appui-react/src/appui-react/backstage/Backstage.tsx
```diff
@@ -0,0 +1,182 @@
...
+      <div className="uifw-backstage-backstage_badge">
+        {/* eslint-disable-next-line @typescript-eslint/no-deprecated */}
+        <Badge type={item.badgeKind || item.badge} />
+      </div>
+      <ListItem.Icon>
+        {/* eslint-disable-next-line @typescript-eslint/no-deprecated */}
+        <ItemIcon iconNode={item.iconNode} icon={item.icon} />
+      </ListItem.Icon>
+      <ListItem.Content>
+        <button
```
mayank99: You can use `ListItem.Action` instead of `button`, and remove the corresponding custom styles.

https://itwinui.bentley.com/docs/list/#with-listitemaction

ui/appui-react/src/appui-react/backstage/Backstage.tsx
```diff
@@ -0,0 +1,182 @@
...
+          "uifw-backstage-backstage_backdrop",
+          !showOverlay && "uifw-transparent"
+        )}
+      />
+      <Dialog.Main
+        aria-label={translate("backstage.label")}
+        {...rest}
+        className={classnames("uifw-backstage-backstage", props.className)}
+      >
+        <Dialog.Content className="uifw-backstage-backstage_content">
```
mayank99, line 67: Above the content, you could use `Dialog.TitleBar` with a visible heading and close button.

mayank99: Also, it would be good to focus the title text when the dialog opens. Currently the dialog itself is focused, which is only good as a fallback.

ui/appui-react/src/appui-react/backstage/Backstage.tsx
```diff
@@ -0,0 +1,182 @@
...
+  const disabled = ConditionalBooleanValue.getValue(item.isDisabled);
+  return (
+    <ListItem
+      data-item-id={item.id}
+      data-item-type="backstage-item"
+      data-item-group-priority={item.groupPriority}
+      data-item-priority={item.itemPriority}
+      data-item-provider-id={isProviderItem(item) ? item.providerId : undefined}
+      disabled={disabled}
+      size="large"
```
mayank99, line 134: The ListItems look a bit chonky (still better than before).

It might look too congested without the `size="large"` though.

ui/appui-react/src/appui-react/backstage/BackstageComposer.tsx
```diff
@@ -179,35 +189,33 @@ export function BackstageComposer(props: BackstageComposerProps) {
...
-          );
-        })
-      )}
-    </NZ_Backstage>
+      {header}
+      {groups.map((group, groupIndex) => {
+        return (
+          <React.Fragment key={groupIndex}>
+            {groupIndex > 0 ? <BackstageDivider /> : null}
+            <List>
```
mayank99: nit: It's a bit strange that `List` is used here and `ListItem` is in a different file. Keep them both in the same file to reduce chances of accidental drift in the future.

ui/appui-react/src/appui-react/backstage/Backstage.scss
```diff
@@ -0,0 +1,63 @@
...
+
+.uifw-backstage-backstage_wrapper {
+  @include zIndex.uicore-z-index(backstage);
+}
+
+.uifw-backstage-backstage {
+  transform: none;
+  inset-block-start: 0;
+  inset-block-end: 0;
+  inset-inline-start: 0;
```
mayank99: Some of these overrides could be avoided by using `placement="top-left"`.

https://itwinui.bentley.com/docs/dialog#placement

ui/appui-react/src/appui-react/backstage/Backstage.tsx
```diff
@@ -0,0 +1,182 @@
...
+      active={isActive}
+      aria-current={isActive ? "true" : undefined}
```
mayank99: `aria-current` should be on the action (button or link) that gets focused.

Also, `aria-current` may not be appropriate if these items don't actually represent the current page/location, or if multiple could be active at the same time.

ui/appui-react/src/appui-react/backstage/Backstage.tsx
mayank99: Part of the reason it looks so chonky is: icons are too small, creating lots of empty space. Not all items have icons, creating even more empty space. Try increasing the icon size? 🤔

ui/appui-react/src/appui-react/backstage/Backstage.scss
GerardasB: Not really, as it's only an issue when `revert-layer` is used. Applied a long form to be safe: [Use long form margin properties](https://github.com/iTwin/appui/pull/1562/commits/b79429f91005ba7791f8777b10884308cfa85617)

ui/appui-react/src/appui-react/backstage/Backstage.tsx
GerardasB: [Use ListItem.Action](https://github.com/iTwin/appui/pull/1562/commits/a165d6b712aca6692a55a5481b405228698e3bcb)

ui/appui-react/src/appui-react/backstage/Backstage.tsx
GerardasB: We have decided to remove the visible title bar as it's not really clear how to name the heading, as "backstage" might not be understood by users. [Remove TitleBar](https://github.com/iTwin/appui/pull/1562/commits/214c74930115a02d25a66825748154eb488bb17d)

[Focus first item](https://github.com/iTwin/appui/pull/1562/commits/1cd03dd168283f4674dab55d7ad7e9351ed3d449)

ui/appui-react/src/appui-react/backstage/Backstage.tsx
GerardasB: Large items look clunky as well (especially when there's no sublabel):
Screenshot 2026-06-03 at 16 05 17

Additionally, seems like `<Icon size="large" />` is overriden by
Screenshot 2026-06-03 at 16 05 43

Let's stick with current visuals. In the future we can re-use https://mui.com/material-ui/react-drawer/

ui/appui-react/src/appui-react/backstage/BackstageComposer.tsx
GerardasB: [Move List](https://github.com/iTwin/appui/pull/1562/commits/1df978b61c4143c9a9fe21a11f7358e08a470c5b)

ui/appui-react/src/appui-react/backstage/Backstage.scss
GerardasB: [Apply top-left placement](https://github.com/iTwin/appui/pull/1562/commits/05f0329d186923e7e69fe3d8d83a6c10d6a97760)

Had to add resets for the margin-start..

ui/appui-react/src/appui-react/backstage/Backstage.tsx
GerardasB: [Move aria-current on a button](https://github.com/iTwin/appui/pull/1562/commits/8f4ada4162b0cea45f84a0258043385e0c811de0)

Only a single item should be marked as current (indicating an active frontstage).

FlyersPh9 (approved):

ui/appui-react/src/appui-react/backstage/Backstage.tsx
mayank99: `Drawer` will not help with this; I assume you mean MUI's `ListItem` specifically.

ui/appui-react/src/appui-react/backstage/Backstage.tsx
mayank99: TitleBar also provides a clear way to dismiss the Dialog.

The stories are not representative of real use cases, so it's hard for me to recommend an alternative. But if there is no good default heading, then it should be left up to consumers to provide one. The custom header example could at least show how to use it.

Would it look weird if the TitleBar was rendered with a close button, but without a title (until passed by consumer)?

GerardasB: > Would it look weird if the TitleBar was rendered with a close button, but without a title (until passed by consumer)?

It would probably look fine, however leaving this as a potential future enhancement, because current solution aligns better with previous design and with MUI Drawer.

PR #1573, opened 2026-06-11
[PropertyGrid]: Reset editor on cancel or abort

### LINKED ISSUE #1543: Property grid: Looking for a way to refresh the property pane
I have a case where a user makes changes to a property via the property pane. If during the commit we determine the value is bad, we want to abandon the change, show a user message, and refresh the property value from what the user changed it to back to the original value in the imodel.

It feels like the property pane should refresh during abandonChanges? But I'd be OK if I could easily ask it to refresh when we call abandon. But I don't see a straightforward way to get it to refresh either.

saskliutas:
## Changes

Closes https://github.com/iTwin/appui/issues/1543

Updated PropertyGrid component to reset editor to initial state when edit is cancelled (`ESC` is pressed) or update is aborted (`onPropertyUpdated` returns `false`).

~~Solution is not straight forward because `onEditCommit` prop used by property renderer to commit value does not return any result. Ideally we could handle aborted commit where editor is rendered. However, those are public APIs and changed signature of `onEditCommit` would be a breaking changes.~~

~~Opted to introduce new `editKey` props to `VirtualizedPropertyGridContext`. It is updated when commit was aborted and can be used in editor component key to reset it. ~~

~~Cancellation is handled next to editor in a similar way where key is updated when `onCancel` is invoked.~~

`FlatPropertyRenderer` now handles editor reset:
- If `onCancel` is invoked new key for editor component is generated. 
- If `onEditCommit` returns `{ status: "cancel" }` new key for editor component is generated.

This should ensure that if new value is not desired either by cancelling edit using `ESC` key or cancelling/aborting during commit editor will reset back to initial state. If commit succeeds it will keep current state to avoid flashing `entered value` -> `reset to initial value` -> `new committed value`.

## Testing

Added unit tests and manually tested in test-app.

### DIFF
--- a/.changeset/property-grid-abort-commit-reset.md
+++ b/.changeset/property-grid-abort-commit-reset.md
@@ -1,0 +1,5 @@
+---
+"@itwin/components-react": patch
+---
+
+`VirtualizedPropertyGrid` now resets the active editor back to its initial value when a commit is aborted by the consumer (`onPropertyUpdated` returns `false`). Cancelling an edit now also resets editor.
--- a/apps/test-providers/src/ui/widgets/PropertyGridWidget.tsx
+++ b/apps/test-providers/src/ui/widgets/PropertyGridWidget.tsx
@@ -8,6 +8,7 @@ import {
   IPropertyDataProvider,
   PropertyData,
   PropertyDataChangeEvent,
+  PropertyUpdatedArgs,
   VirtualizedPropertyGridWithDataProvider,
 } from "@itwin/components-react";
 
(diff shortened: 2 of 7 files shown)


grigasp:
> Solution is not straight forward because `onEditCommit` prop used by property renderer to commit value does not return any result. Ideally we could handle aborted commit where editor is rendered. However, those are public APIs and changed signature of `onEditCommit` would be a breaking changes.

Could we change the return type of `onEditCommit` to `void | boolean` or `void | "cancel"`, where falsy value would keep current behavior and truthy value cancels the edit?

saskliutas:
> > Solution is not straight forward because `onEditCommit` prop used by property renderer to commit value does not return any result. Ideally we could handle aborted commit where editor is rendered. However, those are public APIs and changed signature of `onEditCommit` would be a breaking changes.
> 
> Could we change the return type of `onEditCommit` to `void | boolean` or `void | "cancel"`, where falsy value would keep current behavior and truthy value cancels the edit?

Changed `onEditCommit` to `void | Promise<{ status: "success" | "cancel"}>`. Initially it looked like it would touch more broadly but it looks we just have semi duplicate components that probably were left after `PropertyGrid` -> `VirtualizedPropertyGrid` migration.

grigasp (approved):

GerardasB (approved):

PR #1577, opened 2026-06-19
KOQ property renderer

saskliutas:
## Changes

Introduced renderer for properties with kind of quantity. Name `KOQ_RENDERER_NAME` should be used to specify that this renderer should render property. It looks up `FormatterSpec` and renders formatted value. 

Updated new editors to handle merged values. New `isMerged` flag was introduced to `ValueMetadata` and it is now used by `Text`/`Numeric`/`Quantity`/`Fallback` editors to render `--` label when value is merged and no explicit values is set.

## Testing

Added unit tests

### DIFF
--- a/ui/components-react/src/components-react/UiComponents.json
+++ b/ui/components-react/src/components-react/UiComponents.json
@@ -1,7 +1,7 @@
 {
   "general": {
-    "true": "true",
-    "false": "false",
+    "true": "True",
+    "false": "False",
     "loading": "loading",
     "noData": "The data required for this tree layout is not available in this iModel.",
     "length": "Length",
--- a/ui/components-react/src/components-react/new-editors/editors/FallbackEditor.tsx
+++ b/ui/components-react/src/components-react/new-editors/editors/FallbackEditor.tsx
@@ -14,6 +14,7 @@ import {
   isInstanceKey,
   isNumeric,
   isText,
+  MERGED_VALUE,
 } from "../values/ValueUtilities.js";
 
 /* v8 ignore start */
@@ -25,7 +26,7 @@ import {
 export function FallbackEditor({ value, size, id, metadata }: EditorProps) {
   const textValue = getTextValue(value);
   const displayValue =
-    textValue !== undefined ? textValue : metadata.isMerged ? "--" : "";
+    textValue !== undefined ? textValue : metadata.isMerged ? MERGED_VALUE : "";
   return <Input id={id} readOnly value={displayValue} size={size} />;
 }
 
--- a/ui/components-react/src/components-react/new-editors/editors/FallbackEditor.tsx
+++ b/ui/components-react/src/components-react/new-editors/editors/FallbackEditor.tsx
@@ -22,8 +22,11 @@ import {
  * Fallback editor that renders readonly value if no editor is found.
  * @internal
  */
-export function FallbackEditor({ value, size, id }: EditorProps) {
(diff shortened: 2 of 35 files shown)


MartynasStrazdas (approved):

ui/components-react/src/components-react/new-editors/editors/FallbackEditor.tsx
```diff
@@ -22,8 +22,11 @@ import {
  * Fallback editor that renders readonly value if no editor is found.
  * @internal
  */
-export function FallbackEditor({ value, size, id }: EditorProps) {
-  return <Input id={id} readOnly value={getTextValue(value)} size={size} />;
+export function FallbackEditor({ value, size, id, metadata }: EditorProps) {
+  const textValue = getTextValue(value);
+  const displayValue =
+    textValue !== undefined ? textValue : metadata.isMerged ? "--" : "";
```
grigasp: what do you think about creating a `const` for "--"?

ui/imodel-components-react/src/imodel-components-react/properties/KoqPropertyRenderer.tsx
```diff
@@ -0,0 +1,120 @@
...
+        record={record}
+        context={context}
+        stringValueCalculator={convertRecordToString}
+      />
+    );
+  }
+  return <KoqRendererImpl imodel={imodel} {...props} />;
+}
+
+function KoqRendererImpl(props: {
```
grigasp: `KoqRenderer` and `KoqRendererImpl` are both rendering `PrimitivePropertyValueRendererImpl`, the difference is what they pass as `stringValueCalculator`. In `KoqRendererImpl` it already depends on `imodel`, so why not just merge these two components together and just adjust how `stringValueCalculator` handles missing `imodel`?

```ts
  const stringValueCalculator = React.useCallback(
    async (recordToFormat: PropertyRecord) => {
      return imodel ? formatKoqValue(recordToFormat, imodel) : convertRecordToString(recordToFormat);
    },
    [imodel]
  );
```

ui/components-react/src/components-react/new-editors/editors/FallbackEditor.tsx
saskliutas: Added const for merged value

ui/imodel-components-react/src/imodel-components-react/properties/KoqPropertyRenderer.tsx
saskliutas: Cleaned up

grigasp (approved):

ui/components-react/src/components-react/UiComponents.json
```diff
@@ -1,7 +1,7 @@
 {
   "general": {
-    "true": "true",
-    "false": "false",
+    "true": "True",
```
GerardasB, line 3: Is this change intentional? This is used by `BooleanTypeConverter`.

saskliutas: Yes. TypeConverters are mainly used by property renderers. Previously we provided formatted values to property grid but now we are switching to leave formatting for editors/renderers. With lower case these values look bad and out of place.

Before this change:
image

After:
image

GerardasB: I was worried, that consumers might need to type it in case-sensitively, but after re-reading the code, seems like that is not the case.

GerardasB (approved):

PR #1595, opened 2026-08-04
Update KoqPropertyRenderer to handle format changes

saskliutas:
## Changes

Updated KoqPropertyRenderer to handle format changes. Value is reformatted after unit system of format change.

## Testing

Added unit tests

### DIFF
--- a/ui/imodel-components-react/src/imodel-components-react/properties/KoqPropertyRenderer.tsx
+++ b/ui/imodel-components-react/src/imodel-components-react/properties/KoqPropertyRenderer.tsx
@@ -56,7 +56,9 @@ function KoqRenderer(props: {
 }) {
   const { record, context } = props;
   const imodel = useIModelConnection();
-  const version = useFormattingChangeVersion(record.property.kindOfQuantityName);
+  const version = useFormattingChangeVersion(
+    record.property.kindOfQuantityName
+  );
   const stringValueCalculator = React.useCallback(
     async (recordToFormat: PropertyRecord) => {
       return imodel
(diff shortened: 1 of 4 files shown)


MartynasStrazdas (approved):

ui/imodel-components-react/src/imodel-components-react/properties/KoqPropertyRenderer.tsx
JonasDov, line 1: Asked locally ai to review, a few comments:

1. Provider changes reformat every visible KOQ value.
In useFormattingChangeVersion, onFormatsChanged ignores formatsChanged: "all" | string[]. It should only bump when "all" or the current koqName is included. With many property rows, one format edit currently causes global remounting and async formatting work.

2. Forced remounting creates visible flicker.
The versioned key remounts PrimitivePropertyValueRendererImpl. Its async state resets to undefined, so the existing value disappears or becomes context.defaultValue while formatting completes. Updating while retaining the previous formatted value would provide better behavior.

3. The principal new provider path is untested.
Tests cover active unit-system and quantity-format events, but none emit formatsProvider.onFormatsChanged. That misses both the advertised behavior and the necessary unrelated-KOQ filtering.

JonasDov: My thoughts on the comments:
1. I am not sure if `onFormatsChanged.formatsChanged` includes the `koqName` or not. If not -> then this comment seems invalid.
2. I think it is not that important since format changes should not be that common AND formatting should not take that much time.
3. None of the added tests check if `onFormatsChanged` is raised or not. Would agree that a test is needed.

saskliutas: 1. Looked through formatSet source it looks like it should match koqName. Added handling for `all` and specific koqName.
2. Changed to use `version` state variable in `useCallback` instead of `key` to force reformat.
3. Added tests

JonasDov (approved):

GerardasB (approved):

GerardasB (approved):

PR #1600, opened 2026-08-24
Fix cube navigation opacity issue when theme bridge is enabled

FlyersPh9:
## Changes

@mjkertz found an issue in the cube navigation when theme bridge is turned on where the sides of the cube are transparent.  This is because `--iui-color-background-zebra` maps to a value that includes opacity when theme bridge is enabled.

To retain the original design and also fix the issue when theme bridge is enabled:
- **Apply `background-color: var(--iui-color-background);` to all cube faces.**  This works because `background: linear-gradient();` (which includes the opacity issue) applies _above_ `background-color`.
- **Change `.cube-bottom` from `background: var(--iui-color-background-zebra);` to `background: linear-gradient(var(--iui-color-background-zebra), var(--iui-color-background-zebra));`.** By making it a `linear-gradient` it allows it to apply above the `background-color`.

We're discussing internally if the gradient is worth keeping at all and simplifying it to simply use `background-color` for all cube faces.

Also bumped `@stratakit/bricks` & `@stratakit/foundations` to the latest.

## Testing

| Before | After |
| -- | -- |
| before | after |

### DIFF
--- a/.changeset/hungry-hats-doubt.md
+++ b/.changeset/hungry-hats-doubt.md
@@ -1,0 +1,5 @@
+---
+"@itwin/imodel-components-react": patch
+---
+
+Fixed issue where navigation cube would appear semi-transparent when theme bridge was enabled.
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -82,11 +82,11 @@ catalogs:
       specifier: ^5.8.1
       version: 5.8.1
     '@stratakit/bricks':
-      specifier: ^0.4.5
-      version: 0.4.5
+      specifier: ^0.5.6
+      version: 0.5.6
     '@stratakit/foundations':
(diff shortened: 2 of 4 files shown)


GerardasB:
Seems like there's some CSS specificity (or ordering) issue and the transparency is still visible, after running:
```
pnpm dev
pnpm app
```

And opening http://localhost:3000/blank?frontstageId=main&themeBridge=1

FlyersPh9:
> Seems like there's some CSS specificity (or ordering) issue and the transparency is still visible

Appears to be fixed after https://github.com/iTwin/appui/pull/1600/commits/5de8de0231479f5a541a3f7c8ae88e331f84e393.

GerardasB (approved):
Looks good!

PR #1618, opened 2026-09-13
fix: don't wait 300ms for single widget tab clicks

### LINKED ISSUE #1617: Widget tab clicks unnecessarily wait 300ms for double click detection
### Describe the bug

Simple tab switching on any AppUI application is not very responsive. Digging into this, I found:

https://github.com/iTwin/appui/blob/5f0ebc7cbcf3a04de779d937ada3f66dfdd4857d/ui/appui-react/src/appui-react/layout/widget/Tab.tsx#L199

On mouse/pointer release, it increments a click counter and starts that timer. Only when the timer expires does it call `handleClick()`, which dispatches `WIDGET_TAB_CLICK`. The reducer then changes the active tab.

I don't believe single clicks, especially for something as basic as tab navigation, should be held up for a possible double click. 
I see that double clicking is used for undocked widgets to collapsed/expanded. Would it be acceptable to respond to the first click immediately, then monitor for a possible second click? Because even for an undocked widget, the double click on a different tab does not collapse/expand that widget.

### To Reproduce

Click to switch widget tabs on any AppUI application. 

### Expected Behavior

Switching widget tabs should be near-instantaneous.

ben-polinsky:
Fixes #1617 

## Changes

Widget tabs now respond immediately to clicks instead of waiting 300ms for double-click detection. Double-clicking an inactive tab selects it without collapsing the widget. Double-clicking an active floating tab still expands or collapses it.

## Testing

- Compared the original and updated behavior in the AppUI test app (see videos below)
- Added tests for immediate switching, floating-widget double clicks, rapid switching between tabs, and clicks outside the double-click window. All 19 tab tests pass.
<details>
  <summary>Vids</summary>
After:

https://github.com/user-attachments/assets/9cf2112e-fa34-4cb9-a166-f08f3fc0a2ec

Before:

https://github.com/user-attachments/assets/fc319d4c-6aea-4a7a-b799-dc6cdd3c1767
</details>


ui/appui-react/src/appui-react/layout/state/NineZoneAction.ts
```diff
@@ -198,6 +198,8 @@ export interface WidgetTabDoubleClickAction {
   readonly widgetId: WidgetState["id"];
   readonly floatingWidgetId: FloatingWidgetState["id"] | undefined;
   readonly id: TabState["id"];
+  /** Explicit minimized state after the first click takes effect. */
+  readonly minimized?: boolean;
```
GerardasB: This makes it hard to reason about, we could simplify this by replacing this action with `WidgetTabMinimizeAction` or even better, with `FloatingWidgetMinimizeAction`, since only floating widgets can be minized?

E.g.: https://github.com/iTwin/appui/commit/2609389cd968b61c8e184238519c7a2c1738cfe5

ui/appui-react/src/appui-react/layout/state/NineZoneAction.ts
GerardasB, line 1: Missing changeset, run `pnpm changeset`

GerardasB (approved):
Looks good, although it would be beneficial to simplify this further to either:
- Align with [dblclick_event](https://developer.mozilla.org/en-US/docs/Web/API/Element/dblclick_event) and always fire two click events before a double click
- Remove double click behavior, and collapse only when an active tab of a floating widget is clicked

ui/appui-react/src/appui-react/layout/state/NineZoneAction.ts
ben-polinsky: Yes, for sure.

PR #1619, opened 2026-09-14
fix: update dependencies with high and critical vulnerabilities

ben-polinsky:
## Changes

Bump deps in package.json where possible. Electron major bump is fine as it is for a test app.
No new overrides, just bumped versions in overrides.

Also fixed what seems to be a flaky unit test.

## Testing

all tests run / CI


ben-polinsky:
@GerardasB lemme know about this and my [300ms click fix](https://github.com/iTwin/appui/pull/1618) :)

GerardasB (approved):
Looks good, thanks for fixing those!
Long term we should aim to simplify test apps and remove as many deps as possible.

ui/core-react/package.json
GerardasB, line 1: Would be great to mention version bump in `pnpm changeset`

ben-polinsky: Done

